PR #68, opened 2019-04-15
[small PR] remove native number spinner on firefox

davejsdev:
## Reason for change

Removes native number arrows on number input for firefox.

## Testing

1. Go to custom plan selector in firefox.
2. Ensure number input field doesn't have the native up/down arrows on the right of the number field.


drwpow (approved):
LGTM but @davidleger95 want to hear your thoughts

PR #80, opened 2019-04-19
Add JSON parsing concept

dangodev:
## Reason for change
While browsing the [JS Integration docs](https://stenciljs.com/docs/javascript) for Stencil, this line caught my eye:

```html
<todo-list my-object="{}" my-array="[]"></todo-list>
```

“Oh! Stencil can just parse attributes as JSON! Neat”

Well, there’s a reason why that object & array is empty—in order for it to work, you have to do gross things like this: `my-array="[&quot;itemOne&quot;,&quot;itemTwo&quot;]"`. Hooray! 🙃

Still, Vue.js does support JSON attributes (as long as the attribute name is prefixed with `:`), and I like the idea. This is a stab at implementing that concept in Stencil (especially since the docs seem to hint that it’s easy).

An alternate method is to use a parser like [JSON5](https://github.com/json5/json5), but that’s ~`30 KB` of weight.

### Sidenote:

We should probably make more liberal use of `@Watch`, at least for the entry components. Our components are not good at listening to changes after initialization, and that would really improve support.

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -196,12 +196,12 @@ export namespace Components {
   }
 
   interface ManifoldMarketplaceResults {
-    'featured'?: string;
+    'featured': string[];
     'linkFormat'?: string;
     'services': Catalog.Product[];
   }
   interface ManifoldMarketplaceResultsAttributes extends StencilHTMLAttributes {
-    'featured'?: string;
+    'featured'?: string[];
     'linkFormat'?: string;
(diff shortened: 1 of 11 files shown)


dangodev:
Gonna close it. Still don’t have an answer, short of loading a hefty JSON5 parsing library on the clientside. For now, we can just set things in JavaScript anyway, and the provision button is the only thing that has needed it so far.

PR #87, opened 2019-04-22
Propose event standard

dangodev:
## Reason for change
I’m working on a provision button, and when surveying the events I’m finding what @davidleger95 pointed out a while ago: we have a few events now which don’t stick to a standard. I propose:

`manifold-[componentName]-[event]`

Example: `manifold-marketplace-click` or `manifold-planSelector-change`.

The only exception to that is the template cards in `manifold-marketplace`, which I felt should get a different event from 'manifold-marketplace-click' because they’re not products, and emit a different value. For those, I have `manifold-template-click` as the event.

Also adds tests for events, and updates the READMEs with better documentation on events.

Screen Shot 2019-04-22 at 11 27 52

### DIFF
--- a/src/components/manifold-plan-details/manifold-plan-details.tsx
+++ b/src/components/manifold-plan-details/manifold-plan-details.tsx
@@ -42,7 +42,7 @@ export class ManifoldPlanDetails {
     const features = this.initialFeatures();
     this.features = features; // Set default features the first time
     if (this.plan && this.product) {
-      // Drew: not sure how to tell Stencil that these should always be here on load
+      // This conditional should always fire on component load
       const detail: EventDetail = {
         planId: this.plan.id,
         planLabel: this.plan.body.label,
@@ -57,7 +57,7 @@ export class ManifoldPlanDetails {
     const features = { ...this.features, [name]: value };
     this.features = features; // User-selected features
     if (this.plan && this.product) {
-      // Same as above: parent prevents loading this component till these are set
+      // Same as above: this should always fire; just needed for TS
       const detail: EventDetail = {
         planId: this.plan.id,
         planLabel: this.plan.body.label,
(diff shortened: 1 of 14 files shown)


src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -4,65 +4,75 @@ import { initialFeatures } from '../../utils/plan';
...
-    bubbles: true,
-  })
-  ctaClicked: EventEmitter;
 
   componentWillLoad() {
     const features = this.initialFeatures();
     this.features = features; // Set default features the first time
-    this.updatedPlanHandler({ features }); // Dispatch change event when loaded
+    if (this.plan && this.product) {
+      // Drew: not sure how to tell Stencil that these should always be here on load
```
davidleger95: what do you mean by "these"?

dangodev: Oh `this.plan` and `this.product`.

Stencil makes you either mark a component as optional, or provide a fallback. For Plan & Product, this component doesn’t render until it’s set from its parent. So it throws an error in the typechecker, even though in practice this if statement shouldn’t be doing anything

dangodev: I just wanted to make a note that this component should always load, and fire the conditional

davidleger95: ah, sounds good

dangodev:
@davidleger95 since you’re releasing to npm, want to decide when to merge this? This will break existing events for the demo site (mainly the navigation link from `/` -> `/new/resource`/).

But going forward there are more tests to ensure this doesn’t change in the future.

dangodev:
@davidleger95 gonna merge this to make the region dropdown easier. Ping me if you need help with the changes to the demosite!

PR #88, opened 2019-04-22
Resource thing

dangodev:
## Reason for change
Adds a data wrapper for resource details:

```html
<manifold-resource-details resource-id="my-resource" />
```

_Note: this PR also removes our specs, because I realized Catalog was fine to expose, but for our internal stuff it’s good to not make those public_.

## Testing
The auth method is mostly for demo purposes, so it won’t be documented, but here’s how to use it:

1. Log into the Manifold Dashboard (either staging or prod, depending on the `<manifold-connection>` env)
2. Locate the `resource.id` from a resource you’d like to view (you can find this in your **Network** tab—search for `/resources/:id` or something. Place that in `<manifold-resource-details resource-id="my-id" />`.
3. Next, run the following in your browser console: `localStorage.getItem('manifold_api_token')`. Copy the value.
4. In a browser tab showing the UI docs, run the following in a browser console: `localStorage.setItem('manifold_api_token', '[API token from previous step]')`.
5. Refresh, and voila! You should be looking at all the details for that resource.

Note that the `manifold_api_token` is destroyed whenever you log out of Dashboard, or a week has passed and it will expire on its own.

Screen Shot 2019-04-22 at 12 06 12


.gitignore
```diff
@@ -1,5 +1,9 @@
 dist/
 pkg/
+src/spec/catalog
+src/spec/gateway
+src/spec/marketplace
+src/spec/provisioning
```
dangodev: Note our specs are now Gitignored because this is now a public repo, but these are private

src/readme.md
```diff
@@ -15,6 +15,16 @@ npm run dev
...
+Lastly, copy these specs from [our specs][specs] into here:
+
+- `src/spec/catalog/v1.yaml`
+- `src/spec/gateway/v1.yaml`
+- `src/spec/marketplace/v1.yaml`
+- `src/spec/provisioning/v1.yaml`
+
+This is a public repo, but our API specs aren’t public for now (the types are
+OK though).
+
```
dangodev: Another note here on keeping our internal API specs out of this repo

src/types/catalog.ts
```diff
@@ -140,6 +143,12 @@ declare namespace Catalog {
     // unspecified, then regions have no impact on this
     // resource.
     region?: 'user-specified' | 'unspecified';
+    // Describes the credential type that is supported by this product.
+    //
+    // * `single`: Only one credential is supported at the same time.
+    // * `multiple`: Multiple credentials are supported at the same time.
+    // * `unknown`: The credential type is unknown.
+    credential?: 'single' | 'multiple' | 'unknown';
```
dangodev: The generated types are fine to expose—it skips the auth and other internal bits

package.json
```diff
@@ -30,9 +30,11 @@
     "format:ts": "prettier --write 'src/**/*.{ts,tsx}'",
     "format:watch": "npx onchange 'src/**/*.{css,ts,tsx}' -- npm run format",
     "generate:docs": "node scripts/docs",
-    "generate:specs": "npm run generate:specs:catalog && npm run generate:specs:gateway",
+    "generate:specs": "npm run generate:specs:catalog && npm run generate:specs:gateway && npm run generate:specs:marketplace && npm run generate:specs:provisioning",
```
davidleger95: no sure I know exactly what's going on here. doesn't this command generate the files you're removing for privacy?

davidleger95: or do you need creds to use this command?

dangodev: It does but it’s only used manually. You just have to grab specs & store them locally, once, during setup. Then they’ll stay there & you can run this.

dangodev: And no, no creds necessary. Just the YAML files it’s referring to (link in `src/README`).

PR #92, opened 2019-04-23
added product id to data components

davidleger95:
## Reason for change

Added resource details table

Screen Shot 2019-04-23 at 10 39 19 PM

Screen Shot 2019-04-26 at 1 02 44 AM
Screen Shot 2019-04-26 at 1 03 05 AM


src/components.d.ts
```diff
@@ -72,9 +72,13 @@ export namespace Components {
...
     'connection': Connection;
     /**
+    * Product ID
+    */
+    'productId'?: string;
+    /**
     * URL-friendly slug (e.g. `"jawsdb-mysql"`)
     */
-    'productLabel': string;
+    'productLabel'?: string;
```
davidleger95: is there a way to specify either `productId` xor `productLabel` is required?

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -1,34 +1,46 @@
...
+        .then(response => response.json())
+        .then(data => {
+          this.product = { ...data[0] };
+        });
+    } else if (this.productId) {
+      // Don’t return this promise to invoke the loading state
+      fetch(`${this.connection.catalog}/products/${this.productId}`, withAuth())
+        .then(response => response.json())
+        .then(data => {
+          this.product = { ...data };
```
davidleger95: by productId doesn't wrap response in an array

src/components.d.ts
dangodev: lol no

Stencil (rightly) enforces HTML by design: we can’t make anyone specify an attribute. All we can do is fail gracefully (and maybe even give some friendly `console.error` warnings where helpful).

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
dangodev: Correct. Also one thing I’ve been doing is:

```diff
-data
+products: Catalog.Product[]
```

to be safer. Totally optional, but helps simple mistakes.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -1,34 +1,46 @@
...
+    }
   }
 
   render() {
-    return this.product ? <img src={this.product.body.logo_url} alt={this.alt || this.product.body.name} /> : <slot />
+    return this.product ? (
+      <img src={this.product.body.logo_url} alt={this.alt || this.product.body.name} />
+    ) : (
+      <slot />
+    );
```
dangodev: Sorry; my Prettier bugged out for a week and stopped working. It’s fixed now! But explains why a lot of the files I worked on lately are all outta whack

src/index.html
```diff
@@ -779,6 +780,61 @@ <h2>Example</h2>
...
+              </table>
+
+              <!-- END README -->
+            </div>
+            <h2>Example</h2>
+            <div class="docs-example">
+              <div class="docs-example-inner">
+                <manifold-resource-details
+                  resource-id="268ak39rn9czavv1qd9m93t482gmm"
+                ></manifold-resource-details>
```
davidleger95: A problem with this demo is that it requires an auth token. Not sure if we want to build auth into the docs or not, but that might be what we need to make this work.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: didn't really think about it :P

src/components/manifold-resource-details/manifold-resource-details.tsx
```diff
@@ -3,26 +3,67 @@ import { Component, Prop, Element, State } from '@stencil/core';
...
+        {expanded_features.map(feature => [
+          <FeatureLabel feature={feature} />,
+          <FeatureValue
+            features={customFeatures || initialFeatures(expanded_features)}
+            feature={feature}
+            isExistingResource
+            onChange={() => {}}
+          />,
+        ])}
+      </dl>
```
davidleger95: So I just took this from the plan details. I think it can use some more abstraction but for now, it works. Also might need to refactor the FeatureValue props to make the onChange optional. You had mentioned refactoring to split on static features and custom features, but I'm not sure if that abstraction will lead to more complexity. When I did my refactor of those components initially, I thought about doing that but decided splitting on feature type was the simpler abstraction. 

Could be wrong, but not sure how much benefit we'd get from another refactor 🤷‍♂️

src/components/manifold-resource-details/manifold-resource-details.tsx
```diff
@@ -3,26 +3,67 @@ import { Component, Prop, Element, State } from '@stencil/core';
...
+
+    const { expanded_features = [] } = this.plan.body;
+    const { features: customFeatures } = this.resource.body;
+
+    return (
+      <dl class="features">
+        {expanded_features.map(feature => [
+          <FeatureLabel feature={feature} />,
+          <FeatureValue
+            features={customFeatures || initialFeatures(expanded_features)}
```
davidleger95: I think this makes sense

dangodev: Hm. So both will return something like `{feature1: true, feature2: 'foo'}` which doesn’t have user-friendly strings. We may need something more to display it properly.

dangodev: Also `customFeatures` won’t display static features, so we’ll need to pass in `initialFeatures()` as well. They may need to be handled side-by-side.

src/components/manifold-resource-details/manifold-resource-details.tsx
dangodev: I still feel pretty strongly that static, custom, and measured features should be separated. Static features basically always behave like strings; measured features can only be numbers and not another type, and customizable can be string, number, or boolean. It works for now, but I think there’s more logic than needed. And to the comment above, our backend handles custom features differently than static ones. I think we can simplify display logic as well by separating them (IMO I don’t think they should have been combined in the data structure to begin with—too much polymorphism to wrestle with).

dangodev: I think this was a slow realization as I was trying to test everything as well—I think the display functions I wrote could be organized a lot better. Everything’s tested now, but I think I share your sentiment that everything could be organized better for sure going forward.

src/components/manifold-resource-details/manifold-resource-details.tsx
dangodev: Kinda reiterating part of my comment below (sorry) but I think the easiest display path here is to display static, then measurable, then customizable features in the table view.

Resources also can display their usage for measurable features, and you’ve already found `resource.body.features` which only relates to customizable ones. I may be wrong but I think repeating that logic between string, number, and boolean features is more work, especially considering static features only need `name` and `value_string` really to display.

dangodev: Getting tired! Maybe we can sync up tomorrow

davidleger95: who woulda thought displaying a list of features would be this complex 🙃 

yeah, syncing would be good. It seems there's a lot more under the surface with features that I've yet to see.

src/components/manifold-resource-details/components/CustomFeature.tsx
```diff
@@ -0,0 +1,25 @@
+import { FunctionalComponent } from '@stencil/core';
+import { LockedFeature } from './LockedFeature';
+import { FeatureInput } from './FeatureInput';
+
+const getCustomValue = (feature: Catalog.ExpandedFeature, value: number | string | boolean) => {
+  // TODO get actual value
+  console.log({ feature, value });
+  return value;
```
davidleger95: Any suggestions for how I should do this? I forget if there's a function . somewhere that will get the value. Might have to do this at a higher level too

src/components/manifold-resource-details/components/FeatureInput.tsx
```diff
@@ -0,0 +1,48 @@
+import { FunctionalComponent } from '@stencil/core';
+import { stringFeatureOptions } from '../../../utils/plan';
+
+const onChange = (e: Event) =>
+  new CustomEvent('manifold-resource-feature-change', { detail: e && '' });
```
davidleger95: This is a placeholder. I was thinking of maybe triggering a custom event for onCHange so it doesn't need to be passed down. thoughts on this?

src/components/manifold-resource-details/readme.md
```diff
@@ -1,4 +1,10 @@
-# manifold-resource-details
+# Resource Details
+
+The resource details component displays a table of features belonging to a resource. 
+
+```html
+<manifold-resource-details resource-id="268ak39rn9czavv1qd9m93t482gmm"></manifold-resource-details>
+```
```
davidleger95: Removed this from the docs site for now because it won't work without auth

src/index.html
```diff
@@ -276,42 +276,45 @@ <h1 id="marketplace">Marketplace</h1>
               <p>A list of all Manifold services.</p>
               <pre><code class="html language-html">&lt;manifold-marketplace /&gt;
 </code></pre>
-              <h2 id="blacklistwhitelist">Blacklist/whitelist</h2>
+              <h2 id="catalogcuration">Catalog curation</h2>
```
davidleger95: don't worry about this diff. just pulled from master

src/components/manifold-resource-details/components/MeasurableFeatureUsageDisplay.tsx
```diff
@@ -0,0 +1,14 @@
...
+  // features: UserFeatures;
+  feature: Catalog.ExpandedFeature;
+  value?: string | number | boolean;
+}
+
+export const MeasurableFeatureUsageDisplay: FunctionalComponent<
+  MeasurableFeatureUsageDisplayProps
+> = ({ feature, value = feature.value_string }) => {
+  // TODO build usage UI
+  return <span>{value}</span>;
```
davidleger95: not sure where the value for usage comes from. is it somewhere on the resource?

src/components/manifold-resource-details/components/CustomFeature.tsx
dangodev: I actually forgot where it lived, but it looks like it’s `billing`: https://api.billing.stage.manifold.co/v1/docs#tag/Resource

I can make a PR for it that you can either merge in, or just copy the changes into yours. One sec

dangodev: Whoops replied to the wrong comment! This was for your billing question below 👇

dangodev: But to this question: if this is for displaying a custom value based on the resource, then you’d do it differently based on type:

- string: search `feature.values[]` for something where `values[].value` equals the resource, then display `value[].name`.
- number: you can probably just display the value straight from the resource? 🤷‍♂️ 
- boolean: same—can probably just display the value and style it like the others (“Yes” or ”No”)

So seems like string features are the odd one out here.

src/components/manifold-resource-details/components/FeatureInput.tsx
dangodev: Hm. I guess I’d ask what this functional component is doing—if for example this component’s parent is keeping track of feature state, then is this component needed?

If this component solves a lot of complexity, then a custom event will work, but I’d question if the `switch` below can just be absorbed into the component that’s keeping track of state

src/components/manifold-resource-details/components/CustomFeature.tsx
davidleger95: the boolean value is actually `true`/`false` but the `values[].value` for boolean a string of `"true"`/`"false"` 🙃

davidleger95: but yeah, that's what I was thinking i'd have to do. was hoping there would be some function already that would do it for me :)

src/components/manifold-resource-details/components/FeatureInput.tsx
davidleger95: Yeah, This component it more for organizational purposes. Just trying to keep files small and focused on a single task. Good suggestion though, I'll see if moving it up is less complex

src/components/manifold-resource-details/components/CustomFeature.tsx
dangodev: > the boolean value is actually true/false but the values[].value for boolean a string of "true"/"false" 🙃

🙄 Ugh of course. Things like that always get me. I think in some places in Dashboard I would just add `value.toString === 'true'` just because I was never sure 🤷‍♂️

src/components/manifold-resource-details/components/CustomFeature.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if (!feature.values) return value;
+  switch (feature.type) {
+    case 'boolean': {
+      // must use weak equality because labels are strings
+      // eslint-disable-next-line eqeqeq
+      const valueObj = feature.values.find(val => val.label == value, '');
+      return valueObj && valueObj.name;
+    }
+    case 'number': {
+      const { value: { numeric_details: { suffix = '' } = {} } = {} } = feature;
```
davidleger95: Can we add support for optional chaining please? 😓  

https://github.com/tc39/proposal-optional-chaining

src/components/manifold-resource-details/components/MeasurableFeatureUsageDisplay.tsx
```diff
@@ -0,0 +1,27 @@
...
+  console.log({ cost_ranges });
+  return (
+    <ul>
+      {cost_ranges.map(range => (
+        <li>
+          {$((range.cost_multiple || 0) / COST_COEFFICIENT)} / {suffix} up to {range.limit} {suffix}
+        </li>
+      ))}
+    </ul>
+  );
```
davidleger95: Added this UI which similar to what's in dashboard now. I didn't want to dig into the billing api stuff just yet. Maybe we can do that in a future PR.

src/components/manifold-resource-details/components/FeatureInput.tsx
davidleger95: I started to add the change handler stuff, trying to use the plan selector as a guide, but it started getting kinda messy. It seems like there's a lot of stuff that's not necessary, just not sure exactly what.

There's stuff with regions and the product data that don't seem necessary but I wasn't sure if I could safely ignore those pieces.

davidleger95:
Still need to add tests

PR #93, opened 2019-04-24
Happo integration

sslotsky:
## Reason for change

Set up happo.io for visual regression testing.


src/components/manifold-product-page/manifold-product-page.e2e.ts
```diff
@@ -1,5 +1,5 @@
 import { newE2EPage } from '@stencil/core/testing';
-import { Product, Provider } from '../../spec/mock/catalog';
```
sslotsky: This was failing TS compilation by the way, not sure how it made it in :confused:

davidleger95: I don't think it will fail for test files

davidleger95:
Should I be reviewing these PRs, or is this just a Platform Squad concern? Not really sure what's going on with Happo stuff, so my review won't mean much. 🤷‍♂ I'll click the button if that's all you need though 😄

sslotsky:
@davidleger95 I'd definitely like for DX to be reviewing this. I think it will be up to DX devs to write more visual regression tests, so looking this over for any concerns or questions you might have is important.

davejsdev (approved):
looks good to me!

PR #97, opened 2019-04-25
Chloeedeal/happo testing

batgirl:
Testing to see if TypeScript files are picked up by Happo.


webpack.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
sslotsky: The docs don't mention needing this. I was actually forced to use the TS plugin locally when I was messing with trying to mock fetch calls, and found that I didn't need it either. But I assume you must have found that you needed it, or else it wouldn't be here. What did you run into that required a custom webpack config?

batgirl: Oh yes, it wasn't working for me until I added this - the [happo typescript plugin](https://github.com/happo/happo-plugin-typescript) uses [ts-loader](https://github.com/TypeStrong/ts-loader) and was erroring out until I added it. This webpack config is from the ts-loader docs. Were you using the happo typescript plugin before?

sslotsky: I was, yes :confused:

sslotsky: I definitely need `ts-loader` but not the config file :man_shrugging:

batgirl: Oh maybe that's my case too then - it didn't work when I just installed ts-loader but it did once I added the config, but perhaps I added too much config, if that makes sense 😛

batgirl: I'll try again without the webpack file.

src/components/manifold-select/manifold-select-happo.tsx
```diff
@@ -0,0 +1,22 @@
+export const primary = (_render: void) => {
```
sslotsky: I had the `render` param in there from some async experiments, but we don't really need it. Let's change this to:

```js
export const primary = () => {
```

and then I think we can merge this one and close mine.

sslotsky: Also I suppose we can just name it `.ts` instead of `.tsx` because there's no JSX. Not a big deal.

sslotsky (changes requested):
Some tiny changes and then let's merge it :money_with_wings:

davidleger95:
Should this PR be closed? It seems similar to the one @sslotsky submitted.

batgirl:
Essentially a duplicate of https://github.com/manifoldco/ui/pull/93 for testing purposes. Closing.

PR #104, opened 2019-04-27
fix cors issue maybe?

davidleger95:


dangodev:
What I’ve been doing to test locally is:

```
npm run build && cp -r dist ~/Sites/manifoldco/manifold-ui-react-demo/node_modules/@manifoldco/ui
```

If that runs then ship it

davidleger95:
sounds good, will test it out

PR #107, opened 2019-04-29
Resource Details

davejsdev:
## Reason for change

Screen Shot 2019-04-29 at 12 44 28 AM

Still going to keep adding to this. Just wanted to get a first pass up for any early feedback while I'm working on it.

**TODO**

- [x] Switch to using gateway endpoint.
- [x] Add cost display.
- [x] Loading state.
- [ ] Handle changes to inputs. *

_\* This probably isn't necessary because they can do it in the edit panel with the plan selector. Probably going to save this one for tomorrow._

**Tech debt for later**
- [ ] Refactor component so that network requests happen in a wrapper instead of the UI rendering component.
- [ ] Write thorough tests.


src/components/manifold-resource-details/style.css
```diff
@@ -1,8 +1,42 @@
...
+.heading {
+  grid-area: heading;
+  margin: 0;
+  color: var(--mf-c-gray);
+  font-size: var(--mf-font-d2);
+  text-transform: uppercase;
+}
+
+.details {
+  margin-right: 10vw;
```
davidleger95: layout is a bit weird. going to refactor and hopefully make it feel a bit more natural

dangodev:
Looking good so far! Just leaving a non-comment to say I don’t have any feedback.  👍

davidleger95:
Going to handle input changes in a different PR. Seems a bit more ambiguous

src/components/manifold-resource-details/manifold-resource-details.tsx
```diff
@@ -16,48 +17,94 @@ export class ManifoldResourceDetails {
...
-  async componentWillLoad() {
-    // get resource data
-    const resourceResponse = await fetch(
-      `${this.connection.marketplace}/resources/${this.resourceId}`,
-      withAuth()
-    );
-    const resource = await resourceResponse.json();
-    this.resource = resource;
+  fetchResource(resourceName: string) {
+    return fetch(`${this.connection.gateway}/resources/me/${resourceName}`, withAuth())
```
dangodev: One thing I just learned for the skeleton screens is when you start the fetch, it’s good to set the state to `undefined` so it can at least show the loading state. There are times where you change the page, but the old features still display because they’re set, just re-fetching in the background. Then they update a second or two later.

But here, adding `this.resource = undefined;` above the fetch will start the loading state and feel more fluid.

davidleger95: ahh, good idea. I was thinking of adding a loading flag or something, but that's a lot easier.

PR #113, opened 2019-05-03
Expose CSS variables

davidleger95:
## Reason for change

* Expose CSS vars for custom theming.

## Testing

* Ensure css vars persist in docs site.
* Ensure styling for docs site is unchanged.


scripts/docs.js
```diff
@@ -65,4 +65,45 @@ ${HTML_AFTER}`;
...
+    }
+    const css = readFileSync(filepath, 'utf8');
+
+    const newContents = `<!-- BEGIN CSS ${fileName} -->
+${CSS_BEFORE}
+${css}
+${CSS_AFTER}`;
+    withDocs = withDocs.replace(stylesheet, newContents);
+  });
+}
```
davidleger95: I automated things 🤖

src/global/app.css
```diff
@@ -4,324 +4,4 @@
...
-  box-sizing: border-box;
-  width: calc(100vw - 3rem);
-  max-width: 1200px;
-  max-height: calc(100vh - 3rem);
-  padding: 1rem;
-  overflow: auto;
-  -webkit-overflow-scrolling: touch;
-  background-color: white;
-}
+@import './theme.css';
```
davidleger95: This might not be good. I can copy the variables over and delete the theme file. What's your preference?

src/global/docs.css
```diff
@@ -0,0 +1,337 @@
...
+
+@media (min-width: 750px) {
+  .docs-nav {
+    top: 0;
+    left: 0;
+    width: 15rem;
+    height: 100vh;
+    padding: 1.5rem;
+  }
+}
```
davidleger95: This is what I was talking about. We don't get postCSS features when it's inlined. Media queries have to be at the top level. There was also some nested selectors that had to be refactored.

I tried inserting a postCSS process into the `docs.js` process but I didn't really know what I was doing so I gave up. Refactoring was much quicker.

src/global/app.css
dangodev: I’d rather copy this over. This likely won’t work in production (import path will be off, unless they are actually transforming & bundling imports). But even if it did work it’d be another network request (maybe, again, depending on how they’re handling CSS).

One file is simpler than multiple files + bundling instructions IMO.

src/global/docs.css
dangodev: Ohh I see now. Yeah this has to be transformed—you’re right. Welcome to 2008 lol

PR #119, opened 2019-05-13
Theme switcher

dangodev:
## Reason for change
We need to show how easy it is to switch themes. The themes persist across pageloads, too. 

This also adds a starter doc for themes, but probably needs more expansion there.

## Testing
The themes should work on the preview link.


docs/docs/components/manifold-service-card.md
```diff
@@ -0,0 +1,57 @@
...
+document.addEventListener('manifold-marketplace-click', { detail: { productLabel } } => {
+  alert(`You clicked the card for ${productLabel}`);
+});
+```
+
+The following events are emitted:
+
+| Event Name                   | Description                                     | Data                        |
+| :--------------------------- | :---------------------------------------------- | :-------------------------- |
+| `manifold-marketplace-click` | Fires whenever a user has clicked on a product. | `productId`, `productLabel` |
```
dangodev: Added Service Card component

docs/docs/theming.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Typography
+
+| Name                               | Default                             | Description                          |
+| :--------------------------------- | :---------------------------------- | :----------------------------------- |
+| `--manifold-font-family`           | (System)                            | Default text family                  |
+| `--manifold-font-family-monospace` | `IBM Plex Sans Monospace,monospace` | Code font family (credentials, etc.) |
+| `--manifold-text-color`            | `black`                             | Base text color                      |
+| `--manifold-text-color-body`       | `black`                             | Body text color                      |
+| `--manifold-text-color-accent`     | `white`                             | Accent text color                    |
```
dangodev: What is the difference between `text-color`, `text-color-body`, and `text-color-accent`? Could we find more specific terms for `-body` and `-accent`?

I see that accent is inverted—maybe we should have a `manifold-color-cta` and `manifold-color-cta-text`?

docs/docs/theming.md
```diff
@@ -0,0 +1,64 @@
...
+
+Card components include the **Service Card** and the left-hand side of **Plan Selector**.
+
+| Name                                | Default                        | Description       |
+| :---------------------------------- | :----------------------------- | :---------------- |
+| `--manifold-card-background-hover`  | `transparent`                  | Hover background  |
+| `--manifold-card-background-active` | `transparent`                  | Active background |
+| `--manifold-card-border-active`     | `1px solid rgb(30, 80, 218)`   | Active border     |
+| `--manifold-card-border-hover`      | `1px solid rgba(0, 0, 0, 0.1)` | Hover border      |
+| `--manifold-card-color-active`      | `rgb(30, 80, 218)`             | Active text color |
```
dangodev: Needs more expansion here, too

src/components/manifold-resource-status/manifold-resource-status.css
```diff
@@ -47,13 +47,13 @@
...
 
   &::before {
-    content: '';
     display: block;
     width: 1rem;
     height: 1rem;
     margin-right: 0.5rem;
     background: var(--status-color);
     border-radius: 50%;
+    content: '';
```
dangodev: All the changes in the CSS files were from updating the rational order plugin. I think it fixed a bug.

stencil.config.ts
```diff
@@ -6,7 +6,7 @@ import postCSSPresetEnv from 'postcss-preset-env';
 
 export const config: Config = {
   namespace: 'manifold',
-  globalStyle: 'src/global/theme.css',
+  globalStyle: 'src/global/app.css',
```
dangodev: This changes it back to `app.css` for simplicity.

docs/docs/getting-started.md
```diff
@@ -22,25 +22,44 @@ modern framework like React, Vue, or Angular.
...
+  <script src="https://unpkg.com/@manifoldco/ui/dist/manifold.js"></script>
+</body>
 ```
 
 ### React
 
 ```jsx
 import React from 'react';
 import ReactDOM from 'react-dom';
+import '@manifoldco/ui/dist/manifold.css';
```
dangodev: Add the CSS file to the instructions.

readme.md
```diff
@@ -79,17 +92,38 @@ To solve that, add the following to `tsconfig.json`:
...
+
+Initializing Manifold UI works the exact same as any other Stencil project.
+For more advanced instructions on integrating with your specific stack,
+please refer to Stencil’s [docs on integration][stencil-framework].
 
 [stencil]: https://stenciljs.com/
 [stencil-framework]: https://stenciljs.com/docs/overview
 [ts-include]: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html
+[ts-fix]: https://github.com/Microsoft/TypeScript/pull/26797
+[web-components]: https://www.webcomponents.org/introduction
```
dangodev: These changes just make the README here consistent with `getting-started.md`.

dangodev:
@davidleger95 is there a Manifold theme that adds the gradients, etc. back?

docs/docs/theming.md
davidleger95: `text-color` is just the generic text color that all others fall back to if they aren't defined.

`text-color-alt` is for things that are grayed out a bit more, like "NO" in the features table, "/mo" in prices, etc.

`text-color-accent` is to be used on anything that uses a darker color as the background. So buttons, tags, and maybe some other things. This one is a bit odd because sometimes the _darker_ color is too light and should use `text-color` or some other alternative.

docs/docs/theming.md
davidleger95: Yeah, I'll work on that today

docs/docs/theming.md
dangodev: I’d like to tie together the colors that interact with one another, to make it more clear. “Alt” IMO is super vague, and if we kept that name we’d have to list out what that means for every component. However, a name like `-subdue` is a bit more clear: _whatever the text color is, knocked down from that_. That color works both ways: dark and light theme.

src/global/app.css
```diff
@@ -141,4 +142,91 @@
...
+  --manifold-g-red: linear-gradient(to top right, #a65084, #ec7740);
+  --manifold-g-redFaded: linear-gradient(to top right, #f1b2ce, #ffcbb3);
+  --manifold-g-skyBlue: linear-gradient(-152deg, #388dc7 5%, #4e56a7 98%);
+  --manifold-g-yellow: linear-gradient(to top right, #ffb83a, #ffe268);
+  --manifold-g-yellowFaded: linear-gradient(to top right, #ffd995, #fff7d3);
+  --manifold-g-custom: linear-gradient(to top right, #eadeff, #f9f5ff);
+  --manifold-g-dumper: linear-gradient(to top right, #9b0000, #d2202f);
+  --manifold-g-blitline: linear-gradient(to top right, #00648c, #00a4e6);
+  --manifold-g-elegant-cm: linear-gradient(to top right, #0e517e, #218ab5);
+
```
davidleger95: Are these for the docs site? If this ships to npm, it's a lot of variables that are not needed used in the base theme.

The intention in removing them was that these colors were not used in the base theme. They're not used because gradients only work for backgrounds, which made it complicated to create a simple theme that could be applied to text color, borders, etc.

These variables should be in a Manifold theme that's only for us to use.

dangodev: Ah I see. OK I can add these as a Manifold theme 👍.

But FWIW this was all just in the `theme.css` file which was already shipping to npm (everything in `global/` ships)

stencil.config.ts
davidleger95: I think `theme` is clearer personally, or maybe `base-theme`, or even just `styles`. `app` seems odd to me because when platforms use our components, they're not using an app. Is there something I'm missing?

src/global/app.css
davidleger95: Awesome :) Yes, that made sense to ship these vars when we didn't have theming. Oh, and I'm realizing that these gradients are for products. those are already defined in the product page where they're used so the gradients can be removed altogether. the other ones like `--manifold-c-black-*`, and the rest of them, can be moved to a manifold theme.

docs/docs/theming.md
davidleger95: Yeah, I think that makes sense too. We can play around with it and see what works. What I've been aiming for with these high level vars is to allow easy changes to defaults and then, if further customization is needed, platforms can start redefining the lower level variables as needed. So in that case, `-alt` or `-subdue` should have a default based on the `text-color`. that way platforms can optionally define it, or just leave as a variant of `text-color`

davidleger95: This is something we should pick up in another ticket though since it doesn't involve the theme switcher.

stencil.config.ts
dangodev: Oh I didn’t have much of a preference; just was renaming back to Stencil defaults and thought this was cleanup leftover. I can change it back to `theme`—it works all the same

src/global/app.css
dangodev: Hm. Tried removing some of these, but these are indeed mostly-needed for the default theme.

docs/docs/theming.md
dangodev: Yeah! I agree. It’s just—once these names are set, and in use, then we can’t change them forever. I want to make sure the names make sense for everyone—us and platforms—for any theme, dark or light.

In general, I’ve been bitten by generic names like “accent” that carry more weight than they should. I.e., something that controls CTA color should control that, and nothing else.

docs/src/lib/theme-manifold.ts
```diff
@@ -0,0 +1,12 @@
...
+export default css`
+  body {
+    /* Button */
+    --manifold-button-background: var(--manifold-g-cta);
+
+    /* Tags */
+    --manifold-tag-radius: 1em;
+    --manifold-tag-free-background: var(--manifold-g-green);
+  }
+`;
```
davidleger95: I'm glad to see that you're using my abstraction as intended. I didn't create something that's completely incomprehensible 😄 🎉

dangodev: You certainly didn’t! The system itself is great!

src/global/app.css
davidleger95: weird, I'm not using them anywhere. I'll have to take a closer look at this. Either way, also out of scope for the theme switcher so that can be followup work.

PR #120, opened 2019-05-13
Fix homepage link highlight

dangodev:
## Reason for change
When on the homepage, **Getting Started** wasn’t highlighted. But if you clicked on it, it would highlight on the left, even though the page on the right stayed the same. This fixes that bug.

## Testing
Make sure **Getting Started** highlights on `/`. Also make sure the other active links work.

### DIFF
--- a/docs/docs/theming.md
+++ b/docs/docs/theming.md
@@ -14,7 +14,7 @@ However, there are times when it’s advantageous to overwrite some variables.
 In that usecase, CSS variables are recommended.
 
 We recommend setting variables on `body`, which will override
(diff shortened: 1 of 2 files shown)


docs/docs/theming.md
```diff
@@ -14,7 +14,7 @@ However, there are times when it’s advantageous to overwrite some variables.
 In that usecase, CSS variables are recommended.
 
 We recommend setting variables on `body`, which will override
-`@manifoldco/ui/dist/manifold.css` which are scoped to `:host`:
+`@manifoldco/ui/dist/manifold.css` which are scoped to `:root`:
```
dangodev: Minor typo I caught while looking at the site

PR #122, opened 2019-05-14
Dark Theme and Theme Docs

davidleger95:
## Reason for change

- Added dark theme
- Modified Gatsby setup so that base theme loads before sample themes.
- A start on theme docs (I think this still needs some work. I'll keep adding to it over time)

## Testing

- Apply Dark theme and ensure components render properly. Also test that other themes still work as expected.
- Go to theming section on docs site and check for accuracy and quality of docs.


docs/docs/theming.md
```diff
@@ -5,43 +5,86 @@ path: /theming
...
+
+
+
+```css
+manifold-marketplace {
+  --card-background: #fff;
+  --card-background-hover: #eee;
+  /* Other values go here */
+}
+``` -->
```
davidleger95: This isn't ready for use yet. A lot of the variables are there but not on top-level components, so they're not really accessible yet.

This is more for extremely specific styling cases. So maybe it's not even relevant.

docs/src/components/Example.tsx
```diff
@@ -39,7 +39,7 @@ interface ExampleProps {
 const Example: React.FunctionComponent<ExampleProps> = ({ html }) => (
   <Wrapper>
     <h3>Example</h3>
-    <Inner dangerouslySetInnerHTML={{ __html: html }} />
+    <Inner className="example-inner" dangerouslySetInnerHTML={{ __html: html }} />
```
davidleger95: had to add a class to change the bg color for dark theme

docs/gatsby-browser.js
```diff
@@ -4,4 +4,5 @@
  * See: https://www.gatsbyjs.org/docs/browser-apis/
  */
 
-// You can delete this file if you're not using it
+// Loads base theme before styled-components tag in head.
+import '@manifoldco/ui/dist/manifold.css';
```
davidleger95: Adding the base theme here forces it to come before styled-components

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
...
+
+### Global Colors
 
 | Name                         | Default             | Description                             |
-| :--------------------------- | :------------------ | :-------------------------------------- |
+|------------------------------|---------------------|-----------------------------------------|
 | `--manifold-background`      | `white`             | Default background color                |
-| `--manifold-color-primary`   | `rgb(30, 80, 218)`  | Primary CTA color                       |
-| `--manifold-color-secondary` | `rgb(15, 181, 208)` | Secondary CTA color                     |
+| `--manifold-color-primary`   | `rgb(30, 80, 218)`  | Primary brand color                     |
```
dangodev: Is this a “brand” color though? Not every design system will necessarily want to use their brand color for buttons. I think saying “this is for buttons” matters.

For example a brand may have a primary color of bright yellow, but they want their buttons to be yellow on black rather than black on yellow. I’m not really sure if there is a “primary” and “secondary” that’s universally agreed upon as to how it maps to a system, but I’d want to have “button bg color” and “button text color” and “heading color” as the main variables.

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
 
 # Theming
 
-These web components use [Shadow DOM][shadow-dom] for style encapsulation.
-That means styles can cascade down, but more importantly it means styles
+Manifold UI offers a simple, yet versitle, theming API for customizing the look and feel of components. 
+
+Web components use [Shadow DOM][shadow-dom] for style encapsulation.
+CSS selectors can't penetrate the shadow DOM, but more importantly styles
```
dangodev: You can delete this paragraph; you basically restate it in a different way.

Overall, let’s have more code examples, less writing!

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
...
+
+These grayscale bases provide the foundation for your theme. These two values, define a grayscale range. Intermediate values are then derived by varying the opacity of each base and used throughout Manifold UI components.
+
+By default, `--manifold-grayscale-[opacity]` values range from solid black down to transparent, and `--manifold-grayscale-[opacity]i` (notice the `i` for _inverted_) values range from solid white down to transparent. The main advantage to using these two contrasting grascales, is that a dark theme can be created simply by swapping these base values:
+
+```css
+:root {
+  --manifold-grayscale-base: 255, 255, 255;
+  --manifold-grayscale-base-inverted: 0, 0, 0;
+}
```
dangodev: IMO having “dark theme” and “light theme” variable sets right next to each other would make this section clear.

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
...
 
-However, there are times when it’s advantageous to overwrite some variables.
-In that usecase, CSS variables are recommended.
+### Base Theme
 
-We recommend setting variables on `body`, which will override
-`@manifoldco/ui/dist/manifold.css` which are scoped to `:root`:
+Manifold UI comes with a base theme that provides various levels of abstraction. This allows for easy theming at a high level for all components with just a small set of valiables, while also exposing lower level variables for more precise theming of individual components.
+
+We reccomend that you use this base theme as a starting point for building your own custom theme. To add the base theme to your project:
```
dangodev: `reccomend` (sp)

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
...
-[css-part]: https://meowni.ca/posts/part-theme-explainer/
-[shadow-dom]: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM
+### Common Styles
+
+| Name                      | Default                                   | Description                                         |
+|---------------------------|-------------------------------------------|-----------------------------------------------------|
+| `--manifold-radius`       | `4px`                                     | Default radius for buttons, tags, cards, textfields |
+| `--manifold-border`       | `1px solid var(--manifold-grayscale-10)`  | Default border for cards, text fields               |
+| `--manifold-border-focus` | `1px solid var(--manifold-color-primary)` | Default border for text fields on focus             |
+
```
dangodev: This should probably be in the main section or whatever. Maybe combine with typography?

IMO having a general  section would be great in addition to component-specific sections.

docs/docs/theming.md
davidleger95: yeah, the issue is that the primary color also affects borders and text color too. so Primary CTA color isn't accurate in that context. for just changing button colors, they should use `--manifold-button-background` which overrides the use of primary color for just the buttons.

docs/docs/theming.md
davidleger95: Like this?
```css
:root {
  /* light theme (default) */
  --manifold-grayscale-base: 0, 0, 0;
  --manifold-grayscale-base-inverted: 255, 255, 255;
  
  /* Dark theme */
  --manifold-grayscale-base: 255, 255, 255;
  --manifold-grayscale-base-inverted: 0, 0, 0;
}
```

docs/docs/theming.md
davidleger95: you mean combining the current sections (colors, typography, and common) into one table?

docs/docs/theming.md
davidleger95: Is there a way to get spellcheck on VS Code?

docs/docs/theming.md
```diff
@@ -5,60 +5,133 @@ path: /theming
 
 # Theming
```
davidleger95: what would be nice is a theme editor where devs can build a theme in the docs site and copy over to their project. #docsgoals

docs/docs/theming.md
davidleger95: sounds good

docs/docs/theming.md
dangodev: 🤷‍♂️

docs/docs/theming.md
dangodev: Yup! IMO that explains it in a few lines, and it’s copy+paste-able

docs/docs/theming.md
dangodev: > the issue is that the primary color also affects borders and text color too

Right—I’m trying to point out that that may be too far-reaching. Also maybe “CTA” color isn’t important, especially if we let the clients control slots and they use their own buttons. But the idea is still the same—I should be able to read a variable name, and have a really good expectation of what that would affect.

docs/docs/theming.md
dangodev: Just an idea! Feel free to push back. But maybe it’d be simpler to read _General / Component A / Component B / Component C_ than _Color / Typography / Borders / Spacing / Component A / Component B / Component C_, etc.

docs/gatsby-browser.js
dangodev: 💯 Nice find!!!!

docs/src/components/Sidebar.tsx
```diff
@@ -20,12 +20,14 @@ const DEFAULT = 'default';
 const MANIFOLD = 'manifold';
 const SAPPHIRE = 'sapphire';
 const RUBY = 'ruby';
+const DARK = 'dark';
```
dangodev: We’re using Pokémon games please call this ”Moon” (Pokémon Sun & Moon) 🌑

docs/docs/theming.md
davidleger95: true. I just moved this stuff to an API Reference section as well. That way the first part can stick to explaining how theming is implemented and the reference can be the detailed list of every available property (without all the conceptual stuff getting in the way).

docs/src/components/Sidebar.tsx
davidleger95: will do :)

docs/docs/theming.md
```diff
@@ -5,60 +5,187 @@ path: /theming
...
-That means styles can cascade down, but more importantly it means styles
-declared in the components won’t leak out. This has the benefit of not
-conflicting with any of your existing styles.
+Manifold UI offers a simple, yet versitle, theming API for customizing the look and feel of components. 
 
-However, there are times when it’s advantageous to overwrite some variables.
-In that usecase, CSS variables are recommended.
+<manifold-toast alert-type="warning">
+  Theming is still a work-in-progress. This documentation may not be entirely accurate.
+</manifold-toast>
```
davidleger95: because we can 🙃

Maybe something like this would be useful for a long-tern solution for rich content: https://www.gatsbyjs.org/packages/gatsby-remark-custom-blocks/

PR #125, opened 2019-05-16
Let consumers use their own buttons

dangodev:
#### Old way
1. We generate the buttons
2. We control styling for said buttons, and only allow users to style certain aspects of buttons
3. We emit events for said buttons
4. We write documentation for buttons
5. We require users write more code to handle our buttons
6. We manage custom props for these buttons

#### New way
1. We let users use their own buttons—styling, events, everything
2. …that’s it. Less code for us to manage. We’re happier, consumers are happier.

## Testing
Screen Shot 2019-05-16 at 11 09 24


dangodev:
@davidleger95 note that this will get rid of all the custom button styling. Now the only button we provide is `<manifold-data-provision-button>`, but that doesn’t have shadow DOM or any styles so it’s fully controllable.

src/components/manifold-product/manifold-product.tsx
```diff
@@ -42,13 +37,9 @@ export class ManifoldProduct {
...
     return (
-      <manifold-product-page
-        hideCta={this.hideCta}
-        linkFormat={this.linkFormat}
-        preserveEvent={this.preserveEvent}
-        product={this.product}
-        provider={this.provider}
-      />
+      <manifold-product-page product={this.product} provider={this.provider}>
+        <slot />
```
dangodev: So this was the only concession: I know we agreed on named slots, and I liked that better, but because it was being passed to child components, I couldn’t get named slots to work. I think named slots only work at the top level.

We can revisit this, but I think we’d need to combine certain components into one to take advantage of named slots (which may not be a bad idea in general—if we can abstract some things neatly).

dangodev: For clarity: if you specify no slot, then the slot just appears before the web component.

If you specify named slots all the way down, then nothing appears at all. It’s as if the named slot gets lost trying to be passed down from parent to child.

sslotsky: I'm pretty sure I figured out a way to forward named slots onto children. I'll try to provide an example soon.

dangodev: Feel free to try it in this PR if it’s easier

sslotsky: Confirmed, you just need a wrapper element to pull it off. Maybe a little unfortunate, but workable.

Grandchild:

```html
<div>
  <slot name="my-slot-name" />
</div>
```

Child:

```html
<my-grandchild>
  <span slot="my-slot-name"><slot name="my-slot-name" /></span>
</my-grandchild>
```

Parent:

```html
<my-child>
  <button slot="my-slot-name">This goes into the grandchild</button>
</my-child>
```

I don't have a strong opinion on whether you should always used named slots. It's an obvious need when there's more than one slot, but I could see arguments either way if there's only one. Up to you. But do you think you should update the storybook story to add some buttons?

sslotsky: I guess one good argument in favor of named slots is that it's a bit more future-proof. If you add slots in later versions, you're not breaking anything i.e. forcing consumers to name their slot content.

At any rate, it sure would be a lot cooler if you could do this instead:

```html
<slot slot="my-slot-name" />
```

Seems intuitively like that should work, but it doesn't :man_shrugging:

docs/docs/components/manifold-product.md
```diff
@@ -3,6 +3,7 @@ title: Product
 path: /components/product
 example: |
   <manifold-product product-label="jawsdb-mysql">
+    <manifold-link-button>Get JawsDB MySQL</manifold-link-button>
```
sslotsky: Missed this, needs to be updated with the slot name.

PR #128, opened 2019-05-17
Speed up Netlify builds by skipping Gatsby install

dangodev:
## Reason for change
`postinstall` was originally for developing locally—using Storybook & Gatsby. But with Netlify using Storybook, this will speed things up.

Also, there have been many instances locally this slows down development when installing something for Storybook not needed for Gatsby.


dangodev:
Closing; this isn’t really slowing down Netlify that much and it creates other issues.

PR #130, opened 2019-05-20
Add more screenshot tests

sslotsky:
## Reason for change

Continuing VRT work for UI. This adds a screenshot for:

* The plan selector with JawsDB, using JSON retrieved from the API as mock data for the component.
* The skeleton screen for the marketplace grid 
* Marketplace grid with services

To get the screenshot to work for the marketplace with services, I had to change `skeleton` to `false` by default, and switch it to true when we first try to display services if we detect that none have been given.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6625,8 +6625,7 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -6650,15 +6649,13 @@
           "version": "1.0.0",
           "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
           "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
           "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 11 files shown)


package-lock.json
```diff
@@ -6625,8 +6625,7 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true,
-          "optional": true
```
dangodev: Is this an npm version mismatch? I’m on `6.9.0`

src/components/manifold-marketplace-grid/manifold-marketplace-grid-happo.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  document.body.appendChild(grid);
+};
+
+export const allProducts = () => {
+  const grid = document.createElement('manifold-marketplace-grid');
+  grid.services = fromJSON(products);
+
+  document.body.appendChild(grid);
+};
```
dangodev: These seem to ship to npm under `dist/components/*-happo.js`. It generates types for these, too. Is that desirable to ship our Happo tests to DO, etc.?

I’m not aware of any problems it’d cause, other than a few extra KB, but just wanted to pose the question. I.e., would storing these outside `src/components/` be cleaner for npm?

package-lock.json
sslotsky: Oh, probably. I was on an old version until I realized that I had a test failing locally. I can upgrade and re-run the install script and see if that helps?

src/components/manifold-marketplace-grid/manifold-marketplace-grid-happo.ts
sslotsky: Maybe! How do the `.spec` and `.e2e` tests get excluded?

src/spec/mock/fromJSON.ts
```diff
@@ -0,0 +1,3 @@
+export default function fromJSON(json: object) {
+  return JSON.parse(JSON.stringify(json));
+}
```
dangodev: I think this mock data is also shipping to npm, since it’s required by the Happo tests (the mock data from our test suite doesn’t seem to, though). This would be good not to ship to npm.

package-lock.json
dangodev: Oh maybe. Not a big deal; just curious

sslotsky: Eh, that didn't really seem to help. And I don't think it matters a whole lot either but if you have suggestions I'm open.

src/components/manifold-marketplace-grid/manifold-marketplace-grid-happo.ts
sslotsky: Looks like there is an option to exclude some patterns

https://stenciljs.com/docs/config#excludesrc

PR #131, opened 2019-05-21
Theming Docs Refinements

davidleger95:
## Reason for change

Adding missing docs for theming

## Testing

- Review theming docs page for accuracy and completeness

### DIFF
--- a/docs/docs/theming.md
+++ b/docs/docs/theming.md
@@ -67,7 +67,7 @@ manifold-marketplace {
 ### Grayscales & Dark Theming
 
 | Name                                 | Default         | Description                                        |
-|--------------------------------------|-----------------|----------------------------------------------------|
+| ------------------------------------ | --------------- | -------------------------------------------------- |
 | `--manifold-grayscale-base`          | `0, 0, 0`       | The darkest color on your grayscale in RGB values  |
 | `--manifold-grayscale-base-inverted` | `255, 255, 255` | The lightest color on your grayscale in RGB values |
 
@@ -98,7 +98,7 @@ By default, `--manifold-grayscale-[opacity]` values range from solid black down 
 There are four theming levels:
 
 | Level       | Name              | Description                                                          |
-|-------------|-------------------|----------------------------------------------------------------------|
+| ----------- | ----------------- | -------------------------------------------------------------------- |
 | 1 (highest) | Base              | The base theme provided by Manifold UI                               |
 | 2           | Top-level         | Your custom theme and branding                                       |
 | 3           | Common Components | Reusable UI elements like cards, buttons, and tags                   |
@@ -162,14 +162,14 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
 ### Grayscale
 
 | Name                                 | Default         | Description                                        |
-|--------------------------------------|-----------------|----------------------------------------------------|
+| ------------------------------------ | --------------- | -------------------------------------------------- |
 | `--manifold-grayscale-base`          | `0, 0, 0`       | The darkest color on your grayscale in RGB values  |
 | `--manifold-grayscale-base-inverted` | `255, 255, 255` | The lightest color on your grayscale in RGB values |
 
 ### Top Level Properties
 
 | Name                                         | Default                                   | Description                                         |
-|----------------------------------------------|-------------------------------------------|-----------------------------------------------------|
+| -------------------------------------------- | ----------------------------------------- | --------------------------------------------------- |
 | `--manifold-color-primary`                   | `rgb(30, 80, 218)`                        | Primary color                                       |
 | `--manifold-color-secondary`                 | `rgb(15, 181, 208)`                       | Secondary color                                     |
 | `--manifold-color-info`                      | `rgb(30, 80, 218)`                        | General info message color<sup>\*</sup>             |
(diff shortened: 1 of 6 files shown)


docs/docs/theming.md
```diff
@@ -168,75 +198,105 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
...
 
 #### Plan Selector
 
-TODO
+Coming Soon!
 
 #### Product
 
-TODO
+Coming Soon!
```
davidleger95: I can keep spec-ing these out as well. Are there any others I should include for Unique components? (maybe resource details, but I don't think we even have docs for that yet)

Another thought: perhaps we should duplicate or link between theming and the individual components' docs for the "Unique Component Properties", since it's relevant to both.

docs/docs/theming.md
```diff
@@ -168,75 +196,128 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
...
+| <span class="tag not-implemented" /> `--manifold-text-weight-heading` | `700`                            | Weight for all headings                                                                                  |
+| <span class="tag not-implemented" /> `--manifold-font-size-h1`        | `--`                             | Level 1 heading font size.                                                                               |
+| <span class="tag not-implemented" /> `--manifold-font-size-h2`        | `--`                             | Level 2 heading font size.                                                                               |
+| <span class="tag not-implemented" /> `--manifold-font-size-h3`        | `--`                             | Level 3 heading font size.                                                                               |
+| <span class="tag not-implemented" /> `--manifold-font-size-h4`        | `--`                             | Level 4 heading font size.                                                                               |
```
dangodev: For these, I think that’s a decision we can come back to later. We should keep it so that consumers can set the base size, but with responsive sizes we’re already doing this makes it hard.

Right now we have headings at appropriate sizes based on the base size, so as long as that’s in place (it is), maybe we can come back to this if DO demands it?

docs/docs/theming.md
```diff
@@ -168,75 +196,128 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
...
-| `--manifold-border`                | `1px solid var(--manifold-grayscale-10)`  | Default border for cards, text fields               |
-| `--manifold-border-active`         | `1px solid var(--manifold-color-primary)` | Default border for text fields on focus             |
-| `--manifold-border-focus`          | `1px solid var(--manifold-color-primary)` | Default border for text fields on focus             |
+| Name                                                            | Default                                   | Description                                         |
+| --------------------------------------------------------------- | ----------------------------------------- | --------------------------------------------------- |
+| `--manifold-color-primary`                                      | `rgb(30, 80, 218)`                        | Primary color                                       |
+| `--manifold-color-secondary`                                    | `rgb(15, 181, 208)`                       | Secondary color                                     |
```
dangodev: I’m still confused about these two—maybe examples will help?

Keep in mind that we now let consumers use their own buttons, so maybe this should just go away (any properties we can drop, we probably should)

docs/docs/theming.md
```diff
@@ -168,75 +196,128 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
...
+| `--manifold-color-info`                                         | `rgb(30, 80, 218)`                        | General info message color<sup>\*</sup>             |
+| `--manifold-color-success`                                      | `rgb(49, 186, 162)`                       | Success color<sup>\*</sup>                          |
+| `--manifold-color-warn`                                         | `rgb(254, 113, 75)`                       | Warning color<sup>\*</sup>                          |
+| `--manifold-color-error`                                        | `rgb(213, 15, 73)`                        | Error color<sup>\*</sup>                            |
+| `--manifold-font-family`                                        | (System)                                  | Default text family                                 |
+| `--manifold-font-family-monospace`                              | `IBM Plex Sans Monospace,monospace`       | Code font family (credentials, etc.)                |
+| `--manifold-text-color`                                         | `var(--manifold-grayscale-100)`           | Base text color                                     |
```
dangodev: We have `color-info` and `text-color`. Should these follow the same pattern? i.e., is color a function of component, or vice-versa?

Sorry—that’s an unanswerable question. Maybe we should just do `color-*` just to make things easier to remember.

docs/docs/theming.md
```diff
@@ -3,6 +3,31 @@ title: Theming
...
+    background: var(--manifold-g-yellow);
+    content: 'P'
+  }
+
+  .not-implemented::before {
+    background: var(--manifold-g-red);
+    color: white;
+    content: 'N'
+  }
+</style>
```
dangodev: We can move this to the Example component, at least.

docs/docs/theming.md
davidleger95: The reason why I initially went with `text-color` was because other text properties are `text-*` (except `color` because why should it be consistent?), so if `color` what actually named correctly it would probably be `text-color`. Keep in mind that there is also `text-color-heading` which would become `text-heading-color` (or why not `heading-text-color` 🙃).

I'm fine with changing it though. Just wanted to give full context to the decision. Does that change your perspective at all or does your original view still hold?

docs/docs/theming.md
davidleger95: yeah, secondary is really only used for "Free" tags at the moment. so it can probably go away. We can just default all tags to one color and then overrides can happen lower down

docs/docs/theming.md
davidleger95: so remove the h1-4 size properties here then?

docs/src/components/Page.tsx
```diff
@@ -23,6 +23,31 @@ function Page({ currentPage, links }: PageProps) {
...
+    content: 'P';
+  }
+
+  .not-implemented::before {
+    color: white;
+    background: var(--manifold-g-red);
+    content: 'N';
+  }
+`;
+
```
davidleger95: Wasn't sure exactly where to move this. Is here good, or do you want it somewhere else?

dangodev: That works! And I love the badges, BTW! Forgot to mention it. Very nice!

docs/docs/theming.md
dangodev: Yeah that’s fair. There are a lot of `text-color-*`s. In that case though maybe move it out of the top area, and down below with the other `text-color-*`s.

PR #132, opened 2019-05-22
Replace local lint rules with published version

sslotsky:
## Reason for change

We have published our lint rules publicly, so this PR replaces the local version with the published one.


.eslintrc.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     'prettier',
     'prettier/@typescript-eslint',
   ],
-  plugins: ['@typescript-eslint', 'prettier', 'manifoldco-ui'],
+  plugins: ['@typescript-eslint', 'prettier', '@manifoldco/stencil'],
```
dangodev: Ooh nice! I forgot ESLint removes `eslint-plugin` from the scoped part

PR #133, opened 2019-05-23
Fix IntersectionObserver in Chrome

dangodev:
Fixes manifoldco/engineering#8239.

This fixes a bug where on page load, Chrome wouldn’t load the images in view. It would only load on scroll.

## Reason for change
**Before**
Screen Shot 2019-05-23 at 11 25 30

**After**
Screen Shot 2019-05-23 at 11 40 02

## Testing
Preview the Storybook link in Chrome (not Gatsby). The images have to be in the initial viewport to break (the Gatsby docs require you to scroll, so you won’t notice it).

### DIFF
--- a/src/components/manifold-lazy-image/manifold-lazy-image.tsx
+++ b/src/components/manifold-lazy-image/manifold-lazy-image.tsx
@@ -8,29 +8,25 @@ export class ManifoldLazyImage {
   @State() observer: IntersectionObserver;
 
   componentWillLoad() {
-    this.observer = new IntersectionObserver(this.observe, {
-      rootMargin: '0px 0px 50px 0px',
-      threshold: 0,
-    });
+    this.observer = new IntersectionObserver(this.observe);
   }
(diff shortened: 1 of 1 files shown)


src/components/manifold-lazy-image/manifold-lazy-image.tsx
```diff
@@ -8,29 +8,25 @@ export class ManifoldLazyImage {
...
-    });
+    this.observer = new IntersectionObserver(this.observe);
   }
 
   private observe = (
     entries: IntersectionObserverEntry[],
     observer: IntersectionObserver
   ): void => {
-    const entry = entries[0];
+    const entry = entries.find(({ isIntersecting }) => isIntersecting === true);
```
dangodev: My goal here was to only find the first entry where it’s intersecting, and then stop scanning the array. Probably a tic slower than before, but at least it can still return early if the first item is what it wants.

PR #134, opened 2019-05-23
VRT for the credentials view

sslotsky:
## Reason for change

VRT for the resource credentials view. Also includes a refactor that accomplishes a few things:

1. Makes VRT possible by separating the view from the fetching logic
2. Pulls fetching into a `manifold-resource-container` that will allow DO to share resource data across multiple views (explained further below)
3. Changes our link-button into a real button, because it should be a real button
4. Adds an auth token into local storage for the resource credentials storybook story

### Sup with this refactor?

The screenshot here shows the "resource details" page for DO. There are at least three components that use the same resource data here: the header, the plan selector, and the credentials view. And there are a couple things we should avoid doing:

1. We shouldn't write a composite view that dictates the layout of these components.
2. We shouldn't ask all of those components to fetch the same data from the server.

That means that they should be sharing resource state. In Stencil this is typically done with state tunnels. So I decided that for now, `manifold-resource-credentials` will have to be rendered inside of a `<manifold-resource-container>` in order to function properly. The storybook story was adjusted to account for this.

image (6)


src/components.d.ts
```diff
@@ -10,6 +10,9 @@ import { JSX } from '@stencil/core';
 import {
   Connection,
 } from './utils/connections';
+import {
+  State,
```
sslotsky: I should rename this, or it will cause collisions down the road as we use this pattern more.

src/components.d.ts
```diff
@@ -24,6 +27,13 @@ export namespace Components {
     'selectedResource'?: Gateway.Resource;
   }
   interface ManifoldBadge {}
+  interface ManifoldButton {
+    'color'?: 'black' | 'white';
+    'disabled': boolean;
+    'onClickEvent'?: (e: MouseEvent) => void;
+    'size'?: 'small';
+    'wait': boolean;
```
sslotsky: Oops, I don't need this `wait` prop.

src/components/manifold-button/button.css
```diff
@@ -17,7 +17,7 @@
   --shadow-hover: var(--manifold-button-shadow-hover, var(--shadow));
 }
 
-a {
+button {
```
davidleger95: Are we using this button anywhere?

src/components/manifold-resource-credentials-view/manifold-resource-credentials-view.tsx
```diff
@@ -0,0 +1,88 @@
...
+    this.credentialsRequested.emit(this.resourceState.resource);
+  };
+
+  render() {
+    return [
+      <menu class="secrets-menu" data-showing={!!this.credentials}>
+        <manifold-button color="white" size="small" onClickEvent={this.hideCredentials}>
+          <manifold-icon marginRight icon={lock} />
+          Hide credentials
+        </manifold-button>
```
davidleger95: oh, here it is 🙃

stories/manifold-resource-credentials.stories.js
```diff
@@ -1,9 +1,15 @@
 import { storiesOf } from '@storybook/html';
 import markdown from '../docs/docs/components/manifold-resource-credentials.md';
 
+const credentialsView = `
+<manifold-resource-container resource-label="cms-stage">
+  <manifold-resource-credentials></manifold-resource-credentials>
+</manifold-resource-container>
```
davidleger95: So are we now expecting platforms to have both the container and presentational components?

sslotsky: At the moment, we have:

* manifold-resource-container
* manifold-resource-credentials
* manifold-resource-credentials-view

I doubt that they'll use the third one themselves. But based on the designs, I do expect `manifold-resource-credentials` to be used in a view with other components that are displaying data from the same resource, so `manifold-resource-container` is how they would share that data.

Should we have a version of the component that does the fetching without help of `manifold-resource-container`? At this point I doubt that it's needed. Definitely interested in your thoughts, and will probably make an ADR about this decision.

davidleger95: Yeah, I think that's fine the way you have it then. First ADR! 🙌

PR #135, opened 2019-05-23
top-level and cards theme refinements

davidleger95:
## Reason for change

- Add border state variables
- Refactor Cards theming
- Add white outline to logos to ensure they're always properly contrasted for all themes.

## Testing

Check theming still works as expected for all components on all themes

### DIFF
--- a/docs/docs/theming.md
+++ b/docs/docs/theming.md
@@ -171,20 +171,24 @@ Some variables that are for a specific state `*-hover`, `*-focus`, `*-active` wi
 
 ### Top Level Properties
 
-| Name                                                            | Default                                   | Description                                         |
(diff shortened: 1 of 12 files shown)


davidleger95:
bleh, my stylelint broke down earlier.  🔧

sslotsky:
I thought the Manifold theme was still using gradient buttons? I can't tell the difference currently between the Manifold theme and the Default theme in the plan selector.

image

davidleger95:
> I thought the Manifold theme was still using gradient buttons?

Oops, I got distracted moving some theme variables and forgot to paste them after cutting them 😂

davidleger95:
It's fixed now @sslotsky :)

PR #136, opened 2019-05-23
Adding toast screenshots

batgirl:
## Reason for change
Adding VRT to the toast component for each alert type.
Examples of the test are visible here: https://happo.io/a/67/p/129/report/dev-4bd740f5e331dfdd210d#manifold-toast-text

## Testing
This can be tested by running `npm run happo dev` locally.


src/components/manifold-toast/manifold-toast-happo.ts
```diff
@@ -0,0 +1,34 @@
+export const basic = () => {
+    const plan = document.createElement('manifold-toast');
```
davidleger95: Should these `plan`s be `toast`s? 🍞

sslotsky: Mmmmmm, copy pasta!

batgirl:
Before seeing the suggestions this morning (😩 sorry) I added text to each of the screenshots (in case fonts change for some reason), but I ran into what I can only assume was async weirdness because each time I ran the test one or two of the toasts wouldn't be fully rendered. The current configuration with adding an `asyncTimeout` to the `.happo.js` had all of the toasts rendering. Have you experienced this Sam??

Also I do have prettier in my setup, I had originally had it formatting on save thanks to Drew I believe.. I'll take a look at that. I ran it locally and committed the results in the meantime.

And.. copy-pasta guilty. 😅

davidleger95:
probably fine, but the bottom border seems to be cut off a bit in the screenshots.

batgirl:
I added my `prettier` linter to VSCode and `src/components.d.ts` saw some changes after I ran the linter on it - open to feedback about if that is good or not!

batgirl:
Thank you for the help David! Sam and I ended up needing to add a wrapper to get around the timeout problem, and I think it also fixed the line at the bottom. 👍

sslotsky:
For `components.d.ts`, on one hand it might be no harm done, but probably preferable to avoid formatting changes to it. For my setup, I would only expect that to happen if I saved the file in my editor. Is that what happened in your case, or did something else cause it to format?

My guess is that it well reformat itself next time someone adds/removes/changes a component property. if it's easy to revert, go ahead and do it, but if not, I doubt it makes much difference. This PR does need a rebase though. Let me know if you need help with that at all. Great work!

dangodev:
Yeah what @sslotsky said—I’d recommend running `npm run build` on this PR, and committing the changes to `src/components.d.ts`

PR #143, opened 2019-05-24
lower level tags theming (medium-small PR)

davidleger95:
## Reason for change

- Added more tag-specific theming options
- Updated tag default values (`text-transform: initial;`, `font-weight: 700;`)
- Removed reduced opacity on plan selector costs because it was being applied to the tags as well.

## Testing

Check that tags theming works as expected for each theme.


src/components/manifold-service-card/manifold-service-card.css
```diff
@@ -156,11 +159,11 @@
   margin-left: 0.5em;
   padding: var(--tag-padding);
   color: var(--tag-text-color);
+  font-family: var(--tag-font-family);
   font-weight: var(--tag-font-weight);
-  font-size: var(--manifold-font-d3);
```
dangodev: I’m -1 for this change, because currently this allowed for responsive text resizing, based on the component. We do utilize some of that in the service cards. If we change this value, we’d also have to expose responsive breakpoints and text sizes for each component, which is a lot.

In the interest of wrapping up theming, maybe this can stay as-is for now? It’s a function of the base font size, and IMO it’s a sensible value based on the placement & position of the component.

src/components/manifold-service-card/manifold-service-card.css
```diff
@@ -156,11 +159,11 @@
...
+  font-size: var(--tag-font-size);
   line-height: 1.5;
-  letter-spacing: 0.0625em;
-  text-transform: uppercase;
```
dangodev: It is smart to change this, as `text-transform` does go hand-in-hand with the `font-size` (if it’s lowercase, the font would have to be bigger to compensate). But it also introduces the problems mentioned above around responsive resizing.

src/components/manifold-service-card/manifold-service-card.css
```diff
@@ -156,11 +159,11 @@
   margin-left: 0.5em;
   padding: var(--tag-padding);
   color: var(--tag-text-color);
+  font-family: var(--tag-font-family);
   font-weight: var(--tag-font-weight);
-  font-size: var(--manifold-font-d3);
+  font-size: var(--tag-font-size);
   line-height: 1.5;
-  letter-spacing: 0.0625em;
```
dangodev: This is a departure from our design. This does help readability IMO

src/components/manifold-badge/mf-badge.css
```diff
@@ -1,15 +1,23 @@
...
   display: inline-flex;
   align-items: center;
-  padding: var(--manifold-tag-padding, 0.25em 1em);
-  color: var(--manifold-tag-text-color, var(--manifold-text-color-accent));
-  font-weight: var(--manifold-tag-font-weight, 900);
-  font-size: var(--manifold-font-d2);
+  padding: var(--padding);
+  color: var(--text-color);
+  font-weight: var(--font-weight);
+  font-size: var(--font-size);
```
dangodev: Another instance where we’re losing the ability to control typography responsively based on base size. IMO I’d rather just manage this for them, keep variables to a minimum, and wait until DO requests a change.

src/components/manifold-badge/mf-badge.css
```diff
@@ -1,15 +1,23 @@
 :host {
+  --padding: var(--manifold-tag-padding, 0.25em 1em);
+  --text-color: var(--manifold-tag-text-color, var(--manifold-text-color-accent));
+  --font-family: var(--manifold-tag-font-family, var(--manifold-font-family));
+  --font-weight: var(--manifold-tag-font-weight, 700);
+  --font-size: var(--manifold-tag-font-size, var(--manifold-font-d2));
```
dangodev: Ah, sorry—after all the comments I just noticed this 😓 . I thought it meant we’d have to manage all these different sizes. If it falls back to the original type scale then I’m fine with that!

PR #144, opened 2019-05-24
Adding cost-display screenshots

batgirl:
## Reason for change
This adds visual regression tests for the cost-display component.
Can be viewed here: https://happo.io/a/67/p/129/report/dev-65ed8e64281dc750c0f8#manifold-cost-display

## Testing
You can run the happo test locally with `npm run happo dev`.


dangodev:
Hm. I don’t think we need VRT for this? We have extensive unit tests already for this that test all the permutations here, and then some. Seeing as this is just text, I think that just this VRT component is redundant and we should keep the testing for this inside `manifold-cost-display.e2e.ts`. I’m worried that having 2 tests look for the same thing just becomes a time-sink.

sslotsky:
I added this to our checklist probably not really noticing that it was just text. I'm fine with skipping it.

PR #145, opened 2019-05-24
Adding toggle screenshots

batgirl:
## Reason for change
Adds tests for disabled, checked, labeled and default toggle buttons.
Tests visible here: https://happo.io/a/67/p/129/report/dev-5ae7dc260dae5d035797#manifold-toggle

## Testing
Test can be run locally with `npm run happo dev`.

### DIFF
--- a/src/components/manifold-toggle/manifold-toggle-happo.ts
+++ b/src/components/manifold-toggle/manifold-toggle-happo.ts
@@ -3,3 +3,24 @@ export const primary = () => {
 
   document.body.appendChild(toggle);
 };
+
+export const labeled = () => {
+  const toggle = document.createElement('manifold-toggle');
+  toggle.label = 'Happo Test Label';
+
+  document.body.appendChild(toggle);
+};
+
+export const disabled = () => {
+  const toggle = document.createElement('manifold-toggle');
+  toggle.disabled = true;
+
+  document.body.appendChild(toggle);
+};
+
+export const checked = () => {
+  const toggle = document.createElement('manifold-toggle');
+  toggle.defaultValue = true;
+
+  document.body.appendChild(toggle);
+};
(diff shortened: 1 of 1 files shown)


src/components/manifold-toggle/manifold-toggle-happo.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  document.body.appendChild(toggle);
+};
+
+export const checked = () => {
+  const toggle = document.createElement('manifold-toggle');
+  toggle.defaultValue = true;
+
+  document.body.appendChild(toggle);
+};
```
davidleger95: Do we need screenshot tests for these lower level components? There was a discussion a little while back between @dangodev and @sslotsky about this. I thought we determined that just the top-level (documented) components needed screenshot tests because they would cover all child components. Or did we change our mind on that?

davidleger95: Any thoughts on this @dangodev @sslotsky?

sslotsky: I think it kind of depends. If a higher level component covers it, then oftentimes it will be fine, but if there are a lot of states that the lower level component can display, it might be best to cover them directly. I haven't looked into this one specifically.

davidleger95: Good point, I think that makes sense.

I'm curious what Drew's thoughts are since I think he was originally suggesting to just do high level screenshot tests. I think he would agree with your point as well, but maybe he has other concerns?

dangodev: Agree with @sslotsky. We should just avoid having the same screenshot over and over again in different tests. But for this, I think it makes sense. Higher-level tests the arrangement; lower-level tests all the states. I like that setup rather than attempting to test all possible states of all sub-components at the top-level.

We should avoid duplicating VRTs, but I don’t see this as a duplicate test since it’s testing other states.

PR #151, opened 2019-05-27
lower-level text theming

davidleger95:
## Reason for change

- Added theming for headings
- Fixed buttons in docs examples

## Testing

Check that headings behave as expected for all themes


src/components/manifold-service-card/manifold-service-card.css
```diff
@@ -138,9 +143,10 @@
   margin-top: 0;
   margin-bottom: 0;
   padding-right: 1em;
-  color: var(--text-color);
-  font-weight: 400;
+  color: var(--card-text-color-heading);
+  font-weight: var(--card-font-weight-heading);
```
dangodev: Hm. Maybe `card-heading-font-weight` would make more sense? Then it’s `component-subcomponent-property-subproperty`.

🤔 could it just be `card-heading-weight` and `card-heading-color` instead?

davidleger95: yeah, I was thinking that when I was writing it. I was trying to stay consistent with `--text-color-heading` where `heading` is sort of a modifier. But I think you're right, at this level, the subcomponent perspective makes more sense.

tl;dr: i'll change it

PR #152, opened 2019-05-27
implement lower-level theming for input fields

davidleger95:
## Reason for change

- Theming for input fields (text and select elements)

## Testing

check the default and focus states for all themes work as expected

### DIFF
--- a/docs/docs/theming.md
+++ b/docs/docs/theming.md
@@ -70,7 +70,7 @@ manifold-marketplace {
 ### Grayscales & Dark Theming
 
 | Name                                 | Default         | Description                                        |
-|--------------------------------------|-----------------|----------------------------------------------------|
+| ------------------------------------ | --------------- | -------------------------------------------------- |
 | `--manifold-grayscale-base`          | `0, 0, 0`       | The darkest color on your grayscale in RGB values  |
 | `--manifold-grayscale-base-inverted` | `255, 255, 255` | The lightest color on your grayscale in RGB values |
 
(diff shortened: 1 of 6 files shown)


docs/docs/theming.md
```diff
@@ -226,25 +226,22 @@ Cards are used in Marketplace (service cards) and Plan Selector (plan buttons, p
...
+| ----------------------------------- | ---------------------------------------- | ---------------- |
+| `--manifold-input-background`       | `var(--manifold-grayscale-100i)`         | Background       |
+| `--manifold-input-text-color`       | `var(--manifold-text-color)`             | Text color       |
+| `--manifold-input-border`           | `var(--manifold-border)`                 | Border           |
+| `--manifold-input-radius`           | `var(--manifold-radius)`                 | Radius           |
+| `--manifold-input-shadow`           | `none`                                   | Shadow           |
+|                                     |                                          |                  |
+| `--manifold-input-background-focus` | `var(--manifold-input-background-hover)` | Background:focus |
+| `--manifold-input-border-focus`     | `var(--manifold-input-border-focus)`     | Border:focus     |
+| `--manifold-input-shadow-focus`     | `var(--manifold-input-shadow-hover)`     | Shadow:focus     |
```
dangodev: Same question here with the other one: does `component-state-property` make sense? I kinda like state at the end as it is here but wanted to ask for consistency.

So right now we have `component-subcomponent-property-subproperty-state` as the formula?

davidleger95: It could make sense that way. I think I like it at the end though because that's how the pseudo selectors work in CSS.

davidleger95: Actually... the property should come after the selector, so yeah, you're right. Maybe it should be changed?

I wonder if this could be an opportunity for better name formatting?

What do you think of 
- `--manifold-input--focus--background` or 
- `--manifoldInput-focus-background`?

davidleger95: This will involve a lot of find-and-replace work though 😓

PR #162, opened 2019-05-28
Update paths to css file in readme

trotzig:
I was playing around with this project and noticed that a URL returned a 404.


trotzig:
Just came across 5fb16758, looks like the path will be right as soon as the next release is available on npm. 👍

PR #168, opened 2019-05-30
Adding VRT docs

batgirl:
## Reason for change
Adding documentation for Happo checks and writing Happo tests.

## Testing
See if they make sense! :D

### DIFF
--- a/src/components/manifold-select/manifold-select-happo.ts
+++ b/src/components/manifold-select/manifold-select-happo.ts
@@ -23,4 +23,6 @@ export const primary = () => {
   select.name = name;
 
   document.body.appendChild(select);
+
+  return select.componentOnReady();
 };
--- a/src/readme.md
+++ b/src/readme.md
@@ -49,7 +49,7 @@ A passing Happo check means that the test detected no visual changes.
 New components that are not yet covered by Happo screenshots should include visual regression test coverage. In order to write a new test, add a file to your component directory titled `[my-component]-happo.ts`, substituting [my-component] with the name of your component.
 
 A simple Happo test just requires the component to be appended to the body of the DOM and exported as a function that returns `componentOnReady()` on the component element:
-```
+```js
 export const skeleton = () => {
   const details = document.createElement('manifold-resource-details-view');
 
@@ -60,7 +60,7 @@ export const skeleton = () => {
 ```
 
 Components that use data can add mocked data to the element objects with the `fromJSON` util:
-```
+```js
 import resource from '../../spec/mock/cms-stage/resource.json';
 import fromJSON from '../../spec/mock/fromJSON';
 
@@ -75,7 +75,7 @@ export const available = () => {
 ```
 
 Components that have slots can have the slot elements appended to them using `appendChild` before the element is returned:
-```
+```js
 import fromJSON from '../../spec/mock/fromJSON';
 
 export const jawsDB = () => {
@@ -95,7 +95,7 @@ export const jawsDB = () => {
 ```
 
 If necessary, tests can be written with HTML as a string and interpolated using the `toHTML` util:
-```
+```jsx
 import { lock } from '@manifoldco/icons';
 import toHTML from '../../../test-utils/to-html';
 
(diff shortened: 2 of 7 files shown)


src/readme.md
```diff
@@ -32,6 +32,95 @@ Storybook. Testing every version of a component is recommended.
...
+### Writing Happo Tests
+
+New components that are not yet covered by Happo screenshots should include visual regression test coverage. In order to write a new test, add a file to your component directory titled `[my-component]-happo.ts`, substituting [my-component] with the name of your component.
+
+A simple Happo test just requires the component to be appended to the body of the DOM and exported:
+```
+export const skeleton = () => {
+  const details = document.createElement('manifold-resource-details-view');
+
+  document.body.appendChild(details);
```
sslotsky: This example doesn't do `return details.componentOnReady()`. Could you update the actual test code and then change it here as well?

src/readme.md
```diff
@@ -32,6 +32,95 @@ Storybook. Testing every version of a component is recommended.
...
+
+If necessary, tests can be written with HTML as a string and interpolated using the `toHTML` util:
+```
+import toHTML from '../../../test-utils/to-html';
+
+export const primary = () => {
+  const content = `
+    <manifold-tooltip label-text="This is a tooltip">
+      <span class="value" data-value="42" data-locked>
+        <manifold-icon class="icon" icon="${lock}" margin-right></manifold-icon>
```
sslotsky: For completeness, let's include the import statement for the lock icon as well.

src/readme.md
```diff
@@ -32,6 +32,95 @@ Storybook. Testing every version of a component is recommended.
...
+  `;
+
+  const tooltip = toHTML(content) as HTMLManifoldTooltipElement;
+  document.body.appendChild(tooltip);
+
+  return tooltip.componentOnReady();
+};
+```
+
+Returning on `componentOnReady()` helps ensure the screenshot loads all the necessary data before the test is run.
```
sslotsky: What this really does is ensure that the component is hydrated before Happo calculates the bounding rectangle and takes the screenshot. I wonder if we need a little more emphasis around the importance of ending each test this way. In the future I could maybe see a lint rule to enforce this, but for now, ensuring that those who read the docs can easily find this explanation should go a long way.

src/readme.md
```diff
@@ -32,6 +32,95 @@ Storybook. Testing every version of a component is recommended.
 | `npm run test:spec [pattern]`  | (fast) Run all unit tests matching `[pattern]`                |
 | `npm run test:watch [pattern]` | (fast) Run all unit tests while working, matching `[pattern]` |
 
+## 🖼️ Visual Regression Testing
+
+We use [Happo](https://happo.io/) integrated into our CI pipeline for visual regression testing.
+
+After a PR is made, Happo tests a screenshot of the new PR against a screenshot of the previous version of the same component. If a visual change is detected, the Happo check in the CI pipeline will fail, and it will require human eyes to approve the check.
+
+You can do this by clicking `Details` on a failed Happo check and using the `Review ˅` drop down to Accept or Reject the visual differences you see in the screenshot tests. If the changes are intended PR changes, you can Accept, and if the changes are unexpected you can Reject and the author will be signaled to review their changes.
```
sslotsky: As you mentioned to me in Slack, a note about possible diffs due to animations might be helpful. Henric did suggest a way to prevent the animations but I'm not sure how realistic it will be for us to implement it. I plan to look into it more, though.

src/readme.md
batgirl: Should I check all the tests and make sure they return on componentOnReady?

sslotsky: That would be awesome.

src/components/manifold-select/manifold-select-happo.ts
```diff
@@ -23,4 +23,6 @@ export const primary = () => {
   select.name = name;
 
   document.body.appendChild(select);
+
+  return select.componentOnReady();
```
dangodev: 💯

src/readme.md
dangodev: Minor detail: tag this with ```js for syntax highlighting

src/readme.md
```diff
@@ -32,6 +32,100 @@ Storybook. Testing every version of a component is recommended.
...
+
+  productPage.appendChild(button);
+  document.body.appendChild(productPage);
+
+  return productPage.componentOnReady();
+};
+```
+
+If necessary, tests can be written with HTML as a string and interpolated using the `toHTML` util:
+```
```
dangodev: Tag this with ```jsx

PR #179, opened 2019-06-06
fix resource container

davidleger95:
## Reason for change

Resource container was watching for the wrong field and not fetching new resource data on change.


src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -13,7 +13,7 @@ export class ManifoldResourceContainer {
   @Prop() resourceLabel: string;
   @State() resource?: Gateway.Resource;
   @State() loading: boolean = false;
-  @Watch('resourceName') resourceChange(newName: string) {
+  @Watch('resourceLabel') resourceChange(newName: string) {
```
davidleger95: Do we want `resourceLabel` or `resourceName`?

davidleger95: @sslotsky I think you're the only one around with context for this. I kept the same to not break the current API, but personally I like `resourceName`.

sslotsky: Ah! I switched it to label because that's the name of the field in the back end. I can go either way, I just figured it would be more clear to developers which piece of data controls this.

davidleger95: I'll take a look tomorrow at what some of the other components are doing and update it if it's going against naming convention (I think we use `name` for a lot of things)

dangodev: FWIW the two will be synonymous going forward (I think). We may just make all resource names be slugified.

davidleger95: updated to use `resourceName`

sslotsky: Are there things like storybook stories or test cases that are assigning `resource-label` to any of these? Can you make sure to get those?

davidleger95: ahh, right. I searched for `resourceLabel` but not `resource-label`. I'll check

davidleger95: updated

PR #189, opened 2019-06-10
Update to Stencil v1

dangodev:
## Reason for change
Updates Stencil to v1. Easier to do all the dependency updates in one go

## Testing
Make sure preview links work on Heroku & Netlify


dangodev:
> It probably wasn't introduced with this change, but on the docs site, the Resource Credentials component is broken. I might have forgotten to wrap it in a resource-container?

Hm. I’m not sure about that, but I did notice the `onClickEvent` was misspelled. Just fixed that.

> For the Netlify preview of the Storybook app, I think auth is broken. Any ideas for a long term solution on this one?

Hm. Hopefully not _broken_, just missing? I’m not sure either. One of our goals with web components is to encapsulate API calls & logic, but we need them to talk to real endpoints. Only thing I can think of is to have some `demo` environment that talks to a dummy endpoint that returns fake product data. Would something like that work?

PR #190, opened 2019-06-11
Buttons

dangodev:
## Reason for change
This adds colors for `<manifold-button>` for the new Sphericon designs

Screen Shot 2019-06-10 at 16 37 37
Screen Shot 2019-06-10 at 16 37 40
Screen Shot 2019-06-10 at 16 37 43
Screen Shot 2019-06-10 at 16 37 47
Screen Shot 2019-06-10 at 16 37 50

## Testing

Happo tests were added 🦛


.storybook/config.js
```diff
@@ -11,31 +11,9 @@ addDecorator(addReadme);
...
-  'manifold-data-product-logo',
-  'manifold-data-resource-list',
-  'manifold-data-provision-button',
-  'manifold-data-manage-button',
-];
-
-function lookup(filename) {
-  const index = storiesOrder.findIndex(name => filename.indexOf(name) !== -1);
-  return index !== -1 ? index : Number.MAX_SAFE_INTEGER;
-}
```
dangodev: Maintaining manual order was just confusing for Storybook. Alphabetical is fine 🤷‍♂

docs/docs/components/manifold-button.md
```diff
@@ -0,0 +1,43 @@
...
+<manifold-button size="small">Small size</manifold-button>
+```
+
+## Events
+
+This element will emit a `manifold-button-click` event when clicked.
+
+```js
+document.querySelector('manifold-button').addEventListener('click', () => /* do something */);
+```
```
dangodev: These new MD files are internal and won’t show up on docs, but we can still reference them from Storybook

dangodev:
Yeah @markjamieson this is not a huge PR, but this will help us figure out how best to consume web components within React.

To be clear: we’re still using Möbius for new web components, but as we already had most of the button code already in UI (just was missing some colors), I figured having 1 button component is better than duplicating between UI vs Möbius. If everyone else is on board I think a good approach should be:

1. If it doesn’t exist, try creating it in manifoldco/mobius
2. If this just plain sucks as a web component talking to React, move it to React (Sphericon)
3. If it exists in UI already, then just modify/add an existing component (but flag Drew or Dave on review just to make sure it doesn’t disrupt DO)

PR #191, opened 2019-06-12
fix: local connection is http not https

seriousben:
## Reason for change

When using UI locally, we do not have TLS enabled.

## Testing

Using your local Manifold server, try to use the local connection.

### Before
image

### After

image


sslotsky:
Since testing this is kind of involved, can you show some before & after screenshots?

seriousben:
Changed the description with some screenshots.

sslotsky:
Closing in favor of #192

PR #193, opened 2019-06-12
Fix Types

dangodev:
## Reason for change
Consuming web components in Sphericon yields some type errors (some of which are new since v1) like these:

68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3561623366336664346235383036626332626363303032322f36613337366533612d333933662d343364612d613738372d613730666339616231633530

This PR fixes those to make it usable in a TS + React context. It shouldn’t have any visible change to the components.


package.json
```diff
@@ -52,14 +51,15 @@
     "test:debug": "node --inspect-brk node_modules/.bin/stencil test --spec --e2e --runInBand",
     "test:watch": "stencil test --spec --e2e --watchAll"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@stencil/state-tunnel": "^1.0.1"
+  },
```
dangodev: @stencil/core needs to remain a devDep (client doesn’t need it), but @stencil/state-tunnel contains types that TypeScript needs. Moving it to deps means that all our clients install it, but that’s necessary for TypeScript (it won’t actually affect their client bundles).

src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -1,4 +1,6 @@
 import { h, Component, State, Prop, Watch } from '@stencil/core';
+import { Catalog } from '../../types/catalog';
+import { Gateway } from '../../types/gateway';
```
dangodev: Biggest new discovery is that when Stencil ships, it won’t contain any references to these namespaces. So all these have to be relatively imported now.

tsconfig.json
```diff
@@ -16,10 +16,7 @@
     "jsx": "react",
     "jsxFactory": "h",
     "strictNullChecks": true,
-    "resolveJsonModule": true,
-    "paths": {
-      "*": ["../node_modules/*", "*"]
-    }
```
dangodev: This is probably the most important change: **prevent absolute paths.** The way Stencil ships code to npm, it won’t transform these. So any absolute paths will come back to bite us. I don’t know how to enforce this other than make it impossible.

davidleger95:
The failed CI build failed trying to publish to npm. Should it have tried to publish to npm?

dangodev:
> The failed CI build failed trying to publish to npm. Should it have tried to publish to npm?

Oh I deployed locally; it may have tried to deploy the same version and failed. It’s fine; this needs one more fix anyway I think.

PR #206, opened 2019-06-17
Allow preloaded services for SSR apps

sslotsky:
## Reason for change

This allows `preloadedServices` to be attached to `manifold-marketplace`, which will prevent the fetch call. This change is intended to support the ability to SSR our web components.

## Testing

1. Publish an alpha release from this PR
2. Download the alpha to manifold-www
3. Use [this proof of concept](https://github.com/manifoldco/manifold-www/pull/534/files#diff-f091411e6abadaa4e61d67e0a25571f7) as guidance for how to render the marketplace grid.

### DIFF
--- a/src/components/manifold-marketplace/manifold-marketplace.tsx
+++ b/src/components/manifold-marketplace/manifold-marketplace.tsx
@@ -25,7 +25,7 @@ export class ManifoldMarketplace {
   @Prop() products?: string;
   /** Template format structure, with `:product` placeholder */
   @Prop() templateLinkFormat?: string;
-  @Prop() preloadedServices?: Catalog.Product[] = [];
+  @Prop() preloadedServices?: Catalog.Product[];
   @State() parsedExcludes: string[] = [];
   @State() parsedFeatured: string[] = [];
   @State() parsedProducts: string[] = [];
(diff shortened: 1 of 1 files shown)


src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -36,10 +37,12 @@ export class ManifoldMarketplace {
   }
 
   fetchProducts = async () => {
-    const response = await fetch(`${this.connection.catalog}/products`, withAuth());
-    const products: Catalog.ExpandedProduct[] = await response.json();
-    // Alphabetize once, then don’t worry about it
-    this.services = [...products].sort((a, b) => a.body.name.localeCompare(b.body.name));
+    if (!this.preloadedServices) {
```
dangodev: I think `[]` will always be true? Probably need to check array length or something

sslotsky: Oh, yeah. I meant to make the default value null. Thanks.

davidleger95:
should this be merged?

sslotsky:
> should this be merged?

@davidleger95 not until we've gone through the testing. We'll have to publish an alpha release and then have @TerrySmithDC try to use it in www.

PR #222, opened 2019-06-25
Moved tag styles to manifold-badge. Added free style support.

TerrySmithDC:
## Reason for change
Plan selector didn't support free badge styles.

## Testing
Check the plan selector and the service card.

### DIFF
--- a/src/components/manifold-badge/mf-badge.css
+++ b/src/components/manifold-badge/mf-badge.css
@@ -11,9 +11,6 @@
 
   display: inline-flex;
   align-items: center;
-  margin-top: 0.125rem;
-  margin-bottom: 0.125rem; /* top & bottom help when tags stack */
-  margin-left: 0.5em;
   padding: var(--padding);
   color: var(--text-color);
   font-weight: var(--font-weight);
--- a/src/components/manifold-badge/mf-badge.css
+++ b/src/components/manifold-badge/mf-badge.css
@@ -3,13 +3,17 @@
   --text-color: var(--manifold-tag-text-color, var(--manifold-text-color-accent));
   --font-family: var(--manifold-tag-font-family, var(--manifold-font-family));
   --font-weight: var(--manifold-tag-font-weight, 700);
-  --font-size: var(--manifold-tag-font-size, var(--manifold-font-d2));
+  --font-size: var(--manifold-tag-font-size, var(--manifold-font-d3));
   --text-transform: var(--manifold-tag-text-transform, initial);
-  --background: var(--manifold-color-primary);
+  --background: var(--manifold-tag-free-background, var(--manifold-color-primary));
+  --free-background: var(--manifold-tag-free-background, var(--manifold-color-primary));
   --radius: var(--manifold-tag-radius, 3em);
 
   display: inline-flex;
   align-items: center;
+  margin-top: 0.125rem;
+  margin-bottom: 0.125rem; /* top & bottom help when tags stack */
+  margin-left: 0.5em;
   padding: var(--padding);
   color: var(--text-color);
   font-weight: var(--font-weight);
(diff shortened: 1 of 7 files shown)


dangodev:
Screen Shot 2019-06-26 at 10 01 28

I’m not seeing the “free” tag anymore on the marketplace element on initial load

davidleger95:
Screen Shot 2019-06-26 at 1 24 17 PM

Not sure what you're seeing. this is what i get

TerrySmithDC:
I get the free tag too but am having issues getting the free style to take effect.

src/components/manifold-badge/mf-badge.css
```diff
@@ -19,5 +23,7 @@
   background: var(--background);
   border-radius: var(--radius);
 
-  /* TODO: move .tag styles from service-card to here and unify naming */
+  &[data-tag='free'] {
+    background: var(--free-background);
```
davidleger95: > having issues getting the free style to take effect.

maybe attr selectors don't work on the `:host`?

davidleger95: hmm, try creating a new selector: `:host[data-tag="free"]`. I'm wondering if the nesting is getting messed up.

davidleger95: I just tried this and it works:

Screen Shot 2019-06-26 at 1 46 08 PM

src/components/manifold-badge/mf-badge.css
```diff
@@ -3,13 +3,17 @@
   --text-color: var(--manifold-tag-text-color, var(--manifold-text-color-accent));
   --font-family: var(--manifold-tag-font-family, var(--manifold-font-family));
   --font-weight: var(--manifold-tag-font-weight, 700);
-  --font-size: var(--manifold-tag-font-size, var(--manifold-font-d2));
+  --font-size: var(--manifold-tag-font-size, var(--manifold-font-d3));
   --text-transform: var(--manifold-tag-text-transform, initial);
   --background: var(--manifold-color-primary);
```
davidleger95: Just realized this should be `--background: var(--manifold-tag-background, var(--manifold-color-primary));` so that we can specify an override that's different from the primary color.

davidleger95: That would also mean the free one should change to `--free-background: var(--manifold-tag-free-background, var(--background));` so the background will be applied to free tags if there's no override specified for that one

src/components/manifold-badge/mf-badge.css
dangodev: Could be. Maybe `:host[data-tag="free"]` isn’t a valid CSS selector.

TerrySmithDC: I'm still working on this, I also tried making this a proper web component prop and tried to style that and it's still not taking effect. Is there a component you style based on state I can look at?

davidleger95: Hack: just put `[data-tag="free"] { ... }` in the global styles 😛

TerrySmithDC: I think I figured out a "right" way 👀

TerrySmithDC:
Fixed one thing and I noticed more things. Still working on this. Plan Selector everything says free now? Also I think I need to move the margins over as it doesn't work for plan selector.

dangodev:
Nice! Yeah the free tag is working & styled correctly for me on Marketplace. Also yeah it’s showing for all plans on planSelector.

dangodev:
Hm. The “free” showing may not be you. I think it’s because the cost API is returning `400`. That can be addressed in another PR, or you can just show `plan.body.cost` as a fallback if the call fails in `<manifold-cost-display>`.

PR #231, opened 2019-07-02
Add a plan display component that can be used independently

Minivera:
This component can be used by platforms to display a plan's information if require. This can power resource details page or custom product display pages.

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -230,6 +230,20 @@ export namespace Components {
     'suffix': string;
     'value': number;
   }
+  interface ManifoldPlan {
+    /**
+    * _(hidden)_ Passed by `<manifold-connection>`
+    */
+    'connection': Connection;
+    /**
+    * URL-friendly slug (e.g. `"kitefin"`)
+    */
+    'planLabel'?: string;
+    /**
+    * URL-friendly slug (e.g. `"jawsdb-mysql"`)
+    */
+    'productLabel'?: string;
+  }
   interface ManifoldPlanCost {
     'allFeatures': Catalog.ExpandedFeature[];
     'compact'?: boolean;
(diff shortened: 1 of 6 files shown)


src/components.d.ts
```diff
@@ -230,6 +230,20 @@ export namespace Components {
     'suffix': string;
     'value': number;
   }
+  interface ManifoldPlan {
+    /**
+    * _(hidden)_ Passed by `<manifold-connection>`
+    */
+    'connection': Connection;
```
dangodev: I think we’re inconsistent, but this should be marked as optional. Reason is that within TypeScript projects, it’ll throw an error otherwise (connection is internal, so users shouldn’t have to specify that attribute).

Like I said there may still be some spots where that isn’t the case, but just something to note.

dangodev: Er—whoops! I meant in `manifold-plan.tsx`. Not here; this is the auto-generated file.

src/components/manifold-plan/manifold-plan.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    const newPlan = 'new-plan';
+
+    const planSelector = new ManifoldPlan();
+    planSelector.fetchProductAndPlan = jest.fn();
+    planSelector.productLabel = productLabel;
+    planSelector.planLabel = 'old-plan';
+    planSelector.planChange(newPlan);
+    expect(planSelector.fetchProductAndPlan).toHaveBeenCalledWith(productLabel, newPlan);
+  });
+});
```
dangodev: Nice! In general, we prefer E2E tests—and tests that don’t raise coverage—over spec. Spec is great for certain things, but in general web components being largely HTML are very hard to test & raise coverage in the same way you can with React (there’s no importing; it’s all just HTML).

So all that said, coverage isn’t as big a goal with UI as E2E tests w/ Puppeteer

PR #234, opened 2019-07-03
Adds the auth token provider to the UI library with docs

Minivera:
Work is related to manifoldco/engineering#8651

## Reason for change
This new component allows you to provider the manifold auth token for authentication of the user anywhere in your dom tree. If provided a valid token, all calls to the database will now be done using the token.

## Testing
Test with storybook.


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -8,9 +8,19 @@ export class ManiTunnel {
   /** _(optional)_ Specify `env="stage"` for staging */
   @Prop() env: 'stage' | 'prod' = 'prod';
 
+  @State() authToken?: string = localStorage.getItem('manifold_api_token') || undefined;
+
+  setAuthToken = (token: string) => {
```
sslotsky: Ahh, this should work great! Now some descendant of `<manifold-connection>` can do like

```
<manifold-oauth onTokenReceived={this.setAuthToken} />
```

Minivera: Exactly!

Minivera: Or if we want to have the platfrom/provider handle the auth token, that even could be linked to a React event handler that rerenders the `manifold-auth-token` component with the token attribute changed.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -32,18 +34,18 @@ export class ManifoldDataProductLogo {
   fetchProduct = async (productLabel: string) => {
     this.product = undefined;
     const { catalog } = this.connection;
-    const response = await fetch(`${catalog}/products?label=${productLabel}`, withAuth());
+    const response = await fetch(`${catalog}/products?label=${productLabel}`, withAuth(this.authToken));
```
sslotsky: This really makes me wish that we had an `AuthenticatedQuery` component that could yield data via render props. I enjoy Stencil but I've had many occasions to remember that React can be pretty great!

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -8,9 +8,19 @@ export class ManiTunnel {
   /** _(optional)_ Specify `env="stage"` for staging */
   @Prop() env: 'stage' | 'prod' = 'prod';
 
+  @State() authToken?: string = localStorage.getItem('manifold_api_token') || undefined;
```
dangodev: Nice! Good `null` catch

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
dangodev: Agreed. Though I don’t think this is Stencil’s design decision; I think the web component spec itself adopting the ES Class model, rather than a more functional approach, has limitations. I think Stencil did its best to make everything React-like, but at the end of the day web component spec being OO does make this difficult.

i.e. I’m sure there’s a way to do this but it may have tradeoffs that Stencil didn’t want to absorb. But I’m making all this up in my head 🤷‍♂. Again—agree that I like render props. Maybe we could make this into a function component in Stencil or something 🤔

Minivera: I think stencil does support render props though, when you look at statetunnel https://github.com/ionic-team/stencil-state-tunnel/blob/master/packages/state-tunnel/src/components/state-tunnel.tsx#L11

sslotsky: Hmm you're right!

PR #235, opened 2019-07-04
Add a resource card component to UI

Minivera:
Work is related to manifoldco/engineering#8643

## Reason for change
This component will display the details on a resource and can be clicked to access a resource details page.

## Testing
Test with storybook

### DIFF
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view.css
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view.css
@@ -92,8 +92,8 @@
   grid-area: availability;
   padding-top: 0.25rem;
   color: var(--manifold-text-color);
-  font-family: var(--manifold-font-family);
   font-size: var(--manifold-font-d1);
+  font-family: var(--manifold-font-family);
 
   @media (--viewport-m) {
     padding-top: 0;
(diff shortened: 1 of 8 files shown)


markjamieson:
Screen Shot 2019-07-04 at 2 39 56 PM
Should the font here be the Manifold font?

Minivera:
The font family is inherited, I might not have set it in the wrapper

Minivera:
It's fixed. The issue wasn't happening in the docs and I'm not sure if they would happen in the integrations themselves, but better safe than sorry.

src/components/manifold-resource-card-view/manifold-resource-card-view.css
```diff
@@ -0,0 +1,201 @@
...
+.name {
+  grid-area: name;
+  align-self: start;
+  margin-top: 0;
+  margin-bottom: 0;
+  padding-right: 1em;
+  color: var(--card-heading-text-color);
+  font-weight: var(--card-heading-font-weight);
+  font-size: 1em;
+  font-family: var(--card-heading-font-family);
```
markjamieson: I'm still seeing the same missing font on Storybook after setting it in the wrapper. Like Storybook can't parse this variable. Maybe it can't parse variables in the :host pseudo-class? I'm not that familiar with the concept.

Minivera: I added it here: https://github.com/manifoldco/ui/pull/235/files#diff-a4f7eddfa20039dd50187bdcf2bfa774R44

I've tested in render and it looks fine, I think this is only a storybook issue.

markjamieson: Ok, I won't block on Storybook weirdness.

PR #240, opened 2019-07-08
Remove the input from `manifold-data-provision-button`

Minivera:
Work is related to manifoldco/engineering#8687

## Reason for change
That input can now be added outside the component itself for an easier control of it's position, associated label and stylings. Instead, a `resource-name` attribute must be provided to the provision button for resources to be properly provisioned.

Also added a lot of spec tests to the provision button.

## Testing
Test with storybook


Minivera:
I can confirm in travis that the job did pass, since travis is having issues, I'll merge
https://travis-ci.com/manifoldco/ui

PR #241, opened 2019-07-09
We should not require more than the LTS as minimum version

jeffandersen:
## Reason for change

- We broke Render's production builds because we required a version greater than the LTS

## Testing

- Projects using Node 10.16 should build successfully
- Web components should operate nominally


jeffandersen:
The build failures are beyond my ability, but this suggests there are some minor changes in compatibility between 10.16 and 12.4

PR #244, opened 2019-07-09
Add two new components for displaying a resource's details

Minivera:
Work is related to manifoldco/engineering#8689

## Reason for change
The product and the plan component will display the resource's product and plan respectively thanks to the `resource-container` component. These can be used by platforms to build a resource details page with their own styling thanks to the state tunnel in `resource-container`.

For testing purposes, I also added a `mock-resource` component that allows us or platforms to fake a `resource-container` for testing purposes.

## Testing
Test with storybook


src/utils/gatewayToCatalog.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { convertPlan, convertProduct } from './gatewayToCatalog';
+import { ResolvedPlan, ResolvedProduct } from '../spec/mock/gateway';
+
+describe('The convertPlan method', () => {
+  it('will transform a gateway plan into a valid catalog plan', () => {
+    const productId = ResolvedProduct.id || '';
+    const providerId = ResolvedProduct.provider ? ResolvedProduct.provider.id : '';
+    expect(convertPlan(ResolvedPlan, productId, providerId || '')).toMatchSnapshot();
```
Minivera: This is one rare case where I think using snapshots make sense

Minivera:
> I tested in Storybook and it looks great - the Resource Product Details component is unstyled but I assume that's intentional. And does the as-card="" attribute require the empty string?

For the styling, it's not intended, that's a storybook issue. If you look at the docs deploy, it should look right. For the `as-card=""`, it's only required in pure HTML in storybook and in our docs I think, it works fine when used in react.

> And just since it's a lot to read over and I'm trying to understand - the manifold-service-view is something you're abstracting away because it can be used in both manifold-service-card and manifold-resource-product, is that correct?

That's exactly it! I am trying to abstract display components and data components so we can reuse them in various context, in our current case, the resource-container context.

src/components/manifold-service-view/manifold-service-view.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <h3 class="name" itemprop="name">
+          {this.name}
+        </h3>
+        <div class="info">
+          <p class="description" itemprop="description">
+            {this.description}
+          </p>
+        </div>
+        {!this.hideTags && (<div class="tags">
+          {this.isFeatured && <manifold-badge>featured</manifold-badge>}
```
davidleger95: very small request: can this be changed to `Featured`. Just to keep capitalization consistent with the `Free` tag

src/utils/gatewayToCatalog.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Catalog } from '../types/catalog';
+import { globalRegion } from '../data/region';
+
+export const convertPlan = (plan: Gateway.ResolvedPlan, productId: string, providerId: string): Catalog.Plan => ({
+  id: plan.id || '',
+  body: {
+    ...plan,
+    product_id: productId,
+    provider_id: providerId,
+    regions: [globalRegion.id], // TODO: Until we get graphQL, we're going to lise regions
```
davidleger95: `list regions`?

src/components/manifold-service-view/manifold-service-view.tsx
Minivera: For sure!

src/utils/gatewayToCatalog.ts
Minivera: `lose`, my bad. Essentially, regions are not given from gateway when fetching resources, so unless we do another fetch call, we lose the regions. That's not a problem since this component only displays region, but GraphQL will fix both of these issues.

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -70,45 +73,31 @@ export class ManifoldServiceCard {
...
     return !this.skeleton ? (
       <a
-        class="wrapper"
         role="button"
         itemscope
         itemtype="https://schema.org/Product"
         itemprop="url"
         href={this.href}
         onClick={this.onClick}
+        style={{ textDecoration: 'none' }}
```
davidleger95: should this be here or moved to the stylesheet?

Minivera: I fetch a bit conflicted about having this in a stylesheet since that would mean having a stylesheet and the shadowdom only for that small property :/

src/utils/gatewayToCatalog.spec.ts
dangodev: I agree with this.

Though my personal preference is comparing generated files to an output is clearer than trying to read Jest’s format (also it’s harder to blow away the snapshot with a flag), but that’s an opinion. I think overall this is great!

stencil.config.ts
```diff
@@ -9,7 +9,7 @@ export const config: Config = {
   globalStyle: 'src/global/theme.css',
   globalScript: 'src/global/app.ts',
   outputTargets: [{ type: 'dist' }],
-  excludeSrc: ['**/*-happo.*', '**/spec/mock/*'],
```
dangodev: Make sure this isn’t bundled in the npm output! This was done to exclude that from exporting.

src/components/manifold-service-card/manifold-service-card.tsx
dangodev: Stylesheet please! Cascade is very different for inline styling.

stencil.config.ts
Minivera: Is there any way to have the `manifold-mock-resource` component work in storybook and the docs without this being bundled in NPM? That component uses some mocks.

src/components/manifold-service-card/manifold-service-card.tsx
dangodev: OK I see what the issue is now. … I’ll allow it

stencil.config.ts
dangodev: Hm not sure. If this is causing problems then you can remove it like you did. This won’t affect the actual client-side code, but for changes like this just inspect the `dist/` folder to make sure we’re not shipping super-large files that have to be re-downloaded by build tools every single time (deploys, Docker, Travis, local dev, etc.)

PR #248, opened 2019-07-10
Add a new rename button for resources

Minivera:
Work is related to manifoldco/engineering#8690

## Reason for change
This button can be used as a data component standalone to rename a resource. It will require a resource name and can fetch the resource id if not provided. The button expects the new name for the resource for renaming.

It can also be used as part of the `resource-container` component by using its `manifold-resource-rename` version. Both version are stateless and will need a corresponding input to work.

## Testing
Test with storybook


src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -0,0 +1,177 @@
...
+      resourceId: this.resourceId,
+    };
+    this.clickEvent.emit(clickMessage);
+
+    const body: Marketplace.PublicUpdateResource = {
+      body: {
+        name: this.newName,
+        label: this.newName,
+      },
+    };
```
dangodev: I’m not sure if you like TS yet, but being able to specify `Marketplace.PublicUpdateResource` as the body type IMO is much better than Mayflower 😄. This is pretty clean/safe to me (not to mention 0 KB on the client, unlike Mayflower)!

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -0,0 +1,177 @@
...
+        message: `${this.resourceName} successfully renamed`,
+        resourceName: this.resourceName || '',
+        newName: this.newName,
+        resourceId: this.resourceId,
+      };
+      this.successEvent.emit(success);
+    } else {
+      const result = await response.json();
+
+      // Sometimes messages are an array, sometimes they aren’t. Different strokes!
```
dangodev: Ugh. I’ve dealt with this a lot 😣

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  async fetchResourceId(resourceName: string) {
+    const resourceResp = await fetch(
+      `${this.connection.marketplace}/resources/?me&label=${resourceName}`,
+      withAuth(this.authToken)
+    );
+    const resources: Marketplace.Resource[] = await resourceResp.json();
+
+    if (!Array.isArray(resources) || !resources.length) {
+      console.error(`${resourceName} product not found`);
```
dangodev: Should this dispatch a `manifold-resourceButton-error` event?

src/components/manifold-input/manifold-input.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+@Component({
+  tag: 'manifold-input',
+  styleUrl: 'manifold-input.css',
+  scoped: true,
+})
+export class ManifoldSelect {
+  @Prop() inputId?: string;
+  @Prop() name?: string;
+  @Prop() type?: string = 'text';
```
dangodev: I thought I remembered some weirdness about `type=""` in a TypeScript React app, but now I don’t remember. Works for now!

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
Minivera: Yeah, pretty much this is your code I copied over and over again 😅

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
Minivera: Not sure, all our other buttons like this don't. I think this would be more of a case for the error boundary we want to build.

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
Minivera: My issue with TypeScript has always been that it's still JavaScript under the hood and its compiled nature means they can't implement things like `variable is Type` and that makes me so annoyed sometimes.

PR #249, opened 2019-07-10
Marketplace Header bugfix

TerrySmithDC:
## Reason for change
Marketplace mobile categories weren't being styled correctly.

## Testing
Check `<marketplace>` on mobile scrolling works and is usable.

### DIFF
--- a/docs/docs/theming.md
+++ b/docs/docs/theming.md
@@ -291,6 +291,7 @@ Cards are used in Marketplace (service cards) and Plan Selector (plan buttons, p
 | <span class="tag not-implemented" /> `--heading-color`                 | `var(--manifold-text-color-secondary)` | Category heading text color             |
 | <span class="tag not-implemented" /> `--heading-font-size`             | `--`                                   | Category heading font size              |
 | <span class="tag not-implemented" /> `--heading-font-weight`           | `inherit`                              | Category heading font weight            |
+| `--heading-background`                                                 | `rgba(255,255,255,1)`                  | Category heading background style       |
 |                                                                        |                                        |                                         |
 | <span class="tag not-implemented" /> `--grid-gap`                      | `1rem`                                 | Gap between product/service cards       |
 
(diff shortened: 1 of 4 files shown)


docs/docs/theming.md
```diff
@@ -291,6 +291,7 @@ Cards are used in Marketplace (service cards) and Plan Selector (plan buttons, p
 | <span class="tag not-implemented" /> `--heading-color`                 | `var(--manifold-text-color-secondary)` | Category heading text color             |
 | <span class="tag not-implemented" /> `--heading-font-size`             | `--`                                   | Category heading font size              |
 | <span class="tag not-implemented" /> `--heading-font-weight`           | `inherit`                              | Category heading font weight            |
+| `--heading-background`                                                 | `rgba(255,255,255,1)`                  | Category heading background style       |
```
davidleger95: If this is going to be a global variable, it should be namespaced with `--manifold-*`, and documented . The reason why these ones aren't namespaced that way is because they they're internal and are only defined within the component, so they only are visible to descendants.

Going back on what I said earlier, I don't think this needs to be exposed at the moment. It's hard to add just one of these without adding all of them. The plan for theming at this low level is to have element selectors that override the defaults like this:

```css
manifold-marketplace {
  --heading-background: white;
}
```

I'm not sure if that makes sense. I can elaborate more on this tomorrow if not. This is hard to explain :P

TerrySmithDC: It does, I had it in my head we would want that to be customizable. But I'm totally fine not exposing it.

davidleger95: Yeah, it'll be customizable eventually. Just don't want to introduce a pattern into the theming API that is likely to change if we don't have to, and for our current use cases we don't need to customize it

dangodev: Well, on second-thought—does this need to be customizable? I’d be OK with not exposing it, and enforcing it to be the background color. It really won’t look good being any other color.

davidleger95: I think it should match the background color of whatever page it's on so i blends in. For example, if we were using headings on Render, we would want these to be offwhite to match their background

PR #251, opened 2019-07-10
Make the resource credentials component more standalone

Minivera:
Work is related to manifoldco/engineering#8702

## Reason for change
The `manifold-resource-credentials` component is now only an helper component that wraps the real `manifold-credentials` component. This component can be given its own props that will allow it to find the credentials and fetch its own resource id.

## Testing
Test with storybook


dangodev:
This fixes the wonky spinner in `manifold-credentials.css`:

```diff
.spin {
+  align-items: center;
+  display: flex;
  animation-name: spin;
  animation-duration: 1s;
  animation-timing-function: linear;
  animation-iteration-count: infinite;
}
```

PR #256, opened 2019-07-11
POC for shadowcat integration. (this is just an example with docs, not looking to merge this)

davidleger95:
## Reason for change

- showing how to use shadowcat


docs/src/pages/index.tsx
```diff
@@ -2,11 +2,13 @@ import React from 'react';
 import { ThemeProvider } from 'styled-components';
 import { graphql } from 'gatsby';
 import { defineCustomElements } from '@manifoldco/ui/dist/loader';
+import { defineCustomElements as defineShadowcatCustomElement } from '@manifoldco/shadowcat/dist/loader';
 import Page from '../components/Page';
 import theme from '../lib/theme';
 
 if (typeof window !== 'undefined') {
   defineCustomElements(window);
+  defineShadowcatCustomElement(window);
```
davidleger95: @batgirl just thought I'd push this up to show how I'm using shadowcat.

docs/src/types/custom.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace JSX {
+  export interface IntrinsicElements {
+    'manifold-oauth': any;
+  }
```
davidleger95: For prod we will want to actually use the correct types here. `types/custom-elements.d.ts` in TacoCloud (https://github.com/manifoldco/manifold-ui-react-demo) have examples of this.

docs/src/pages/index.tsx
```diff
@@ -32,7 +34,10 @@ function HomePage({ data }: HomePageProps) {
 
   return (
     <ThemeProvider theme={theme}>
-      <Page links={links} currentPage={currentPage} />
+      <>
+        <manifold-oauth oauth-url="https://manifold-shadowcat-test-server.herokuapp.com/signin/oauth/web" />
```
davidleger95: So I think we will want to do something similar to this in TacoCloud, and put `<manifold-auth-token />` next to it as well. then we can listen for `recieveManifoldToken` which will then pass the token into  `<manifold-auth-token />` (I think that's how it works).

Alternatively, we could just listen for the event inside the connection, but idk if that reduces the flexibility that @Minivera wants to keep. I can think of ways that it's still just as flexible, but maybe I'm not considering all the use cases.

batgirl: I had been thinking of putting `<manifold-oauth>` inside of the `<manifold-auth-token>` component and listening for `receiveManifoldToken` there, so that platforms only have to place one auth-related component and don't get confused about which is which. Then `<manifold-auth-token>` can pass it to `<manifold-connection>` and that can do whatever it wants with it for loading state, etc.

Then `<manifold-auth-token>` can do what @Minivera said:
```
- On load, if it has a token, give that to the manifold-connection
- Start fetching a newer auth token
- When it receives a auth token, update connection and trigger the dom event for the platform to get the newer token
```

davidleger95: Hmm, is that what Sam was saying the plan last week or was that just your assumption? I don't remember exactly what he said, but I'm wondering if that defeats the purpose of having them as separate packages? maybe it doesn't? I'd say go for it, try it out today and we can ask Sam when he's back if that was his plan.

Maybe we can try a couple approaches and see what works best?

davidleger95:
don't think we need this anymore

PR #257, opened 2019-07-11
Refactor resource name to resource label for clarity

Minivera:
Work is related to manifoldco/engineering#8729

## Reason for change
Everywhere in our docs we are using resource label for the slug we can use to fetch and resource name for the display name. In UI, resource name was the slug and the display name. That distinction has now been fixed and we should have a more clear standard when it comes to names and labels.

This PR also adds a new documentation page for how to create a fully connected resource details view with the components I have been creating.

## Testing
Test with storybook and the docs

### DIFF
--- a/docs/docs/components/manifold-credentials.md
+++ b/docs/docs/components/manifold-credentials.md
@@ -1,0 +1,16 @@
+---
+title: '🔒 Resource Credentials'
+path: '/components/credentials'
+example: '<manifold-credentials resource-name="cms-stage"></manifold-credentials>'
+---
+
+# 🔒 Credentials
+
+Display credentials for a resource. 🔒 Requires authentication.
+
+The resource label needs to be provided for the component to be able to fetch the resource's credentials on demand.
+
+```html
+<manifold-credentials resource-name="my-resource"></manifold-credentials>
+```
+
--- a/docs/docs/components/manifold-credentials.md
+++ b/docs/docs/components/manifold-credentials.md
@@ -11,6 +11,6 @@ Display credentials for a resource. 🔒 Requires authentication.
 The resource label needs to be provided for the component to be able to fetch the resource's credentials on demand.
 
 ```html
-<manifold-credentials resource-name="my-resource"></manifold-credentials>
+<manifold-credentials resource-label="my-resource"></manifold-credentials>
 ```
 
--- a/docs/docs/data/manifold-data-deprovision-button.md
+++ b/docs/docs/data/manifold-data-deprovision-button.md
@@ -16,7 +16,7 @@ An unstyled button for deprovisioning resources. 🔒 Requires authentication.
 Set the CTA text by adding anything between the opening and closing tags:
 
 ```html
-<manifold-data-deprovision-button>
+<manifold-data-deprovision-button resource-label="my-resource">
   Deprovision My Resource
 </manifold-data-provision-button>
 ```
@@ -28,22 +28,22 @@ Set the CTA text by adding anything between the opening and closing tags:
 For validation, error, and success messages, it will emit custom events.
 
 ```js
-document.addEventListener('manifold-deprovisionButton-click', ({ detail: { resourceName } }) =>
-  console.info(`⌛ Derovisioning ${resourceName} …`)
+document.addEventListener('manifold-deprovisionButton-click', ({ detail: { resourceLabel } }) =>
+  console.info(`⌛ Derovisioning ${resourceLabel} …`)
 );
 document.addEventListener(
   'manifold-deprovisionButton-success',
-  ({ details: { resourceName } }) =>alert(`${resourceName} deprovisioned successfully!`)
+  ({ details: { resourceLabel } }) =>alert(`${resourceLabel} deprovisioned successfully!`)
 );
 document.addEventListener('manifold-deprovisionButton-error', ({ detail }) => console.log(detail));
-// {message: "bad_request: bad_request: No plan_id provided", resourceid: "1234", resourceName: "my-resource"}
+// {message: "bad_request: bad_request: No plan_id provided", resourceid: "1234", resourceLabel: "my-resource"}
 ```
 
 | Name                               |                       Returns                        | Description                                                                                                                 |
 | :--------------------------------- | :--------------------------------------------------: | :-------------------------------------------------------------------------------------------------------------------------- |
-| `manifold-deprovisionButton-click`   |                    `resourceName`                   | Fires immediately when button is clicked. May be used to trigger a loading state, until `-success` or `-error` is received. |
-| `manifold-deprovisionButton-success` |     `message`, `resourceId`, `resourceName`         | Successful deprovision. Returns a resource ID and resource Label.                                                                              |
-| `manifold-deprovisionButton-error`   |     `message`, `resourceId`, `resourceName`         | Erred provision, along with information on what went wrong.                                                                 |
+| `manifold-deprovisionButton-click`   |                    `resourceLabel`                 | Fires immediately when button is clicked. May be used to trigger a loading state, until `-success` or `-error` is received. |
+| `manifold-deprovisionButton-success` |     `message`, `resourceId`, `resourceLabel`       | Successful deprovision. Returns a resource ID and resource Label.                                                                              |
+| `manifold-deprovisionButton-error`   |     `message`, `resourceId`, `resourceLabel`       | Erred provision, along with information on what went wrong.                                                                 |
 
 ## Styling
 
--- a/docs/docs/data/manifold-data-deprovision-button.md
+++ b/docs/docs/data/manifold-data-deprovision-button.md
@@ -33,7 +33,7 @@ document.addEventListener('manifold-deprovisionButton-click', ({ detail: { resou
 );
 document.addEventListener(
   'manifold-deprovisionButton-success',
-  ({ details: { resourceLabel } }) =>alert(`${resourceLabel} deprovisioned successfully!`)
+  ({ detail: { resourceLabel } }) =>alert(`${resourceLabel} deprovisioned successfully!`)
 );
 document.addEventListener('manifold-deprovisionButton-error', ({ detail }) => console.log(detail));
 // {message: "bad_request: bad_request: No plan_id provided", resourceid: "1234", resourceLabel: "my-resource"}
--- a/src/components/manifold-credentials/manifold-credentials.tsx
+++ b/src/components/manifold-credentials/manifold-credentials.tsx
@@ -11,20 +11,20 @@ export class ManifoldCredentials {
   @Prop() connection: Connection = connections.prod;
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() authToken?: string;
-  @Prop() resourceName: string = '';
+  @Prop() resourceLabel: string = '';
   @Prop({ mutable: true }) resourceId?: string = '';
   @State() loading?: boolean = false;
   @State() credentials?: Marketplace.Credential[];
 
-  @Watch('resourceName') nameChange(newName: string) {
+  @Watch('resourceLabel') labelChange(newLabel: string) {
     if (!this.resourceId) {
-      this.fetchResourceId(newName);
+      this.fetchResourceId(newLabel);
     }
   }
 
   componentWillLoad() {
-    if (this.resourceName && !this.resourceId) {
-      this.fetchResourceId(this.resourceName);
+    if (this.resourceLabel && !this.resourceId) {
+      this.fetchResourceId(this.resourceLabel);
     }
   }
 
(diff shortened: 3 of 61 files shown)


src/components/manifold-resource-rename/manifold-resource-rename.tsx
```diff
@@ -1,24 +1,30 @@
...
+  tag: 'manifold-resource-rename',
+  shadow: true,
+})
 export class ManifoldResourceRename {
+  @Prop() data?: Gateway.Resource;
+  @Prop() loading: boolean = true;
+
   render() {
     return (
-      <ResourceTunnel.Consumer>
```
Minivera: Did you know that this consumer pattern prevented slots from working in these components? I did not.

dangodev:
I remember talking about this with design, and treating resource name _as_ the label going forward (i.e. not distinguishing them). I got the same sentiment from (I believe James) as well. Is that no longer the case?

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -60,15 +61,18 @@ export class ManifoldDataResourceList {
       e.preventDefault();
       const detail: EventDetail = {
         resourceId: resource.id,
-        resourceName: resource.body.label,
+        resourceLabel: resource.body.label,
+        resourceName: resource.body.name,
```
davidleger95: What's the resourceName here?

Minivera:
I don't think that's the case at all, the pattern in GraphQL is to have a `label` (slug) and a `displayName` https://github.com/manifoldco/marketplace/blob/master/graphql/schema/product.graphql#L5. Our old pattern in REST was very similar with the label and name being the slug and the non-unique display value respectively.

It makes things very confusing to think of the resource's name as the label since names are now display names and are not unique. This get even more confusing when you think about the fact that fetched resources have both a name and a label https://github.com/manifoldco/ui/pull/257/files#diff-b4a3574eaa0ee363fc1cd18bdf94a008R64.

All our other endpoints, including the future graphQL endpoint, uses the label as the slug for fetching resources and the name as the displayName, not the name as both. This would make UI extremely inconsistent with our docs and other codebases.

Maybe the discussion was led by a bug in the gateway specs that made it ignore the name even though it was in the data returned by the endpoints? I have fixed the spec bug, but I still need to fix the same bug in the create resource endpoint on gateway.

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -225,7 +226,7 @@ export class ManifoldPlanDetails {
       const { expanded_features, name: planName } = this.plan.body;
 
       return (
-        <section class="scroll" itemscope itemtype="https://schema.org/IndividualProduct">
+        <section class={this.scrollLocked ? 'scroll' : ''} itemscope itemtype="https://schema.org/IndividualProduct">
```
davidleger95: A pattern that we've been using for conditional styles is using `data-*` attributes. would we be able to switch this to something like `data-scroll-locked={this.scrollLocked}`?

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
Minivera: The real resource name, it's the non-unique display name resources are given when created. The label is usually auto-generated on dashboard from that name using dashify.

src/components/manifold-resource-rename/manifold-resource-rename.tsx
davidleger95: weird 🤔

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
Minivera: The label should not be used as the display name for resources, that's the name's (or in graphQL, the displayName) job and we didn't give that data to platforms for some reason.

src/components/manifold-plan-details/manifold-plan-details.tsx
Minivera: Sure, that won't be a problem.

dangodev:
> Maybe the discussion was led by a bug in the gateway specs that made it ignore the name even though it was in the data returned by the endpoints?

I’m asking if you can verify with design & product on this. I’m telling you they made the call, and to check with them if they’re overriding that decision. It wasn’t an accident that everything is named `resourceName` consistently, even though yes it doesn’t match with our current API.

docs/docs/components/manifold-credentials.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+# 🔒 Credentials
+
+Display credentials for a resource. 🔒 Requires authentication.
+
+The resource label needs to be provided for the component to be able to fetch the resource's credentials on demand.
+
+```html
+<manifold-credentials resource-name="my-resource"></manifold-credentials>
```
davidleger95: Should this also be label, or is it different?

Minivera: Should be label, probably and oversight.

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -11,20 +11,20 @@ export class ManifoldCredentials {
   @Prop() connection: Connection = connections.prod;
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() authToken?: string;
-  @Prop() resourceName: string = '';
+  @Prop() resourceLabel: string = '';
```
dangodev: This is a breaking change. I’d like to not introduce these if possible—a PR like this is **highly dangerous** at this stage, especially now that 2 platforms are reading these docs.

If you want to rename something, make a new one but keep the existing one. You can change them internally but **an external-facing change like this would break our integrations**.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -9,25 +9,28 @@ import { Marketplace } from '../../types/marketplace';
...
+})
 export class ManifoldDataDeprovisionButton {
   @Element() el: HTMLElement;
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() connection: Connection = connections.prod;
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() authToken?: string;
   /** The label of the resource to deprovision */
-  @Prop() resourceName?: string;
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
```diff
@@ -26,7 +30,7 @@ export class ManifoldDataManageButton {
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() authToken?: string;
   /** Name of resource */
-  @Prop() resourceName?: string;
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -17,18 +17,18 @@ export class ManifoldDataProductLogo {
   /** URL-friendly slug (e.g. `"jawsdb-mysql"`) */
   @Prop() productLabel?: string;
   /** Look up product name from resource */
-  @Prop() resourceName?: string;
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

src/components/manifold-data-product-name/manifold-data-product-name.tsx
```diff
@@ -15,18 +15,18 @@ export class ManifoldDataProductName {
   /** URL-friendly slug (e.g. `"jawsdb-mysql"`) */
   @Prop() productLabel?: string;
   /** Look up product name from resource */
-  @Prop() resourceName?: string;
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -34,8 +34,8 @@ export class ManifoldDataProvisionButton {
   @Prop() planLabel?: string;
   /** Region to provision (complete name), omit for all region */
   @Prop() regionName?: string;
-  /** The name of the resource to provision */
-  @Prop() resourceName?: string;
+  /** The label of the resource to provision */
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
davidleger95: so if label is unique but generated from the name, does it add something to the end for duplicate labels?

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -17,15 +17,15 @@ export class ManifoldPlanSelector {
   /** Specify region order */
   @Prop() regions?: string;
   /** Is this tied to an existing resource? */
-  @Prop() resourceName?: string;
+  @Prop() resourceLabel?: string;
```
dangodev: Breaking change

docs/docs/data/manifold-data-deprovision-button.md
```diff
@@ -28,22 +28,22 @@ Set the CTA text by adding anything between the opening and closing tags:
...
 ```js
-document.addEventListener('manifold-deprovisionButton-click', ({ detail: { resourceName } }) =>
-  console.info(`⌛ Derovisioning ${resourceName} …`)
+document.addEventListener('manifold-deprovisionButton-click', ({ detail: { resourceLabel } }) =>
+  console.info(`⌛ Derovisioning ${resourceLabel} …`)
 );
 document.addEventListener(
   'manifold-deprovisionButton-success',
-  ({ details: { resourceName } }) =>alert(`${resourceName} deprovisioned successfully!`)
+  ({ details: { resourceLabel } }) =>alert(`${resourceLabel} deprovisioned successfully!`)
```
dangodev: Is this `details` or `detail`?

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -57,7 +60,7 @@ export class ManifoldDataDeprovisionButton {
     // We use Gateway b/c it’s much easier to provision w/o generating a base32 ID
     this.clickEvent.emit({
       resourceId: this.resourceId,
-      resourceName: this.resourceName || '',
+      resourceLabel: this.resourceLabel || '',
```
dangodev: Also a breaking change—I’d like to emit the original `resourceName` value as well, in addition to this new one.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -82,22 +85,22 @@ export class ManifoldDataDeprovisionButton {
       const message = Array.isArray(body) ? body[0].message : body.message;
       const error: ErrorMessage = {
         message,
-        resourceName: this.resourceName || '',
+        resourceLabel: this.resourceLabel || '',
```
dangodev: Breaking change here, too

docs/docs/data/manifold-data-deprovision-button.md
Minivera: Looks like it's `detail`, out docs uses `details` in a lot of places

https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

src/components/manifold-resource-rename/manifold-resource-rename.tsx
Minivera: Very, but I think it's because of this line here: https://github.com/ionic-team/stencil-state-tunnel/blob/master/packages/state-tunnel/src/components/state-tunnel.tsx#L5

The state tunnel does not forward slots nor does it implement the shadow dom, so I think that slots are simply lost for components wrapper with the render props pattern. I'll make an issue.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
Minivera: The backend will add a number to the end to prevent duplicated labels yes. Though the backend itself will not generate the label from the name, it expects both. In the case of gateway or dashboard, we generate the label automatically.

src/components/manifold-resource-rename/manifold-resource-rename.tsx
dangodev: Slots are still somewhat mysterious to me, but that makes sense—slots do require shadow DOM to work, so this makes sense.

PR #258, opened 2019-07-12
Use shadowcat for oauth 

batgirl:
## Reason for change
Probably over-engineered this, but learned a ton about TypeScript and Stencil in the process!!! There are some linting errors here, but maybe new eyes on this will help us figure out a reasonable way to get the token from the iframe into the token Prop (and then update it when needed). Any way you know to do it is welcome to me.

These are the necessities of this component per @Minivera:
- On load, if it has a token, give that to the manifold-connection
- Start fetching a newer auth token
- When it receives a auth token, update connection and trigger the dom event for the platform to get the newer token
 
I did successfully add `manifoldco/shadowcat` to the `manifold-auth-token` component however. :)

## Testing
Ran this in Storybook and checked the console for the `this.token` value, which was hard.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@graphql-codegen/typescript-operations": "^1.3.1",
     "@manifoldco/eslint-plugin-stencil": "^0.2.1",
     "@manifoldco/icons": "0.0.3",
+    "@manifoldco/shadowcat": "0.0.2",
     "@reach/observe-rect": "^1.0.3",
     "@stencil/core": "^1.1.7",
     "@stencil/postcss": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@graphql-codegen/typescript-operations": "^1.3.1",
     "@manifoldco/eslint-plugin-stencil": "^0.2.1",
     "@manifoldco/icons": "0.0.3",
-    "@manifoldco/shadowcat": "^0.1.3",
+    "@manifoldco/shadowcat": "^0.1.5",
     "@reach/observe-rect": "^1.0.3",
     "@stencil/core": "^1.1.7",
     "@stencil/postcss": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@graphql-codegen/typescript-operations": "^1.3.1",
     "@manifoldco/eslint-plugin-stencil": "^0.2.1",
     "@manifoldco/icons": "0.0.3",
-    "@manifoldco/shadowcat": "0.0.2",
+    "@manifoldco/shadowcat": "^0.1.3",
     "@reach/observe-rect": "^1.0.3",
     "@stencil/core": "^1.1.7",
     "@stencil/postcss": "^1.0.1",
--- a/src/components/manifold-auth-token/manifold-auth-token.spec.ts
+++ b/src/components/manifold-auth-token/manifold-auth-token.spec.ts
@@ -2,7 +2,10 @@ import { ManifoldAuthToken } from './manifold-auth-token';
 
 describe('<manifold-auth-token>', () => {
   it('calls the set auth token on load', () => {
-    const token = 'test';
+    const expiry = new Date();
+    expiry.setDate(expiry.getDate() + 1);
+    const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
+    const token = `test.${encodedExpiry}`;
 
     const provisionButton = new ManifoldAuthToken();
     provisionButton.setAuthToken = jest.fn();
--- a/src/components/manifold-auth-token/manifold-auth-token.spec.ts
+++ b/src/components/manifold-auth-token/manifold-auth-token.spec.ts
@@ -1,19 +1,40 @@
 import { ManifoldAuthToken } from './manifold-auth-token';
 
 describe('<manifold-auth-token>', () => {
-  it('calls the set auth token on load', () => {
+  describe('when the token is not expired', () => {
     const expiry = new Date();
     expiry.setDate(expiry.getDate() + 1);
     const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
-    const token = `test.${encodedExpiry}`;
 
-    const provisionButton = new ManifoldAuthToken();
-    provisionButton.setAuthToken = jest.fn();
+    it('calls the set auth token on load', () => {
+      const token = `test.${encodedExpiry}`;
 
-    provisionButton.token = token;
-    provisionButton.componentWillLoad();
+      const provisionButton = new ManifoldAuthToken();
+      provisionButton.setAuthToken = jest.fn();
 
-    expect(provisionButton.setAuthToken).toHaveBeenCalledWith(token);
+      provisionButton.token = token;
+      provisionButton.componentWillLoad();
+
+      expect(provisionButton.setAuthToken).toHaveBeenCalledWith(token);
+    });
+  });
+
+  describe('when the token is expired', () => {
+    const expiry = new Date();
+    expiry.setDate(expiry.getDate() - 1);
+    const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
+
+    it('calls the set auth token on load', () => {
+      const token = `test.${encodedExpiry}`;
+
+      const provisionButton = new ManifoldAuthToken();
+      provisionButton.setAuthToken = jest.fn();
+
+      provisionButton.token = token;
+      provisionButton.componentWillLoad();
+
+      expect(provisionButton.setAuthToken).not.toHaveBeenCalled();
+    });
   });
 
   it('calls the set auth token on change', () => {
--- a/src/components/manifold-auth-token/manifold-auth-token.spec.ts
+++ b/src/components/manifold-auth-token/manifold-auth-token.spec.ts
@@ -24,7 +24,7 @@ describe('<manifold-auth-token>', () => {
     expiry.setDate(expiry.getDate() - 1);
     const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
 
-    it('calls the set auth token on load', () => {
+    it('does not call the set auth token on load', () => {
       const token = `test.${encodedExpiry}`;
 
       const provisionButton = new ManifoldAuthToken();
--- a/src/components/manifold-auth-token/manifold-auth-token.spec.ts
+++ b/src/components/manifold-auth-token/manifold-auth-token.spec.ts
@@ -4,7 +4,8 @@ describe('<manifold-auth-token>', () => {
   describe('when the token is not expired', () => {
     const expiry = new Date();
     expiry.setDate(expiry.getDate() + 1);
-    const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
+    const unixTime = Math.floor(expiry.getTime() / 1000);
+    const encodedExpiry = Buffer.from(unixTime.toString()).toString('base64');
 
     it('calls the set auth token on load', () => {
       const token = `test.${encodedExpiry}`;
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -1,19 +1,25 @@
-import { h, Component, Prop, Watch } from '@stencil/core';
+import { h, Component, Prop, Watch, Event } from '@stencil/core';
 import Tunnel from '../../data/connection';
+import { EventEmitter } from 'events';
 
 @Component({ tag: 'manifold-auth-token' })
 export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() setAuthToken?: (s: string) => void;
+  @Prop() setAuthToken: (s: string) => void = () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
-  @Prop() token?: string;
+  @Prop({ mutable: true }) token?: string;
+  @Event() manifoldOauthTokenChange: EventEmitter;
 
   @Watch('token') tokenChange(newToken: string) {
-    return this.setAuthToken && this.setAuthToken(newToken);
+    this.setAuthToken(newToken);
+
+    if ((!this.token && newToken) || (this.token && !newToken)) {
+      this.manifoldOauthTokenChange.emit(newToken);
+    }
   }
 
   componentWillLoad() {
-    if (this.setAuthToken && this.token) {
+    if (this.token) {
       this.setAuthToken(this.token);
     }
   }
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -1,5 +1,4 @@
 import { h, Component, Prop, Watch, Event, EventEmitter } from '@stencil/core';
-import '@manifoldco/shadowcat';
 import { AuthToken } from '@manifoldco/shadowcat';
 import Tunnel from '../../data/connection';
 import { isExpired } from '../../utils/auth';
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -1,4 +1,5 @@
 import { h, Component, Prop, Element, Watch } from '@stencil/core';
+import { defineCustomElements } from '@manifoldco/shadowcat/dist/loader';
 import Tunnel from '../../data/connection';
 
 @Component({ tag: 'manifold-auth-token' })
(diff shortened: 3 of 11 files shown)


src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -20,8 +19,22 @@ export class ManifoldAuthToken {
     }
   }
 
+  async componentDidLoad() {
+    document.addEventListener('receiveManifoldToken', (e: CustomEvent) => {
+      this.token = e.detail;
```
davidleger95: so right now we're getting the _not-a-token_ from the test server. we should update that value to be a real token from dashboard. idk if we can automate that somehow to get us a new token each time. At the very least, we can add it to an env var, so we can just update it in the Manifold project and redeploy when it needs a new one?

davidleger95: the problem with getting a new token each time is if multiple people are using it at the same time, everyone's tokens will expire as soon as a new person loads a new token. :/

batgirl: redeploy dashboard? or which part? it sounds like `puma` is almost ready to be returning auth tokens so we may want to switch to that soon, in the meantime whatever fast way of testing a real auth token with the test server works for me (I imagined just copy pasta-ing but I guess that's not secure).

src/global/app.ts
```diff
@@ -0,0 +1 @@
+import '@manifoldco/shadowcat';
```
batgirl: 😮wow! that's cool

src/components/manifold-auth-token/manifold-auth-token.tsx
davidleger95: no, just redeploy the test server. this is assuming it just has a single, hard-coded, token it can deliver. that's not an ideal solution though

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -20,8 +19,22 @@ export class ManifoldAuthToken {
     }
   }
 
+  async componentDidLoad() {
```
batgirl: what makes this component need to be async?

src/components/manifold-auth-token/manifold-auth-token.tsx
davidleger95: I'm excited for puma :)

src/global/app.ts
davidleger95: Yeah! Someone in the Stencil Slack helped me out with this

src/components/manifold-auth-token/manifold-auth-token.tsx
davidleger95: oh, oops. was trying to fix the promise type issue but then ended up not using it. I'll take that out

batgirl: oh I thought there was functionality I didn't know about 😅

src/components/manifold-auth-token/manifold-auth-token.tsx
batgirl: oohhh, use the manifold project to automate that? I haven't worked much with manifold (lol). that single hard coded token is currently the case - I feel like changing it sort of depends on if we want to keep the test server around for demos or e2e tests or something, or if we'll retire it after getting puma running. 🤷‍♀ 

it makes me think of our ideas for automated testing this flow though... no plans for it right now, but could we run a test that uses the test server every time? or set up a thin proxy like Guillaume suggested that sends back a token we could test with...... 🤔just brainstorming

davidleger95: yeah, I'm also just brainstorming. We'll have to figure that out before merging this though

sslotsky: Shouldn't you be able to add an `onReceiveManifoldToken` prop to the component rather than adding this listener? e.g.

```ts
        <manifold-oauth onReceiveManifoldToken={this.someHandlerFunction} oauthUrl="https://manifold-shadowcat-test-server.herokuapp.com/signin/oauth/web"></manifold-oauth>
```

davidleger95: oh right! forgot about those props that get generated. I'll give that a try

sslotsky:
@batgirl @davidleger95 I added a commit which I'll explain here, since I thought that would be easiest.

## Overview

The main point is to raise an event that can alert a platform partner if the token has changed from null to present, or from present to null. 

### But Why?

This allows a Platform to save the token wherever/however they want. Global variable, local storage, cookie. Doesn't matter. It's worth mentioning that _render doesn't need this_. @Minivera suggested that future platforms might want control of how the token is stored so that they can feel more confident that the security of the token is in their hands. If it's easy to make that happen, we might as well do it up front. But we should keep Render's immediate needs centered here in the event that we find problems with this.

### Implications

When a platform user loads the page, the platform may already have a token saved from a previous oauth dance. They can use that token as a prop to `manifold-auth-token`. That component will render our `manifold-oauth` component, which is going to result in a new token being set in the connection, but that token will not be broadcast to the platform if a token was already present. (@Minivera, I'm rethinking this part. Wouldn't it be better for the platform to always have the latest token? What's the downside?)

Because the platform may provide us with an expired token, it's possible that a component will attempt to make an authenticated call before oauth has completed and retrieved a new token. That will result in a 401, which we'll have to decide how to handle as we convert our calls to GraphQL.

### Other changes in this commit

Function props must either be optional or have default values, according to our lint rules. `manifold-oauth-token` was previously using an optional prop for `setAuthToken`. I changed it to a default no-op value instead.

sslotsky:
Also realized that we were using local storage as a backing store for our previously retrieved token, but we weren't setting it in local storage when we retrieve it! Added another commit to fix that.

davidleger95:
> That component will render our manifold-oauth component, which is going to result in a new token being set in the connection, but that token will not be broadcast to the platform if a token was already present.

When a new token is created doesn't it force older ones to expire? That was my understanding of how our auth worked. If this is the case, then we will always need the latest one.

Minivera:
> (@Minivera, I'm rethinking this part. Wouldn't it be better for the platform to always have the latest token? What's the downside?)

I don't think there is any from our side, but I'd wonder about the load of constant oauth requests on our servers and the platform's servers. Should we only do the oauth dance if we know the token is invalid?

The token invalidity could be done in a few ways. One could be to encore the expiration in the token as suggested in DM, then if the token as expired, we redo the dance. Another could be to trigger a DOM event when an endpoint triggers a 401, in that case the platform should give us a `null` value as the tokens so we can redo the dance?

sslotsky:
> I'd wonder about the load of constant oauth requests on our servers and the platform's servers. Should we only do the oauth dance if we know the token is invalid?

@Minivera we might want to get advice on that. My gut says it won't be that bad, and we can adjust if it turns out to be a real problem.

src/utils/auth.ts
```diff
@@ -15,3 +15,18 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
     },
   };
 }
+
+export function isExpired(token: string) {
+  const [, expiry] = token.split('.');
```
batgirl: Does the `,` here indicate a variable you're not going to use?

sslotsky: Yes! The first part of the token is the access token itself, the second is the expiry. So `split` returns an array where I only care about the second item.

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -1,26 +1,46 @@
...
-      this.setAuthToken(this.token);
+    if (this.token) {
+      if (!isExpired(this.token)) {
+        this.setAuthToken(this.token);
+      }
+    }
+  }
+
+  setInternalToken(e: CustomEvent) {
+    const payload = e.detail as AuthToken;
```
batgirl: Will this payload always be in the AuthToken format? I'm not sure how you usually handle type issues like that in TS. Is that sort of what you mentioned in your platform channel update?

src/utils/auth.ts
```diff
@@ -15,3 +15,18 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
...
+
+  if (expiry) {
+    const d = new Date(Buffer.from(expiry, 'base64').toString());
```
batgirl: This line looks a little different from line 30 in `manifold-auth-token` - is this one decoding while the other one is encoding?

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: Yeah it will be

https://github.com/manifoldco/shadowcat/blob/master/src/components/manifold-oauth/manifold-oauth.tsx#L13

:point_up: That's what shadowcat emits every time. Notice this line:

```ts
  @Event() receiveManifoldToken: EventEmitter<AuthToken>;
```

That ensures that the data we emit from shadowcat is always in the shape of the `AuthToken`. What I was talking about this morning has more to do with optional props. Currently, all the fields on the `AuthToken` object are optional. I'd like to look into a way to tighten that up so that we can expect only two possibilities:

1. We are guaranteed to have both a `token` and an `expiry` and no errors
2. We have an error and no `token` or `expiry`

If this were something like ReasonML you would do it sort of like this:

```reason
type token = string;
type expiry = string;
type error = { code: number, message: string };

type token =
  | Success(token, expiry)
  | Failure(error);
```

sslotsky: I know there's a way to do something like that in TypeScript, but it's a bit more verbose and I don't remember off the top of my head how it looks.

src/utils/auth.ts
sslotsky: Precisely

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -2,16 +2,29 @@ import { h, Component, Prop, State } from '@stencil/core';
...
 export class ManiTunnel {
   /** _(optional)_ Specify `env="stage"` for staging */
   @Prop() env: 'stage' | 'prod' = 'prod';
 
-  @State() authToken?: string = localStorage.getItem('manifold_api_token') || undefined;
+  @State() authToken?: string = getDefaultToken();
 
   setAuthToken = (token: string) => {
     this.authToken = token;
+    localStorage.setItem(TOKEN_KEY, token);
```
batgirl: Here we're setting the token to localStorage by default, but since the authToken is optional that lets the platform choose to handle it differently if they prefer, right? Will there be any errors from calling localStorage here on or `line 10` if they don't want to use our default token handling?

sslotsky: This should work because if the platform wants to store the token themselves, then they will render `<manifold-auth-token>` and pass their token in as a prop. That's going to cause `setAuthToken` to be called in the `manifold-auth-token` component, which will overwrite the value that we're storing in the `manifold-connection`.

src/components/manifold-auth-token/manifold-auth-token.spec.ts
```diff
@@ -1,16 +1,40 @@
...
+      expect(provisionButton.setAuthToken).toHaveBeenCalledWith(token);
+    });
+  });
+
+  describe('when the token is expired', () => {
+    const expiry = new Date();
+    expiry.setDate(expiry.getDate() - 1);
+    const encodedExpiry = Buffer.from(expiry.toUTCString()).toString('base64');
+
+    it('calls the set auth token on load', () => {
```
sslotsky: Woops this should say 'does not call the set auth token on load'

src/utils/auth.ts
```diff
@@ -4,14 +4,36 @@
...
+    const now = new Date();
+
+    if (d > now) {
```
sslotsky: Maybe this should be `>=`

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -1,26 +1,46 @@
...
     }
   }
 
   render() {
-    return <input type="hidden" value={this.token ? 'true' : 'false'} />;
+    return (
+      <div>
+        <manifold-oauth
+          onReceiveManifoldToken={this.setInternalToken}
+          oauthUrl="https://manifold-shadowcat-test-server.herokuapp.com/signin/oauth/web"
```
seriousben: This should be passed in from `manifold-auth-token` no? Or maybe it should be based on the env.

seriousben:
Sorry I asked a few questions in https://manifoldco.slack.com/archives/CDW4B94RK/p1563569842030800

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: Yeah ultimately we might want this URL in `manifold-connection`, good point

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -1,26 +1,46 @@
...
+        this.setAuthToken(this.token);
+      }
+    }
+  }
+
+  setInternalToken(e: CustomEvent) {
+    const payload = e.detail as AuthToken;
+    if (!payload.error) {
+      const encodedExpiry = Buffer.from(payload.expiry || '').toString('base64');
+      this.token = `${payload.token}.${encodedExpiry}`;
```
seriousben: What if token contains a `.`? It is not a JWT right now but what if we switched to a JWT in the future?

Why not making `token = payload` and saving a `JSON.stringify()` in `localStorage`?

seriousben:
Added minor comments. LGTM ✅

src/components/manifold-auth-token/manifold-auth-token.tsx
batgirl: There are places we are testing `<manifold-oauth>` outside of `ui` (for instance in `puma`), where we want to denote a specific testing URL - not sure if that affects what you mean here. 🤷‍♀

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: I like it. One thing to consider though is that whatever we choose to store in localStorage, it will be the same thing that we pass in the DOM event to the platform for them to store themselves. Giving them stringified JSON in that event data feels a little weird. Of course, giving them a delimited string doesn't feel _a lot_ better. Maybe a little though. What do you think?

package.json
```diff
@@ -65,6 +65,7 @@
     "@graphql-codegen/typescript-operations": "^1.3.1",
     "@manifoldco/eslint-plugin-stencil": "^0.2.1",
     "@manifoldco/icons": "0.0.3",
+    "@manifoldco/shadowcat": "^0.1.3",
```
dangodev: shadowcat is now a required dependency for UI, so this should be in deps (the `import` in a component makes it so).

Now that this is required by UI, we should be aware of how this affects the bundle size, if at all.

src/global/app.ts
dangodev: I’d recommend against using this, if possible. This will load shadowcat universally, but it will slow down the overall load time for UI. Are there any heavy dependencies shadowcat has?

dangodev: Put another way, we’re already `import`ing it from `manifold-auth-token`, so this shouldn’t be necessary.

To check the impact, compare thr size of `dist/loader/loader.mjs` to [the current file](https://unpkg.com/@manifoldco/ui@0.4.4/dist/esm/loader.mjs)

src/utils/auth.ts
dangodev: Is this affected by timezones?

package.json
sslotsky: Thanks for catching this

src/utils/auth.ts
sslotsky: Hum. That's a darn good question.

package.json
```diff
@@ -65,6 +65,7 @@
     "@graphql-codegen/typescript-operations": "^1.3.1",
     "@manifoldco/eslint-plugin-stencil": "^0.2.1",
     "@manifoldco/icons": "0.0.3",
+    "@manifoldco/shadowcat": "^0.1.5",
```
dangodev: This (still) needs to be moved to `deps` in order to work

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -1,26 +1,47 @@
-import { h, Component, Prop, Element, Watch } from '@stencil/core';
+import { h, Component, Prop, Watch, Event, EventEmitter } from '@stencil/core';
+import '@manifoldco/shadowcat';
```
dangodev: What is this part doing?

davidleger95: probably should be removed since the next line imports the component

sslotsky: You mean this one?

```ts
import { AuthToken } from '@manifoldco/shadowcat';
```

That's only importing a type definition. I can try removing the import on line 2 and see if things still work. AFAIK though that's how you go about using one component library from another.

sslotsky: I removed the import from line 2 and it seems to still work, so that's good. Not sure what's up with this test failure.

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -2,24 +2,46 @@ import { h, Component, Prop, State } from '@stencil/core';
...
```
Minivera: We're locking all our platforms into using localStorage without them knowing we do so, are we sure this is something we want?

src/utils/auth.ts
```diff
@@ -4,14 +4,37 @@
...
     headers: {
       ...((options && options.headers) || {}),
-      authorization: `Bearer ${authToken}`,
+      authorization: `Bearer ${token}`,
     },
   };
 }
+
+export function isExpired(token: string) {
+  const [, expiry] = token.split('.');
```
Minivera: Access tokens might be in multiple parts if they're a JWT, with the dashboard login token for example. I think it'd be safer here to take the last element with `pop` rather than destructure.

package.json
dangodev: @sslotsky @batgirl did this ever get moved to `deps`?

PR #261, opened 2019-07-12
[WIP] REST 👉 GraphQL (Product Logo Component)

davidleger95:
## Reason for change

- Converting the fetch call from REST to GraphQL

## Testing

- go to Product Logo and check that the logo still displays properly.
- you can also check out the network calls to see what's being requested/returned


src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -18,7 +30,7 @@ export class ManifoldDataProductLogo {
   @Prop() productLabel?: string;
   /** Look up product name from resource */
   @Prop() resourceName?: string;
-  @State() product?: Catalog.Product;
+  @State() product?: any;
```
davidleger95: @dangodev is there an easy way to get the proper types for the GraphQL endpoint like we have in Sphericon?

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -33,16 +45,30 @@ export class ManifoldDataProductLogo {
...
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ query, variables: { productLabel } }),
+      })
+    );
+    const { data, error } = await response.json();
+    if (error) {
+      console.error(error);
+    }
+    this.product = data.product;
```
davidleger95: This should be abstracted out into a reusable function.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -33,16 +45,30 @@ export class ManifoldDataProductLogo {
...
   };
```
davidleger95: `fetchResource` also needs to be converted over. I image the query will look something like this:
```gql
query PRODUCT_LOGO($resourceLabel: String!) {
  resource(label: $resourceLabel) {
    product {
      displayName
      logoUrl
    }
  }
}
```

However we don't have resources in GraphQL yet

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
dangodev: Yup! The `generate:gql` script in Sphericon uses `@graphql-codegen/cli` which I’d recommend. Just copy deps, the `config.yml` file, and the script:

```json
  "generate:gql": "graphql-codegen --config codegen.yml",
```

davidleger95: thanks!

davidleger95: @batgirl Would you like to give that a try while I clean up the rest of this? moving the graphql-codegen stuff over from Sphericon

src/utils/graphqlFetch.ts
```diff
@@ -0,0 +1,28 @@
...
+    },
+    body: JSON.stringify({ query, variables }),
+  });
+
+  const body = await response.json();
+
+  return body;
+};
+
+export default graphqlFetch;
```
davidleger95: I'm wondering if we should just pass this down to the functions through `manifold-connection`? that way we have fetch injected and we can set the endpoint based on the `env` and set the auth token right from there too. 

I'd convert this to a factory function then to pass in the options from the connection.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
sslotsky: TBH it sorta bugs me that this component supports fetching the logo for both a product and a resource. What if I supply both? Then it will make two calls and the second will override the first. They could be totally different products, too. That makes it unclear how this component should be used. 

I realize it was like that already, but maybe there's no time like the present for splitting this out?

sslotsky: Alternatively, we could make a separate PR to split them out and merge that one before this.

davidleger95:
@batgirl @sslotsky I think this is probably blocking you from doing graphql work. If that's the case, I think we should merge this now so you can get started. I'll clean up the merge conflict. Do you agree?

sslotsky:
> @batgirl @sslotsky I think this is probably blocking you from doing graphql work. If that's the case, I think we should merge this now so you can get started. I'll clean up the merge conflict. Do you agree?

Is it possible to at least get the graphql codegen in there?

davidleger95:
> Is it possible to at least get the graphql codegen in there?

Yes @sslotsky. I can add if after we merge this: https://github.com/manifoldco/ui/pull/268

batgirl:
@sslotsky @davidleger95 codegen PR is merged!

package.json
```diff
@@ -55,6 +55,7 @@
     "test:watch": "stencil test --spec --e2e --watchAll"
   },
   "dependencies": {
+    "@manifoldco/gql-zero": "^0.2.0",
```
davidleger95: should this be moved to devDependencies @sslotsky? (I think the answer is yes because everything else is in there)

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -50,17 +64,26 @@ export class ManifoldDataProductLogo {
...
     const productResp = await fetch(`${catalog}/products/${productId}`, withAuth(this.authToken));
     const product: Catalog.Product = await productResp.json();
-    this.product = product;
+
+    // NOTE: Temporary util GraphQL supports resources
+    const newProduct = {
+      displayName: product.body.name,
+      logoUrl: product.body.logo_url,
+    };
+    this.product = newProduct as Product;
```
davidleger95: had to do `as Product` because we don't have the full product. This is one issue with our types is that they confirm that some fields are present when they may not be requested in the query.

package.json
dangodev: No this happens at runtime, so this would actually need to be in deps. Without it, you’d probably get a `Can’t find @manifoldco/gql-zero` error.

davidleger95: I'll just put that back then 🙃

dangodev: We don’t need `@stencil/core` at runtime (only build-time), so that’s not necessary to install. We actually don’t need `@stencil/state-tunnel`, either, but that’s in deps because of types needed in a React/TypeScript setup.

If you add something to deps it doesn’t necessarily mean it’ll be loaded into the client; it’s more of a question of: when they run `npm i @manifoldco/ui`, are we installing things into their `node_modules` directory they don’t really need?

src/utils/graphqlFetch.ts
Minivera: i think we should, that way we can leverage the auth token directly stored in `manifold-connection`

davidleger95: Cool, I'm working on that now :)

PR #262, opened 2019-07-12
Add a product card that can fetch data to fill the service card

Minivera:
## Reason for change
The `manifold-service-card` had no "fetch" option and could only be created using the product's data. Most platforms won't have that data available and so the `manifold-product-card` can fill that gap.

## Testing
Test with storybook


docs/docs/components/manifold-service-card.md
```diff
@@ -21,6 +21,15 @@ Compact view of a Manifold Product.
 ></manifold-service-card>
 ```
 
+## Fetching using the label
+
+If you don't have the data for the product and would rather fetch using the product's label, the
+`manifold-product-card` component can be used instead.
+
+```html
+<manifold-service-card product-label="logdna"></manifold-service-card>
```
dangodev: `manifold-product-card` ?

Minivera: Arf, good catch

docs/docs/components/manifold-service-card.md
```diff
@@ -21,6 +21,15 @@ Compact view of a Manifold Product.
...
 
+## Fetching using the label
+
+If you don't have the data for the product and would rather fetch using the product's label, the
+`manifold-product-card` component can be used instead.
+
+```html
+<manifold-product-card product-label="logdna"></manifold-product-card>
+```
+
```
davidleger95: I wonder if we should just recommend product card over service card. Service card seems like it's just a `*-view` component now that's used in multiple places.

Minivera: Service card has a smart part to it where it fetches if a product is free. However, I think I can port that over the marketplace and the product card. I really dislike how we now have three components for displaying the product card, so I'll be merging two of them before I get this PR in.

dangodev:
Screen Shot 2019-07-15 at 08 25 53

I’m good with this in general but we’re not using `var(--manifold-font-family)` here so the typeface isn’t correct.

Minivera:
Look up the docs, the font-familly works. I don't know why it doesn't display as it should in some stories 🤔

dangodev:
> Look up the docs, the font-familly works. I don't know why it doesn't display as it should in some stories 

That’s because the docs are loading the CSS, and it’s cascading down. Actually we probably need to enforce the Manifold CSS for Storybook anyway; ignore my comment. I think this is good to merge!

PR #267, opened 2019-07-16
Add the ability for the deprovision and rename button to render unstyled

Minivera:
## Reason for change
~~By rendering the button using the `as` attribute, you can either render it as a submit button or an anchor, this allows the button to be effectively unstyled.~~

See my comment about the shadow dom, it has been removed from both components so they can now be styled.

## Testing
Test using the docs, the `resource-details` page uses the new render as attribute.

### DIFF
--- a/docs/docs/resource-details.md
+++ b/docs/docs/resource-details.md
@@ -16,7 +16,12 @@ a `manifold-resource-container` component with access to the resource label.
   <!--- Your structure goes here --->
 </manifold-resource-container>
 ```
-
+<style>
+  manifold-data-rename-button button, manifold-data-deprovision-button button {
+    border: none;
+    padding: 0;
+  }
+</style>
 <manifold-mock-resource>
   <div style="display: flex; justify-content: space-between; margin-bottom: 1em;">
     <div style="display: flex; align-items: center;">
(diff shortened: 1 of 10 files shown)


sslotsky:
Does it mess with semantics a little bit to use an anchor for something that's supposed to function as a button? And did you consider just supporting some kind of `theme` or `unstyled` property?

Minivera:
I considered adding a an `unstyled` property, but that would have required adding css or inline style, which I wasn't too keen on. While the components are named `buttons`, I feel like they could be anything ask long as they're clickable, since `data` components are about the data, not the styling.

Sadly, because I have to use shadowdom to be able to use slots properly with these components, we can't style the internal button from the exterior 😞 I'd personally prefer to not have to need this PR.

I feel like this is a good in-between, since both anchors and buttons are clickable in my mind.

Minivera:
The shadow dom on the data button was apparently to make the slots work, however they work without it. This means we don't have to use the shadow dom on the data buttons or the wrapper component, which allows external styling! 🎉

davidleger95:
> The shadow dom on the data button was apparently to make the slots work, however they work without it.

I was just going to say this haha 😃

docs/docs/resource-details.md
```diff
@@ -16,16 +16,25 @@ a `manifold-resource-container` component with access to the resource label.
   <!--- Your structure goes here --->
 </manifold-resource-container>
 ```
-
+<style>
+  manifold-data-rename-button button, manifold-data-deprovision-button button {
+    border: none;
+    padding: 0;
+  }
+</style>
```
davidleger95: Is this necessary to have in the UI project? couldn't this just be handled by the platform?

Minivera: That's in the docs to make the example unstyled :)

It removes the border around buttons here: 

Capture d’écran 2019-07-16 à 11 20 46

davidleger95: oh haha, didn't look at the file this was in 😅

PR #268, opened 2019-07-16
Added codegen script, yml and dependencies and generated types and sc…

batgirl:
…hema

## Reason for change
Added `codegen` generated types and schema for graphql in typescript.

## Testing
Run `npm run generate:gql` at the command line and check the generated files:
`src/spec/graphql.schema.json`
`src/types/graphql.ts`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -18,3 +18,4 @@ src/types/catalog.ts
 src/types/gateway.ts
 src/types/marketplace.ts
 src/types/provisioning.ts
+src/types/graphql.ts
--- a/codegen.yml
+++ b/codegen.yml
@@ -1,0 +1,11 @@
+overwrite: true
+schema: 'https://api.manifold.co/graphql'
+documents: '{,!(node_modules)/**/}*.graphql'
+generates:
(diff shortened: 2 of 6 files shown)


batgirl:
Screen Shot 2019-07-16 at 10 06 32 AM
The linter is failing because of `src/types/graphql.ts` but since it's generated I'm not sure what I should change..

batgirl:
@dangodev thank you!! I'm not sure I understand the purpose of the JSON which makes me wary to remove it.

Do you have an opinion on the linting? Did that happen in sphericon also, or are there different linting rules there?

dangodev:
> I'm not sure I understand the purpose of the JSON which makes me wary to remove it.

I think it’s only used as a middle-file to generate the TS types. It’s not hurting anything, but it’s also a big file we may not need stored in VC. Not a big deal either way, though.

> Do you have an opinion on the linting? Did that happen in sphericon also, or are there different linting rules there?

Yup! I had to add all generated files to `.eslintignore`. No point in linting code we’re not maintaining.

PR #269, opened 2019-07-16
Fix the deprovision and rename button click

Minivera:
## Reason for change
When used outside stencil, the `handleClick` method didn't have the right `this` bound, this fix will make sure it works and the cases added are tested.

## Testing
Test with storybook.

### DIFF
--- a/src/components/manifold-resource-deprovision/manifold-resource-deprovision.tsx
+++ b/src/components/manifold-resource-deprovision/manifold-resource-deprovision.tsx
@@ -1,4 +1,4 @@
-import { h, Component, Prop } from '@stencil/core';
+import { h, Component, Prop, Event, EventEmitter } from '@stencil/core';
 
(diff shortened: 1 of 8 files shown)


src/components/manifold-resource-deprovision/manifold-resource-deprovision.tsx
```diff
@@ -7,13 +7,19 @@ import { Gateway } from '../../types/gateway';
 export class ManifoldResourceDeprovision {
   @Prop() data?: Gateway.Resource;
   @Prop() loading: boolean = true;
+  @Event({ eventName: 'manifold-deprovisionButton-click', bubbles: true }) click: EventEmitter;
```
Minivera: Events do not bubble up through slots, so using a ref on this component did not catch the `manifold-data-deprovision-button`'s event

PR #275, opened 2019-07-17
Add the curly rule to our eslint config

Minivera:
The title says it all. Ran prettier and eslint on all our files to fix.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,7 +55,7 @@ module.exports = {
     'lines-between-class-members': 'off', // class members don’t need that space!
     'max-len': 'off', // let Prettier decide
     'object-curly-newline': 'off', // let Prettier decide,
-    'curly': 'error',
+    "curly": ["error", "all"],
     'prettier/prettier': 'error',
   },
 };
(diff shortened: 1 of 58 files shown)


.eslintrc.js
```diff
@@ -55,5 +55,6 @@ module.exports = {
     'lines-between-class-members': 'off', // class members don’t need that space!
     'max-len': 'off', // let Prettier decide
     'object-curly-newline': 'off', // let Prettier decide,
+    'curly': 'error',
```
dangodev: Since a lot of these are prettier errors, should we add `'prettier/prettier': 'error'` as well?

src/utils/plan.ts
```diff
@@ -233,11 +233,11 @@ export function initialFeatures(features: Catalog.ExpandedFeature[]): Gateway.Fe
     if (!feature.value) return obj;
 
     if (feature.type === 'boolean')
-      return { ...obj, [feature.label]: booleanFeatureDefaultValue(feature.value) };
+      {return { ...obj, [feature.label]: booleanFeatureDefaultValue(feature.value) };}
```
dangodev: This seems weird? How come the `{}` aren’t on their own lines?

Minivera: Heum... Not sure, gonna fix that 😅

PR #276, opened 2019-07-17
Converting product page to graphql

batgirl:
***BREAKING:*** Chloe realized in writing this PR that since the data handling is happening in `<manifold-product>` the graphql should probably be there as well, of course! So I guess converting the product-page means converting the other components that are connected to it/wrapping it first..does that seem true? I want to post this for review just so people can give me feedback on my attempt though, and I'll see if doing it in the product component makes more sense. :)

## Reason for change
Adding a graphql query and new fetch statement to the `<manifold-product-page>`. Got a little confused on how to know where the variables for the query are coming from, unrelated to graphql but just as far as data handling goes (I assume it should get passed in as a prop?).

## Testing
Storybook renders the `product-page` inside of `product` which uses fetch and catalog to pass data to the `product-page`, so testing there isn't accurate to this component. Might mean we need a new story, or to add graphql to `product`, or might very well mean I misunderstood how the data in this component was supposed to work.

### DIFF
--- a/src/components/manifold-product-details/manifold-product-details.tsx
+++ b/src/components/manifold-product-details/manifold-product-details.tsx
@@ -44,7 +44,7 @@ export class ManifoldProductDetails {
           </span>
         </h1>
         <ul class="value-prop-list" itemprop="description">
-          {value_props.map(({ body, header }) => (
+          {valueProps.map(({ body, header }) => (
             <li class="value-prop" key={header}>
               <h3>
                 <manifold-skeleton-text>{header}</manifold-skeleton-text>
--- a/src/components/manifold-product-details/manifold-product-details.tsx
+++ b/src/components/manifold-product-details/manifold-product-details.tsx
@@ -1,5 +1,5 @@
 import { h, Component, Prop } from '@stencil/core';
-import { Catalog } from '../../types/catalog';
+import { Product } from '../../types/graphql';
 import skeletonProduct from '../../data/product';
 
 @Component({
@@ -8,12 +8,10 @@ import skeletonProduct from '../../data/product';
   shadow: true,
 })
 export class ManifoldProductDetails {
-  @Prop() product?: Catalog.Product;
+  @Prop() product?: Product;
 
   render() {
-    const {
-      body: { tagline, value_props, images = [] },
-    } = this.product || skeletonProduct;
+    const { tagline, valueProps, screenshots = [] } = this.product || skeletonProduct;
 
     if (this.product) {
       return (
@@ -22,15 +20,15 @@ export class ManifoldProductDetails {
             <span class="tagline">{tagline}</span>
           </h1>
           <ul class="value-prop-list" itemprop="description">
-            {value_props.map(({ body, header }) => (
+            {valueProps.map(({ body, header }) => (
               <li class="value-prop" key={header}>
                 <h3>{header}</h3>
                 <p>{body}</p>
               </li>
             ))}
           </ul>
-          {images.length > 0 && (
-            <manifold-image-gallery images={images}>Screenshots</manifold-image-gallery>
+          {screenshots.length > 0 && (
+            <manifold-image-gallery images={screenshots}>Screenshots</manifold-image-gallery>
           )}
         </div>
       );
--- a/src/components/manifold-product-page/manifold-product-page.tsx
+++ b/src/components/manifold-product-page/manifold-product-page.tsx
@@ -1,31 +1,8 @@
-import { h, Component, Prop, Watch } from '@stencil/core';
-import { gql } from '@manifoldco/gql-zero';
+import { h, Component, Prop } from '@stencil/core';
 import { arrow_up_right, book, life_buoy } from '@manifoldco/icons';
-import graphqlFetch from '../../utils/graphqlFetch';
 import { Product, Provider } from '../../types/graphql';
 import skeletonProduct from '../../data/product';
-import Tunnel from '../../data/connection';
-// import { withAuth } from '../../utils/auth';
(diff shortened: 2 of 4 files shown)


src/components/manifold-product-page/manifold-product-page.tsx
```diff
@@ -1,30 +1,86 @@
...
+    }
+
+    this.product = undefined;
+    this.provider = undefined;
+    const variables = { productId };
+    const { data, error } = await graphqlFetch({ query, variables });
+    if (error) {
+      console.error(error);
+    }
+    this.product = data.product;
```
batgirl: the `<manifold-product-details>` below did not like this data type, but I guess that just means I would have to convert that component to gql at the same time as well

src/components/manifold-product-page/manifold-product-page.tsx
```diff
@@ -1,30 +1,86 @@
...
+      logoUrl
+      categories {
+        label
+      }
+      provider {
+        displayName
+      }
+    }
+  }
+`;
```
davidleger95: Query looks good! 💯 

As you alluded to in your description, the network calls should be handled by the wrapper component, `manifold-product`. Sorry, this component has many more pieces than the example one I did with the logo. You'll notice this pattern with a lot of the UI/view (as opposed to data components) components. We wanted to separate the network calls from the UI components to make testing easier. (And if we discover some UI/view components don't have this wrapper and make network calls directly in the view component, we should try to refactor it to use this pattern along the way.)

Basically with all these components we want to replace the contents of the `fetchWhatever()` functions in the wrapper components with calls to the GraphQL API and then adjust the props of the components underneath as needed.

So for example, in this component we have both `product` and `provider` because with REST, we had to make 2 separate calls for that data. With GraphQL, we can eliminate `provider` because it's part of `product` now.

src/components/manifold-product-page/manifold-product-page.tsx
davidleger95: Yes, if the underlying components don't match the GraphQL data we should update them. The only place where we have to be careful is with components that are documented on https://ui.sandbox.manifold.co/ because those are meant to be consumed externally. Any that aren't documented should be fine to change the props. 

We should still add a changelog entry though to communicate internal API changes for our own benefit, because perhaps @Minivera is using the undocumented components somewhere?

Minivera: We should consider all components to be external as all components are exported by stencil and added to the window. Those we didn't document needs to be and we can't really assume no-one will use them sadly.

src/components/manifold-product-details/manifold-product-details.tsx
```diff
@@ -8,12 +8,10 @@ import skeletonProduct from '../../data/product';
...
 })
 export class ManifoldProductDetails {
-  @Prop() product?: Catalog.Product;
+  @Prop() product?: Product;
 
   render() {
-    const {
-      body: { tagline, value_props, images = [] },
-    } = this.product || skeletonProduct;
+    const { tagline, valueProps, screenshots = [] } = this.product || skeletonProduct;
```
batgirl: This line in the linter is saying these properties don't exist on a graphql Product, but I can clearly see they do. Anyone have ideas of why? 🤔

batgirl: @davidleger95 do you know why this might not be working based on what I have so far?

davidleger95: `skeletonProduct` is still referring to `Catalog.Product` so the types don't line up there

batgirl: Ooooohhhhh thank you

davidleger95:
Is this still WIP @batgirl?

dangodev:
@batgirl what’s the status on this? I have availability to help with this if needed. It’d be good to get this in this week!

src/components/manifold-product-details/manifold-product-details.tsx
dangodev: `skeletonProduct`’s type can be changed, if that’s helpful. It’s a loading state for a product, so there’s no reason for it to be a different type. Either the type changed and this one was left, or it was a mistake that should probably be corrected.

batgirl:
@dangodev this was on hold while we finished more pressing Render priorities. But now I should be free to finish it up! It can definitely be finished early this week!

src/components/manifold-product-details/manifold-product-details.tsx
batgirl: @dangodev Do you mean that I should replace the `Catalog.Product` skeleton with a graphql `Product` skeleton? My instinct is to just create a second graphql skeleton and temporarily leave the catalog version in place in case it is being used elsewhere, since they have different body structures.

batgirl: Nevermind, I see why this is a silly question :P ignore me

dangodev: I think so. I’d prefer to not duplicate code if possible, especially not for a loading state.

> temporarily leave the catalog version in place in case it is being used elsewhere, 

I hear you, and I really appreciate that you’re thinking about this in a backwards-compatible way! But I think it’s important to verify if it is being used elsewhere, and know for certain. Backwards compatibility is great, but if we don’t need it, it’s debt we shouldn’t incur.

dangodev:
Based on feedback we’ve gotten around our graphql endpoint being slow, I’m going to close this WIP PR (but I’ll keep the branch for future reference). Basically it’ll be a while before we can merge this, and in that time this will become further diverged from the codebase.

PR #277, opened 2019-07-18
add graphql fetch to manifold connection

davidleger95:
## Reason for change

- add graphql fetch to manifold connection

## Testing

- Go to product logo and check that it loads


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -16,11 +17,14 @@ export class ManiTunnel {
...
-        connection: connections[this.env],
-        authToken: this.authToken,
-        setAuthToken: this.setAuthToken,
-      }}>
+      <Tunnel.Provider
+        state={{
+          connection: connections[this.env],
+          authToken: this.authToken,
+          setAuthToken: this.setAuthToken,
+          graphqlFetch: createGraphqlFetch({ authToken: this.authToken }),
```
davidleger95: I'm not sure why this isn't working. it ends up `undefined` inside the component. Tried a few different things and still no luck. going to keep trying things but please make suggestions if you have any.

sslotsky: What ends up as `undefined`? The `graphqlFetch` function?

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -26,11 +26,16 @@ export class ManifoldDataProductLogo {
...
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() authToken?: string;
+  /** _(hidden)_ Passed by `<manifold-connection>` */
+  @Prop() graphqlFetch?: <T>(body: GraphqlRequestBody) => GraphqlResponseBody<T>;
   /** URL-friendly slug (e.g. `"jawsdb-mysql"`) */
   @Prop() productLabel?: string;
   /** Look up product name from resource */
   @Prop() resourceLabel?: string;
   @State() product?: Product;
+  @Watch('graphqlFetch') graphqlFetchChange() {
```
sslotsky: I'm confused about how this function would change. Are you expecting it to go from undefined to defined?

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -39,18 +44,18 @@ export class ManifoldDataProductLogo {
...
 
   componentWillLoad() {
-    if (this.productLabel) this.fetchProduct(this.productLabel);
+    // if (this.productLabel) this.fetchProduct(this.productLabel);
     if (this.resourceLabel) this.fetchResource(this.resourceLabel);
   }
 
   fetchProduct = async (productLabel: string) => {
-    if (!this.connection) {
+    if (!this.graphqlFetch) {
```
sslotsky: This seems like a case where we should prefer default props over optional props.

src/components/manifold-connection/manifold-connection.tsx
davidleger95: yes

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: I was thinking the token being redefined might trigger it because we would need to create a new graphqlFetch. idk if that would actually trigger it though depending on what the watch pays attention to. it's still the same function but a new version of it, if that makes sense.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: I was thinking potentially we wouldn't have this defined until the token is set. I didn't get that far into the implementation though.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
sslotsky: Hmm, I see. We're initializing the fetch with a token, but I think that it should be more like a closure, where the fetch always has access to what the current token is.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
sslotsky: We'll still have some public endpoints, so we won't want to wait for a token to initialize it. As per a comment I made elsewhere, the custom fetch should just have access to the current token on demand.

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: so you mean we should move the graphqlFetch definition into manifold-connection so it's in lexical scope and we don't need all this `createGraphqlFetch` business?

sslotsky: Yeah I think that's one possibility. Another possibility is to require any components that are calling `graphqlFetch` to pass in the token. Maybe there are other options too.

src/components/manifold-connection/manifold-connection.tsx
davidleger95: 🤦‍♂ there's no `manifold-connection` on the docs site...

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: Yeah, I wanted to eliminate the amount of things needed to be passed in by manifold-connection. if they're both in the connection I feel like we should simplify it as much as possible there. I can see adding a watch is just shifting that complexity though. I think adding graphqlFetch in the scope of the connection is the simplest way to achieve what I want. Maybe I can do some fancy dynamic scoping though to keep the function separate without needing a factory function. maybe I just need to bind `this` to the fetch function

sslotsky: > maybe I just need to bind this to the fetch function

Classic trick!

davidleger95: Not sure why but the bind didn't work. I came up with a workaround though. it still doesn't quite work but it's getting closer :)

src/components/manifold-badge/mf-badge.css
```diff
@@ -22,5 +23,6 @@
 }
 
 :host([data-tag='free']) {
+  color: var(--free-text-color);
```
davidleger95: just a small theming fix while I'm at it

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -2,25 +2,35 @@ import { h, Component, Prop, State } from '@stencil/core';
...
       <Tunnel.Provider
         state={{
           connection: connections[this.env],
           authToken: this.authToken,
           setAuthToken: this.setAuthToken,
+          graphqlFetch: createGraphqlFetch({
+            getAuthToken: this.getAuthToken,
+            setAuthToken: this.setAuthToken,
+            endpoint,
+          }),
```
davidleger95: So this is just the first phase of making this work. I think once the shadowcat change is merged we can build off that to make the auth flow run smoothly. What I'm thinking is setting flags in the connection that components can watch for to decide when to wait to fetch, when to fetch, when to refetch, etc... and also add logic in the `manifold-auth-token` to decide when to get a new token (maybe @sslotsky is already working on that?).

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -20,77 +16,52 @@ const query = gql`
...
     this.product = data.product;
   };
 
-  fetchResource = async (resourceLabel: string) => {
-    if (!this.connection) {
-      return;
+  render() {
+    if (this.resourceLabel) {
+      console.warn('DEPRCATION WARNING: Use `manifold-data-resource-logo` instead.');
+      return <manifold-data-resource-logo resourceLabel={this.resourceLabel} />;
```
Minivera: Very good! Could you also add an warning toast in the docs saying that this deprecation is in effect starting in UI v0.5.0 and also add a `Deprecated` thing in the changelog?

dangodev: Also `DEPRCATION` is misspelled—can this be fixed in another PR?

src/components/manifold-data-resource-logo/manifold-data-resource-logo.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  render() {
+    return this.product ? (
+      <img src={this.product.logoUrl} alt={this.alt || this.product.displayName} />
+    ) : null;
+  }
+}
+
+Tunnel.injectProps(ManifoldDataResourceLogo, ['connection', 'authToken']);
```
dangodev: Why was this a separate component, rather than a new prop on `manifold-data-product-logo`?

src/utils/graphqlFetch.ts
```diff
@@ -1,28 +1,57 @@
...
+  getAuthToken = () => undefined,
+  setAuthToken = () => {},
+}: CreateGraphqlFetch = {}): (<T>(body: GraphqlRequestBody) => Promise<GraphqlResponseBody<T>>) => {
+  const graphqlFetch = async <T>(
+    requestBody: GraphqlRequestBody
+  ): Promise<GraphqlResponseBody<T>> => {
+    const response = await fetch(endpoint, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
```
dangodev: Our GraphQL endpoint sometimes errs on `Content-type` (capital). I think it prefers `content-type` (lowercase)

src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
davidleger95: 🤦‍♂ 🤦‍♂ 🤦‍♂ not again...

src/components/manifold-data-resource-logo/manifold-data-resource-logo.tsx
davidleger95: It was getting overly complex with resources not using graphql yet. maybe once everything is on graphql we can merge these components back together (or not, just a possibility)? for now it was easier to keep it simple by separating the components.

PR #278, opened 2019-07-18
Add lazyloading instructions

dangodev:
## Reason for change
Screen Shot 2019-07-24 at 11 13 02

A platform reported an entry bundle increase from adding our web components. This changes the instructions to prefer lazyloading by default.

This PR adds a **Performance** section to docs, outlining some tips on improving the speed of consuming UI.


dangodev:
cc @jeffandersen @mattcreager just for visibility this is up

jeffandersen:
The title says instructions, but really it's just showing an example. Is there a reason we don't have a section in the readme or docs about loading and the different methods? or what the impact is?

I ask this because even our own internal team included it in a way that caused the bundle to inflate. Perhaps if there was more explanation we could have avoided this? Unsure

dangodev:
> Is there a reason we don't have a section in the readme or docs about loading and the different methods? or what the impact is?

Yeah that’s a good suggestion. At first thought it felt like this falls under more general front-end bundling knowledge, but you’re right that if internally we’re not checking for it, that means others probably aren’t, either. Even if it’s outside the scope of UI, just adding a small section shows we’re thinking about it.

dangodev:
@jeffandersen added a performance section to the docs, along with an “Advanced” section where we can put more guides

docs/docs/advanced/resource-details.md
```diff
@@ -1,21 +1,24 @@
...
 
-A specific set of component are available to use for creating a resource details view without having to keep
-giving the resource label as attributes to all the components in the library.
+A specific set of component are available to use for creating a resource
+details view without having to keep giving the resource label as attributes
+to all the components in the library.
 
-To setup a resource details view, wrap all the structure you want to be part of the resource details view in
-a `manifold-resource-container` component with access to the resource label.
+To set up a resource details view, wrap all the structures you want to be
```
dangodev: typo: `setup`

docs/docs/advanced/resource-details.md
```diff
@@ -42,10 +45,11 @@ a `manifold-resource-container` component with access to the resource label.
   <manifold-resource-credentials></manifold-resource-credentials>
 </resource-mock-resource>
 
-## The resource credentials
+## Resource credentials
```
dangodev: Removed the word “the” from titles because it’s wasteful of the Primacy Effect

docs/src/components/Sidebar.tsx
```diff
@@ -299,6 +310,8 @@ const Nav = styled.nav`
     width: 15rem;
     height: 100vh;
     padding: 1.5rem;
+    -webkit-overflow-scrolling: touch;
+    overflow-y: auto;
```
dangodev: Fixes sidebar not scrolling on laptops

PR #291, opened 2019-07-24
Add field to set API token in docs

davidleger95:
## Reason for change

Make it easy to update the token in the docs so that people don't have to go into the console to set the token.

## Testing

- Go to docs and set token
- check local storage to see that token is set correctly
- ensure examples fetch with token


dangodev:
Screen Shot 2019-07-24 at 17 22 19

I’m not a big fan of having it at the top of every page, and pushing the content down.

What if we had it be part of the example component? (We could also make that **Example** heading better too). That way there’s a direct association between that token input, and the component that’ll be using it.

Plan___Manifold_UI___Manifold_UI_and_Desktop

docs/src/components/ApiToken.tsx
```diff
@@ -0,0 +1,88 @@
+import React, { FormEvent, useState, ChangeEvent } from 'react';
+import styled from 'styled-components';
+
+const Form = styled.form`
+  display: grid;
+  grid-gap: 0.5rem;
+  grid-template-columns: 1fr auto;
+  padding: 1rem;
+`;
```
dangodev: Nice! I like that this refreshes the page on submit.

docs/src/components/ApiToken.tsx
```diff
@@ -0,0 +1,92 @@
...
+    filter: brightness(90%);
+  }
+`;
+
+const ApiToken = () => {
+  const [token, setToken] = useState('');
+
+  useEffect(() => {
+    setToken(localStorage.getItem('manifold_api_token') || '');
+  }, []);
```
davidleger95: had to do this because gatsby build doesn't have localStorage access

dangodev: Ah that makes sense

docs/src/components/Example.tsx
```diff
@@ -32,13 +34,24 @@ const Inner = styled.div`
...
 interface ExampleProps {
   html: string;
 }
 
 const Example: React.FunctionComponent<ExampleProps> = ({ html }) => (
-  <Wrapper>
-    <h3>Example</h3>
+  <Wrapper id="example">
+    <Heading>Example</Heading>
+    <ApiToken />
```
Minivera: Could we make this configurable in the frontmatter? Pages for the marketplace or product/plans data are mocked.

davidleger95: Yeah, I think that makes sense. Better experience than having to go get a token or implement auth in docs. I'll give that a try

davidleger95: @Minivera I'm looking into this and wondering if it's worth the effort to source from the REST endpoint when we'll be moving everything to GraphQL eventually. I'm thinking maybe we should just source the GraphQL API and convert the docs to static data as we convert components. That way we're not doing rework on docs.

That would mean this PR for setting the API key could be a temporary solution until everything is converted over. Does that make sense?

cc: @dangodev

sslotsky:
Should this be closed?

dangodev:
Well, we should either do this, or mock resource creation/credential/resizing for docs. Maybe the latter?

sslotsky:
Yeah, we should be mocking them. Every example on the docs site should work without requiring authentication. https://github.com/manifoldco/adrs/blob/master/0028-use-mock-for-docs-to-make-it-stable-in-ui.md

PR #292, opened 2019-07-24
Fix service card view naming

davidleger95:
## Reason for change

- fix the naming on the service card view
- @TerrySmithDC discovered the docs for this aren't correct. The docs seem like they should be correct, so I changed the code instead of the docs


src/components/manifold-service-view/manifold-service-view.css
```diff
@@ -1,3 +1,4 @@
+/* DEPRICATED: replaced by manifold-service-card-view */
```
dangodev: Why not just delete this file?

src/components/manifold-service-view/manifold-service-view.tsx
```diff
@@ -1,3 +1,4 @@
+/* DEPRICATED: replaced by manifold-service-card-view */
```
dangodev: If this component is deprecated, maybe throw a `console.error('<manifold-service-view> is deprecated. Please use <manifold-service-card-view> instead.')` message

src/components/manifold-service-view/manifold-service-view.css
dangodev: Oh actually rather than using this CSS file, just link to `manifold-service-card-view.css` in the other component

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed the provision button requiring the label to be set, preventing or automatic label generation from working.
+- Fixed name for `manifold-service-view` to be `manifold-service-card-view` to match documentation.
```
davidleger95: @Minivera You'll want to update your usage of this to the new component once this is released. You'll get a deprecation warning in the console for the old one.

PR #293, opened 2019-07-24
Docs auto header links

dangodev:
## Reason for change
Screen Shot 2019-07-24 at 17 07 26

Adds autolinked headers for docs

## Testing
Make sure the heading links work in the preview link


davidleger95:
Merge conflict on `docs/package-lock.json`

docs/gatsby-browser.js
```diff
@@ -6,3 +6,11 @@
...
 // Loads base theme before styled-components tag in head.
 import '@manifoldco/ui/dist/manifold/manifold.css';
+
+export function shouldUpdateScroll({ routerProps: { location } }) {
+  if (location.hash) {
+    return false;
+  }
+
+  return true;
+}
```
dangodev: @Minivera this is the code that’s supposed to work, but doesn’t. Gatsby claims this API still works, so maybe it only needs a tiny tweak? At any rate this is supposed to control scroll behavior.

davidleger95: the theme query param is blocking route changes. I have the same issue when trying to add a hash route to auto scroll to the examples after refreshing the token

docs/src/components/Sidebar.tsx
```diff
@@ -69,9 +69,13 @@ function Sidebar({ pages }: SidebarProps) {
     if (typeof window !== 'undefined') {
       const params = new URLSearchParams(window.location.search);
       params.set('theme', theme);
-      window.history.replaceState(null, '', `${window.location.pathname}?${params.toString()}`);
+      window.history.replaceState(
+        null,
+        '',
+        `${window.location.pathname}?${params.toString()}${window.location.hash}`
```
davidleger95: hashes need to go before the params

davidleger95:
Is this PR still a WIP, on hold, or in some other state?

dangodev:
@davidleger95 I thought you were taking this on? I said I couldn’t find a path forward with this and wanted to close it, but I think you said you wanted to take a crack at it. Feel free to close this if you don’t have time, etc.

davidleger95:
> I thought you were taking this on?

I don't think I said this haha, I can but it's lowest priority so I might not get to it this week

docs/src/components/Sidebar.tsx
davidleger95: @dangodev so this seems not worth fixing. :/

The issue seems to be that the window is locked from programmatically scrolling for some reason. I even tried this:

```js
setTimeout(() => {
    window.scrollTo(0, 200);
  }, 5000);
```

dangodev: Yeah it’s part of Reach router—the scroll resets so that focus is made when it loads. They have a poorly-documented way of overriding it, but it doesn’t seem to be working for me. Based on the GitHub issue discussions, it seems like something that frequently breaks in the codebase because maintainers forget it’s even a thing.

I’m OK not having it for now.

davidleger95: Should we close the PR then?

dangodev: Yeah I think it’s fine to close

davidleger95:
sadly not worth the effort to fix

PR #294, opened 2019-07-25
Fix the resource list e2e test by removing it

Minivera:
This replaces the resource list's flaky e2e test with a new shiny spec test that can actually test things and give us more confidence in the resource list code.


src/components/manifold-resource-list/manifold-resource-list.spec.ts
```diff
@@ -29,24 +31,87 @@ const provisionOperation: Provisioning.Operation = {
...
+        components: [ManifoldResourceList],
+        html: `
+          <manifold-resource-list>
+            <div data-test="no-resources" slot="no-resources"></div>
+          </manifold-resource-list>
+        `,
+      });
+
+      // @ts-ignore
+      expect(page.root.querySelector('[data-test="no-resources"]')).toBeTruthy();
```
dangodev: Let’s select the slot name instead here

src/components/manifold-resource-list/manifold-resource-list.e2e.ts
```diff
@@ -1,30 +0,0 @@
...
-        <manifold-resource-list>
-          <span itemprop="no-resources" slot="no-resources">No resources</span>
-        </manifold-resource-list>`,
-    });
-    await page.$eval(
-      'manifold-resource-list',
-      (elm: any, props: any) => {
-        elm.resources = props.resources; // TODO: This doesn't actually work, you can't set the state in tests
-      },
-      { resources: [] }
```
dangodev: This test is breaking because you can’t set state externally like a prop

src/components/manifold-resource-list/manifold-resource-list.spec.ts
dangodev: Even better—can we have the text `No resources` appear, and can we query an element by the text `No resources`? Then we can guarantee that the text is passing through properly.

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -53,8 +53,8 @@ export class ManifoldResourceList {
   }
 
   componentWillLoad() {
+    this.fetchResources();
```
dangodev: Good catch. This makes more sense to me

src/components/manifold-resource-list/manifold-resource-list.e2e.ts
dangodev: Is there a way to repair this test rather than delete it? E2E tests are more valuable than spec tests because they render the shadow dom, and actually run the `render()` function.

When we were talking about testing the render function, E2E is the way to do it. Keeping our E2E tests is how we maintain confidence in our components. Spec tests, by comparison, are useful and easy-to-write but *do not* test rendering in a browser-like environment. I’d argue removing E2E tests reduces confidence, not raises it.

Minivera: This test was breaking for many reasons.

The test added no confidence for the reason you state, but I left it in there with a TODO so we can at least look at fixing it.

However, the fetch-mock usage in our other spec tests lead to very weird behavior where the `window.fetch` function might be mocked when puppeteer cones the window object and we can't control that in any way.

The `newSpecPage` function _does_ call the render function on component like the e2e tests do, but it doesn't do it in a real browser, only a weird mocked version. Stencil recommends to switch to those tests and we can be confident that they give us a very similar level of test quality as the e2e tests. https://stenciljs.com/docs/unit-testing#newspecpage-

The only thing the spec page does not give us is the ability to test our components in a render context, but this is covered by happo tests. I think we can move away from e2e tests with confidence.

PR #298, opened 2019-07-25
Parsing authToken as a JWT

batgirl:

### DIFF
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -24,7 +24,7 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
 }
 
 export function isExpired(token: string) {
-  const [, expiry] = token.split('.');
+  const expiry = token.split('.').pop();
 
   if (expiry) {
     const decodedExpiry = parseInt(Buffer.from(expiry, 'base64').toString(), 10);
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -24,17 +24,15 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
 }
 
 export function isExpired(token: string) {
-  const expiry = token.split('.').pop();
-
-  if (expiry) {
+  try {
+    const expiry = token.split('.').pop() || '';
     const decodedExpiry = parseInt(Buffer.from(expiry, 'base64').toString(), 10);
+
     const d = new Date(decodedExpiry * 1000);
     const now = new Date();
 
-    if (d > now) {
-      return false;
-    }
+    return d < now;
+  } catch (error) {
+    return true;
   }
-
-  return true;
 }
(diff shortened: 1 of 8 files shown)


src/utils/auth.ts
```diff
@@ -24,7 +24,7 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
 }
 
 export function isExpired(token: string) {
-  const [, expiry] = token.split('.');
+  const expiry = token.split('.').pop();
```
sslotsky: One thing that crosses my mind - and maybe it's nothing to worry about? - is if the token is malformatted, this might not do what we think. e.g.

```
"blah".split(".").pop()
"blah"
```

So that check on line 29 for `if (expiry)` would always be truthy. Thoughts @batgirl @Minivera ?

davidleger95: is a badly formatted token something we should expect? How would that happen?

Minivera: It might happen if someone tampers with their localstorage? 🤷‍♂ 

I don't think this is an issue, all that will happen here is that the endpoints will all 401.

davidleger95: @sslotsky is correct though. `if(expiry)` will always be truthy. Is there a way to check that the expiry is actually an expiry? maybe `if (typeof expiry === 'number')` or `if(new Date(expiry))`

sslotsky: Line 30 or 31 could blow up if it's not. We might need to wrap that in a try/catch instead of a conditional.

davidleger95: yeah, that sounds good to me

src/utils/auth.ts
```diff
@@ -24,17 +24,18 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
...
-  const [, expiry] = token.split('.');
+  try {
+    const expiry = token.split('.').pop() || '';
+    const decodedExpiry = parseInt(expiry, 10);
+    if (Number.isNaN(decodedExpiry)) {
+      throw new Error('Expiry not a number.');
+    }
 
-  if (expiry) {
-    const decodedExpiry = parseInt(Buffer.from(expiry, 'base64').toString(), 10);
```
davidleger95: @sslotsky Buffer was throwing an error. is there any reason why a buffer is being used here? It's just a normal string isn't it?

sslotsky: It's because we were base64 encoding the expiry info. I suppose we don't have to do that if we're using these unix stamps. It will simplify things.

davidleger95: so it's fine without it then and I can safely remove it?

sslotsky: Yeah, let's remove all the base64 logic and just use the timestamp itself (keeping the division though of course)

PR #301, opened 2019-07-26
Use the profile graphQL endpoint if ownerId is not set

Minivera:
## Reason for change
This adds the ability for the provision button to fetch the profile ID using graphQL for the provisioning, making the owner id not required.

If the ownerId is still given, the component will not send a fetch call.

## Testing
Test with storybook, you should see the graphql call be sent and fail in the console.


src/utils/connections.ts
```diff
@@ -5,6 +5,7 @@ export interface Connection {
   marketplace: string;
   provisioning: string;
   connector: string;
+  graphQl: string;
```
dangodev: This seems inconsistent to me. I think `graphql`, `gql`, or `graphQL` should work, but not `graphQl`. Could you use one of the first three options?

CHANGELOG.md
```diff
@@ -5,23 +5,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+### Added
+- Added `oauth-url` prop to `manifold-auth-token` component.
+
+### Fixed
+- Fixed name for `manifold-service-view` to be `manifold-service-card-view` to match documentation.
+- Added missing support for theme variable `--manifold-tag-free-text-color`.
 
 ### Deprecated
-- Deprecated `resource-label` attibute on `manifold-data-product-logo`. Use `manifold-data-resource-logo` component instead.
+- Deprecated `resource-label` attribute on `manifold-data-product-logo`. Use `manifold-data-resource-logo` component instead.
```
dangodev: 👍

PR #302, opened 2019-07-26
🌲@logger decorator

dangodev:
## Reason for change
Front-end Fantasia onsite project: we need the ability for UI to be able to handle error logging. This adds the **ability** for us to log components, but **it does not actually implement error logging in all components** (followup PR). This PR’s purpose is to gather feedback and merge, then we can port the decorator to all components.

## Testing
- [ ] Tests should pass
- [ ] [Error page](https://manifold-ui-stage-pr-302.herokuapp.com/advanced/errors/?theme=default) should werd good


src/utils/logger.e2e.tsx
```diff
@@ -0,0 +1,45 @@
...
+  it('doesn’t affect existing render() function', () => {
+    const myComponent = new MockComponent();
+    expect(myComponent.render()).toBe('success');
+  });
+
+  it('reports errors if render() fails', async () => {
+    const browser = await puppeteer.launch();
+    const page = await browser.newPage();
+
+    // https://github.com/GoogleChrome/puppeteer/blob/master/examples/custom-event.js
```
dangodev: This is the test that’s failing currently. I want to be able to test 2 things:

1. That the `manifold-error` event is called
2. That it’s called with the error message passed to it

I know we can do this with Chrome Puppeteer, but not sure how exactly to complete this test. But once that test satisfies these 2 things, I’d be fine with merging this!

src/utils/logger.e2e.ts
```diff
@@ -0,0 +1,36 @@
...
+      new MockComponent().render('oops');
+    } finally {
+      // do nothing
+    }
+
+    const [eventDetail] = errorListener.mock.calls[0];
+
+    expect(eventDetail.bubbles).toBe(true); // event should bubble
+    expect(eventDetail.detail).toEqual({ error: 'oops' }); // event should contain original error message
+  });
```
dangodev: I tried testing this in Puppeteer, but couldn’t. Hopefully JSDom is good enough?

docs/docs/advanced/errors.md
```diff
@@ -0,0 +1,36 @@
...
+
+Our components also log out the error to `console.error()` as well, which
+many APM tools such as Datadog and Rollbar will automatically pick up on. For
+many of these tools, you should see error reporting automatically happen
+there with no action on your part.
+
+## Manual reporting
+
+If you’d optionally like to handle errors yourself, reporting can be handled
+safely via the `manifold-error` [custom event][custom-event]. This same
```
sslotsky: Whoa! It took me a minute there to figure out how the `custom-event` and `toast` links were working. TIL!

docs/docs/advanced/errors.md
```diff
@@ -0,0 +1,36 @@
...
+custom event fires for all components:
+
+```js
+document.addEventListener('manifold-event', e => {
```
sslotsky: Should this be `'manifold-error'`?

src/components/manifold-active-plan/manifold-active-plan-happo.ts
```diff
@@ -11,3 +11,13 @@ export const jawsDB = () => {
 
   return plan.componentOnReady();
 };
+
+export const planError = () => {
+  const plan = document.createElement('manifold-active-plan');
+  plan.plans = fromJSON(product); // pass the wrong data model here to throw an error
```
sslotsky: OMG :clap: :clap: :clap: :clap: :clap: :clap: :clap:

src/utils/logger.e2e.ts
sslotsky: Let's iterate! We can take another crack at testing puppeteer later. For now I'm good with this. There are at least a couple follow up points:

1. We obv. need to get this added to all components and introduce a lint rule that requires it, etc
2. We might want to add optional arguments to the decorator that give us more specific output or modify the event details.

Both of those things could be opportunities to try puppeteer again.

src/utils/logger.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export default function logger<T>() {
+  return function loggerDecorator(
+    _target: T,
+    _propertyKey: string,
+    descriptor: PropertyDescriptor
+  ) {
+    const originalMethod = descriptor.value;
+
+    descriptor.value = function render(...args: T[]) {
```
sslotsky: What are the args to render, and what is `T`?

docs/docs/advanced/errors.md
dangodev: Yeah I’m a fan of doing this in markdown in general. Keeps the text much cleaner, and allows you to reuse links (which I do a lot)

src/utils/logger.tsx
dangodev: I’m actually not sure… this part was in here by day 2 when I joined

sslotsky: From what I can tell, `render` doesn't take any args. Can we try without the args and without the generic signature and see if it still works?

TerrySmithDC: I think that's left over from noodling, pretty sure it's safe to remove.

docs/docs/advanced/errors.md
TerrySmithDC: TIL

src/utils/logger.e2e.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('passes render() contents through', () => {
+    const render = new MockComponent().render();
+    expect(render).toBe('success');
+  });
+
+  it('reports errors if render() fails', async () => {
+    const errorListener = jest.fn();
+    window.addEventListener('manifold-error', errorListener);
+
+    try {
```
sslotsky: I forgot to ask, why is the `try` needed? Shouldn't the error be caught within `render`?

dangodev: This is something that Puppeteer could help with. I think the `try` wouldn’t be needed if we were actually testing our components. But here, because the class is in our mock component in our test scope, throwing an error here will throw the test. I just tried it and the test doesn’t pass without the `try`.

src/components/manifold-active-plan/manifold-active-plan-happo.ts
dangodev: Hm this doesn’t actually work, it seems. It only shows the skeleton loader. It fails to re-render with the skeleton screen already showing, and doesn’t replace it with the toast.

src/utils/logger.tsx
dangodev: Hm. Actually it throws the test with this removed. I don’t see the harm in keeping it; this seems like the proper way to call a function in general 🤷‍♂

src/utils/logger.e2e.ts
dangodev: Actually scratch that—we don’t need this! This was leftover from Puppeteer. I thought I needed it because I tried after removing `args` from `logger`, but that was the actual reason. This works without this block.

src/components/manifold-active-plan/manifold-active-plan-happo.ts
sslotsky: :thinking: 

That's really odd. I wonder if we can confirm that this still happens in a simpler case. e.g. we have a component that takes a `shouldThrowError` prop that's `false` by default. If that's set to `false` it displays a `Hello world` or something. If it's set to true, then it throws, catches, and returns something else.

Waiting for `componentOnReady` seems to usually work fine when we have to add some props after creating the component. So I'd expect it to work the same here. But maybe this is somehow different when there's a try/catch in render.

dangodev: It seems for errors that happen async like this, the decorator just doesn’t do anything—doesn’t log the error to the console, doesn’t show the error…

I suspect it’s because for async errors, that all happens before `render()` is even called so it doesn’t have an opportunity to intercept it.

dangodev: For this test I’ll just have to think of a different way to throw a rendering error

src/utils/logger.tsx
sslotsky: It really smells off though. The code suggests that `render` takes multiple args, and that all args are of the same type. What does it throw?

src/components/manifold-active-plan/manifold-active-plan-happo.ts
sslotsky: I'm confused, how is this an async error?

dangodev: Oh well maybe “async” isn’t the best term here, but the component allows for `plan?` and `product?` to be undefined so it’ll go ahead with rendering the skeleton UI whether it has data or not, then try to re-render once it receives data.

In Storybook with real data, I confirmed that `render()` isn’t even being called if an error happens with bad data. It gets stuck in the methods that happen before the VDOM diff, and before the render update. But I agree with you that I’m a little confused why Happo even renders the skeleton UI in this test.

src/utils/logger.tsx
sslotsky: Hmm, I think I see the problem actually. I think it's because our mock component's render method takes an argument.

sslotsky: What we should probably be doing here is changing `errorMsg` to be a class property and passing it into the constructor. e.g.

```ts
class MockComponent {
  errorMsg?: string;

  constructor(errorMsg?: string) {
    this.errorMsg = errorMsg;
  }

  @logger()
  public render() {
    if (this.errorMsg) {
      throw new Error(this.errorMsg);
    }

    return 'success';
  }
}
```

src/components/manifold-active-plan/manifold-active-plan-happo.ts
sslotsky: > In Storybook with real data, I confirmed that render() isn’t even being called if an error happens with bad data.

That's.... unexpected. Why the heck wouldn't it be called!?

src/utils/logger.tsx
```diff
@@ -0,0 +1,32 @@
...
+      try {
+        const result = originalMethod.apply(this, args); // call render()
+        return result;
+      } catch (e) {
+        console.error(e); // report error (Rollbar, Datadog, etc.)
+        const detail: ErrorDetail = { error: e.message };
```
sslotsky: It occurs to me that we probably need a bit more information here than just the error message. It should also at least include the name of the component that it's coming from. We could see if this information is available in `_target` or something. If it's not, it could be passed into the `logger` function.

src/components/manifold-active-plan/manifold-active-plan-happo.ts
dangodev:
```
TypeError: "newPlans[0] is undefined"
    plansChange manifold-active-plan.entry.js:45
```

Error if I give `manifold-plan-selector` bad data, for instance. It throws in the `planChange` watch method, before `render()` is called

dangodev: So basically we’re only catching errors within `render()`, but errors are still likely to happen in any other component method without us knowing

sslotsky: Ohh! That totally makes sense.

Now it feels like we ultimately want to expand the scope of this so that we can catch errors that are thrown in watch methods and lifecycle methods as well. That's probably out of scope for this PR and it might be hard to do cleanly, but it's definitely a place where an error could occur if we don't get data in the shape that the code is expecting, which is a big part of what we're trying to account for here.

dangodev:
Screen Shot 2019-07-29 at 15 15 27

yussssss

src/components/manifold-active-plan/manifold-active-plan-happo.ts
```diff
@@ -11,3 +11,18 @@ export const jawsDB = () => {
...
   return plan.componentOnReady();
 };
+
+export const planError = () => {
+  const plan = document.createElement('manifold-active-plan');
+  plan.plans = fromJSON(plans);
+  plan.product = fromJSON(product);
+  plan.selectedResource = {
+    get region() {
+      throw new Error('oops'); // strangely-specific render() error for this component
```
dangodev: As far as I could tell, whenever you’re passing references on to child Stencil components, `render()` will just pass those on without evaluating those or anything. As far as `render()` is concerned, once it renders the HTML it’s through (I agree with this, FWIW—I think it’s smart).

So the only way we’d ever throw a `render()` error is in that magic space between `render() {` and `return`. This injects a function as an object property that will throw an error.

On the one hand, I’m excited that we are able to test this. But on the other, at this point I now think this decorator doesn’t cover, well… most of our rendering needs for components. We should really be decorating _every_ class method, not just `render()`.

src/utils/logger.tsx
dangodev: Added a component name (which will log as the class name, e.g. `ManifoldActivePlan`). Also added to the test.

src/components/manifold-active-plan/manifold-active-plan-happo.ts
dangodev: Yeah I’m now realizing what little % of our problems occur within `render()`.

src/utils/logger.tsx
dangodev: Also you were spot-on—it was available in `target`.

src/components/manifold-active-plan/manifold-active-plan-happo.ts
dangodev: Also figured out why Happo rendered this OK—we’re only decorating `manifold-active-plan`, and when you give it bad data, the error occurs in the child components (outside their `render()` functions, too, no less). 

Stencil components can’t tell if their children err because that’s how web components / HTML works! It’s designed to be very fault-tolerant, and that’s proven true here.

docs/docs/advanced/errors.md
```diff
@@ -0,0 +1,36 @@
...
+title: Errors
+path: /advanced/errors
+---
+
+# Errors
+
+## Automatic reporting
+
+Any error our components have while rendering logs out to `console.error()`.
+Many frontend APM tools such as Datadog and Rollbar will automatically pick
```
davidleger95: should Datadog and Rollbar link off to their websites?

docs/docs/advanced/errors.md
davidleger95: NOt sure if that specifically is in here but there are a lot of lesser-known markdown tips here: [Mastering Markdown (a course by Wes Bos)](https://masteringmarkdown.com/) :)

src/utils/logger.tsx
davidleger95: maybe it's the type of `render()`?

src/components/manifold-active-plan/manifold-active-plan-happo.ts
sslotsky: > So the only way we’d ever throw a render() error is in that magic space between render() { and return

Hmm, I'm not sure about that. What if you had a `parent-component` with a render method that looked like this:

```ts
return <child-component baz={foo.bar.baz} />;
```

And what if your type definition missed the fact that `bar` is optional? In that case I do believe the `parent-component` would throw in render.

src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -41,6 +41,7 @@ export class ManifoldActivePlan {
     return this.plans.find(({ id }) => id === this.selectedPlanId);
   }
 
+  @logger<ManifoldActivePlan>()
```
sslotsky: Do we actually need to specify `T` here? I notice in your test code you leave it unspecified, which suggests that TS is able to infer the type.

src/utils/logger.tsx
```diff
@@ -0,0 +1,42 @@
...
+    target: T & StencilComponent,
+    _propertyKey: string,
+    descriptor: PropertyDescriptor
+  ) {
+    const originalMethod = descriptor.value;
+
+    descriptor.value = function render() {
+      try {
+        const result = originalMethod.apply(this); // call render()
+        return result;
```
sslotsky: Not a big deal or anything but we could just `return originalMethod.apply(this)`, right?

src/components/manifold-active-plan/manifold-active-plan.tsx
dangodev: Yeah I was confused because `target` had a type of `T`. I accidentally left this in after trying it, and found that it did nothing 🤷‍♂. I’ll just remove it

src/utils/logger.tsx
dangodev: Ha I had the same thought. Probably a good reason to change it, I guess 😃

src/components/manifold-active-plan/manifold-active-plan-happo.ts
dangodev: Oh yeah that’s true—I think it would throw in your scenario too.

docs/docs/advanced/errors.md
dangodev: Couldn’t hurt 🤷‍♂

src/components/manifold-active-plan/manifold-active-plan-happo.ts
```diff
@@ -11,3 +11,18 @@ export const jawsDB = () => {
 
   return plan.componentOnReady();
 };
+
+export const planError = () => {
+  const plan = document.createElement('manifold-active-plan');
+  plan.plans = fromJSON(plans);
+  plan.product = fromJSON(product);
+  plan.selectedResource = {
+    get region() {
```
sslotsky: This is clever, well done :clap: :clap:

PR #304, opened 2019-07-26
use gateway

domenicrosati:
WDYT ?

In the backend we need this for local testing


dangodev:
I don’t have an opinion on this but @Minivera or @seriousben might! I don’t use the `local` env so it’s whatever y’all need

PR #307, opened 2019-07-27
[WIP: just an example] make it work

domenicrosati:
This is mostly from Guillaumes PR with several fundemental differneces:
1.  The problematic nature of the expiry segment code 
2. Several issues with actually using `setAuthToken` 
3. the data structure of the profile request

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -26,9 +26,11 @@ export class ManifoldAuthToken {
 
   setInternalToken = (e: CustomEvent) => {
     const payload = e.detail as AuthToken;
+    console.log(payload);
     if (!payload.error && payload.expiry) {
       const expiry = payload.expiry.toString();
       this.token = `${payload.token}.${expiry}`;
+      console.log(this.token);
     }
   };
 
@@ -37,7 +39,7 @@ export class ManifoldAuthToken {
       <div>
         <manifold-oauth
           onReceiveManifoldToken={this.setInternalToken}
-          oauthUrl="https://manifold-shadowcat-test-server.herokuapp.com/signin/oauth/web"
+          oauthUrl="http://login.arigato.tools/signin/oauth/web"
         ></manifold-oauth>
       </div>
     );
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -19,20 +19,22 @@ function getDefaultToken() {
 @Component({ tag: 'manifold-connection' })
 export class ManiTunnel {
   /** _(optional)_ Specify `env="stage"` for staging */
-  @Prop() env: 'stage' | 'prod' = 'prod';
+  @Prop() env: 'local' | 'stage' | 'prod' = 'prod';
 
   @State() authToken?: string = getDefaultToken();
 
   setAuthToken = (token: string) => {
     this.authToken = token;
-    localStorage.setItem(TOKEN_KEY, token);
+    const accessToken = token.split('.', token.split('.').length - 1).join('.');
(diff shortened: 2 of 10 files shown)


src/utils/connections.ts
```diff
@@ -18,6 +19,7 @@ export const connections: Connections = {
     marketplace: 'http://api.marketplace.arigato.tools/v1',
     provisioning: 'http://api.provisioning.arigato.tools/v1',
     connector: 'http://api.connector.arigato.tools/v1',
+    graphQl: 'http://graphql.arigato.tools/graphql',
```
jelmersnoeck: Oh this is interesting. Seeing my other comment wrgt the puma URL for the component, would it be possible to put that in here too? And then it'll automatically load the correct one based on the env?

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -26,9 +26,11 @@ export class ManifoldAuthToken {
 
   setInternalToken = (e: CustomEvent) => {
     const payload = e.detail as AuthToken;
+    console.log(payload);
     if (!payload.error && payload.expiry) {
       const expiry = payload.expiry.toString();
       this.token = `${payload.token}.${expiry}`;
+      console.log(this.token);
```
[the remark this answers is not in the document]
davidleger95: image

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -37,7 +39,7 @@ export class ManifoldAuthToken {
       <div>
         <manifold-oauth
           onReceiveManifoldToken={this.setInternalToken}
-          oauthUrl="https://manifold-shadowcat-test-server.herokuapp.com/signin/oauth/web"
+          oauthUrl="http://login.arigato.tools/signin/oauth/web"
```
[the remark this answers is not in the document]
davidleger95: I just merged a PR yesterday that makes it customizable

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -19,29 +19,29 @@ function getDefaultToken() {
...
   };
 
-  getAuthToken = () => this.authToken;
+  getAuthToken = () => this.accessToken;
 
   get accessToken() {
     if (this.authToken) {
-      const [token] = this.authToken.split('.');
+      // the auth token's last segment is its expiry
+      const token = this.authToken.split('.', this.authToken.split('.').length - 1).join('.');
```
davidleger95: we made a util function that gets the expiry. maybe we need one that gets the token too?

(Perhaps we can make a regex that separates these pieces for us as a long term solution? We'll see what the priorities are next week though)

src/utils/connections.ts
davidleger95: +1 on this

src/components/manifold-connection/manifold-connection.tsx
Minivera: This PR predates that util function, which is why that change is necessary. Rebasing on master should do the trick.

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -21,13 +23,31 @@ interface ErrorMessage {
   resourceLabel?: string;
 }
 
+interface Profile {
+  id: string;
+}
+
+interface ProfileMessage {
+  profile: Profile;
```
Minivera: 👍

src/utils/connections.ts
dangodev: Minor nitpick: can this be `graphQL` (`L` capitalized)? I’m open to `graphql`, `gql`, or `graphQL`, but IMO anything else is awkward

PR #308, opened 2019-07-29
Add bundlesize check

dangodev:
## Reason for change
I originally explored using bundlesize on UI, and decided against it because it seemed to code-split very well. However, we’ve now discovered some new ways our bundle size could expand without noticing, so this adds 2 checks in place:

- [ ] our entry bundle is super light (`esm/loader.mjs`)
- [ ] each individual component doesn’t exceed `32 kB` without warning
- [ ] other parts of Stencil don’t get bigger without notifying us

## Testing
Tests should pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "now-build": "npm run build-storybook",
     "postinstall": "cd docs && npm i",
     "prepare": "stencil build",
+    "profile": "stencil build && npx bundlesize",
     "publish": "node scripts/publish",
     "storybook": "start-storybook -p 6006",
     "test": "stencil test --spec --e2e",
--- a/package.json
+++ b/package.json
@@ -63,8 +63,13 @@
       "compression": "none"
     },
     {
-      "path": "./dist/esm/*.js",
+      "path": "./dist/esm/manifold-!(mock)*.js",
       "maxSize": "32 kB",
+      "compression": "none"
+    },
+    {
+      "path": "./dist/esm/chunk-*.js",
+      "maxSize": "66 kB",
       "compression": "none"
     }
   ],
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -55,6 +55,18 @@
     "test:debug": "node --inspect-brk node_modules/.bin/stencil test --spec --e2e --runInBand",
     "test:watch": "stencil test --spec --e2e --watchAll"
   },
+  "bundlesize": [
+    {
+      "path": "./dist/esm/loader.mjs",
```
dangodev: Test #1: critical entry size

package.json
```diff
@@ -55,6 +55,18 @@
...
+      "maxSize": "15 kB",
+      "compression": "none"
+    },
+    {
+      "path": "./dist/esm/*.js",
```
dangodev: Test #2: testing all component sizes (this is a little more generous)

package.json
dangodev: Note: for different build systems, Stencil creates a few different loaders (e.g. `cjs/`), but we don’t need to test them all; one will suffice. If one bloats out because something heavy gets loaded globally, they all will.

package.json
```diff
@@ -55,6 +56,23 @@
...
     "test:watch": "stencil test --spec --e2e --watchAll"
   },
+  "bundlesize": [
+    {
+      "path": "./dist/esm/loader.mjs",
+      "maxSize": "15 kB",
+      "compression": "none"
+    },
+    {
+      "path": "./dist/esm/manifold-!(mock)*.js",
```
dangodev: Test #2: testing individual web component chunks (this is a little more generous, in case a library is needed for an individual component)

package.json
```diff
@@ -55,6 +56,23 @@
...
+      "maxSize": "32 kB",
+      "compression": "none"
+    },
+    {
+      "path": "./dist/esm/chunk-*.js",
```
dangodev: Test #3: testing other Stencil.js chunks (not sure what’s in these exactly, but good for us to be notified if they change)

PR #309, opened 2019-07-30
HOLD: remove expiry appending code

domenicrosati:


dangodev:
@domenicrosati if you don’t want this to show up in the #guild-frontend channel, just add a `wip` label to it (same goes for the other bots too!)

PR #313, opened 2019-07-30
Add a wait of 15 seconds for the token to exist during fetches

Minivera:
Work is related to manifoldco/engineering#8858

## Reason for change
When a request for fetching data on graphql or rest is sent to the servers, all the async functions will now wait for a maximum of 15 seconds (Configurable) for the token to exist before returning an error.

All the components using rest have been modified to now use the function added to the manifold-connection. Tests have been adapted to make sure the rest and graphql functions are properly injected in components.

## Testing
Test with storybook or the docs, things should fetch as expected.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,7 +57,7 @@ module.exports = {
     'object-curly-newline': 'off', // let Prettier decide,
     'curly': ['error', 'all'],
     'prettier/prettier': 'error',
-    'no-console': 'off',
+    'no-console': ['error', { allow: ['warn', 'error'] }],
     'func-names': 'off',
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,7 +55,9 @@ module.exports = {
     'lines-between-class-members': 'off', // class members don’t need that space!
     'max-len': 'off', // let Prettier decide
     'object-curly-newline': 'off', // let Prettier decide,
-    "curly": ["error", "all"],
+    'curly': ['error', 'all'],
     'prettier/prettier': 'error',
+    'no-console': 'off',
+    'func-names': 'off',
   },
 };
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -4,8 +4,10 @@ import Tunnel from '../../data/connection';
 import { connections } from '../../utils/connections';
 import { createGraphqlFetch } from '../../utils/graphqlFetch';
 import { isExpired } from '../../utils/auth';
+import { createRestFetch } from '../../utils/restFetch';
 
 const TOKEN_KEY = 'manifold_api_token';
+const baseWait = 15000;
 
 function getDefaultToken() {
   const token = localStorage.getItem(TOKEN_KEY);
@@ -20,6 +22,8 @@ function getDefaultToken() {
 export class ManiTunnel {
   /** _(optional)_ Specify `env="stage"` for staging */
   @Prop() env: 'local' | 'stage' | 'prod' = 'prod';
+  /** _(optional)_ Wait time for the fetch calls before it times out */
+  @Prop() waitTime: number = baseWait;
 
   @State() authToken?: string = getDefaultToken();
 
@@ -28,7 +32,7 @@ export class ManiTunnel {
     localStorage.setItem(TOKEN_KEY, token);
   };
 
-  getAuthToken = () => this.authToken;
+  getAuthToken = () => (this.authToken && !isExpired(this.authToken) ? this.authToken : undefined);
 
   get accessToken() {
     if (this.authToken) {
@@ -43,13 +47,18 @@ export class ManiTunnel {
     return (
       <Tunnel.Provider
         state={{
-          connection: connections[this.env],
-          authToken: this.accessToken,
           setAuthToken: this.setAuthToken,
+          restFetch: createRestFetch({
+            endpoints: connections[this.env],
+            getAuthToken: this.getAuthToken,
+            setAuthToken: this.setAuthToken,
+            wait: this.waitTime,
+          }),
           graphqlFetch: createGraphqlFetch({
             getAuthToken: this.getAuthToken,
             setAuthToken: this.setAuthToken,
             endpoint: connections[this.env].graphql,
+            wait: this.waitTime,
           }),
         }}
       >
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -38,7 +38,7 @@ export class ManiTunnel {
 
   get accessToken() {
     if (this.authToken) {
-      const [token] = this.authToken.split('|');
+      const [token] = this.authToken.split('.');
       return token;
     }
 
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -32,11 +32,12 @@ export class ManiTunnel {
     localStorage.setItem(TOKEN_KEY, token);
   };
 
-  getAuthToken = () => (this.authToken && !isExpired(this.authToken) ? this.authToken : undefined);
+  getAuthToken = () =>
+    this.authToken && !isExpired(this.authToken) ? this.accessToken : undefined;
 
   get accessToken() {
     if (this.authToken) {
-      const [token] = this.authToken.split('.');
+      const [token] = this.authToken.split('|');
       return token;
     }
 
--- a/src/components/manifold-credentials/manifold-credentials.tsx
+++ b/src/components/manifold-credentials/manifold-credentials.tsx
@@ -2,15 +2,17 @@ import { h, Component, Prop, State, Watch } from '@stencil/core';
 
 import { Marketplace } from '../../types/marketplace';
 import ConnectionTunnel from '../../data/connection';
-import { withAuth } from '../../utils/auth';
-import { Connection, connections } from '../../utils/connections';
+import { Connection } from '../../utils/connections';
 
 @Component({ tag: 'manifold-credentials' })
 export class ManifoldCredentials {
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() connection?: Connection = connections.prod;
-  /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() authToken?: string;
+  @Prop() restFetch?: <T>(
+    service: keyof Connection,
+    endpoint: string,
+    body?: object,
+    options?: object
+  ) => Promise<T | Error>;
   @Prop() resourceLabel: string = '';
   @Prop({ mutable: true }) resourceId?: string = '';
   @State() loading?: boolean = false;
@@ -29,33 +31,43 @@ export class ManifoldCredentials {
   }
 
   fetchCredentials = async () => {
-    if (!this.connection || !this.resourceId) {
+    if (!this.restFetch || !this.resourceId) {
       return;
     }
 
     this.loading = true;
-    const { marketplace } = this.connection;
-    const response = await fetch(
-      `${marketplace}/credentials/?resource_id=${this.resourceId}`,
-      withAuth(this.authToken)
+    const response = await this.restFetch<Marketplace.Credential[]>(
+      'marketplace',
+      `/credentials/?resource_id=${this.resourceId}`
     );
-    this.credentials = await response.json();
+
+    if (response instanceof Error) {
+      console.error(response);
+      return;
+    }
+
+    this.credentials = response;
     this.loading = false;
   };
 
   async fetchResourceId(resourceLabel: string) {
-    if (!this.connection) {
+    if (!this.restFetch) {
       return;
     }
 
-    const resourceResp = await fetch(
-      `${this.connection.marketplace}/resources/?me&label=${resourceLabel}`,
-      withAuth(this.authToken)
+    const response = await this.restFetch<Marketplace.Resource[]>(
+      'marketplace',
+      `/resources/?me&label=${resourceLabel}`
     );
-    const resources: Marketplace.Resource[] = await resourceResp.json();
+
+    if (response instanceof Error) {
+      console.error(response);
+      return;
+    }
+    const resources: Marketplace.Resource[] = response;
 
     if (!Array.isArray(resources) || !resources.length) {
-      console.error(`${resourceLabel} product not found`);
+      console.error(`${resourceLabel} resource not found`);
       return;
     }
 
@@ -78,4 +90,4 @@ export class ManifoldCredentials {
   }
 }
 
-ConnectionTunnel.injectProps(ManifoldCredentials, ['connection', 'authToken']);
+ConnectionTunnel.injectProps(ManifoldCredentials, ['restFetch']);
--- a/src/components/manifold-credentials/manifold-credentials.tsx
+++ b/src/components/manifold-credentials/manifold-credentials.tsx
@@ -2,17 +2,12 @@ import { h, Component, Prop, State, Watch } from '@stencil/core';
 
 import { Marketplace } from '../../types/marketplace';
 import ConnectionTunnel from '../../data/connection';
-import { Connection } from '../../utils/connections';
+import { RestFetch } from '../../utils/restFetch';
 
 @Component({ tag: 'manifold-credentials' })
 export class ManifoldCredentials {
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() restFetch?: <T>(
-    service: keyof Connection,
-    endpoint: string,
-    body?: object,
-    options?: object
-  ) => Promise<T | Error>;
+  @Prop() restFetch?: RestFetch;
   @Prop() resourceLabel: string = '';
   @Prop({ mutable: true }) resourceId?: string = '';
   @State() loading?: boolean = false;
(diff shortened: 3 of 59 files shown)


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -4,8 +4,10 @@ import Tunnel from '../../data/connection';
 import { connections } from '../../utils/connections';
 import { createGraphqlFetch } from '../../utils/graphqlFetch';
 import { isExpired } from '../../utils/auth';
+import { createRestFetch } from '../../utils/restFetch';
 
 const TOKEN_KEY = 'manifold_api_token';
+const baseWait = 15000;
```
domenicrosati: want to add a comment on the reasoning why this number was choosen?

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -225,4 +223,4 @@ export class ManifoldDataProvisionButton {
   }
 }
 
-Tunnel.injectProps(ManifoldDataProvisionButton, ['connection', 'authToken', 'graphqlFetch']);
+Tunnel.injectProps(ManifoldDataProvisionButton, ['restFetch', 'graphqlFetch']);
```
domenicrosati: i love what you've done here. kudos

Minivera:
I could have it in multiple PRs, but the only thing I could remove from there is the wait thing, which would remove maybe a dozen lines or so :/

src/components/manifold-connection/manifold-connection.tsx
Minivera: It was chosen a bit randomly, I felt like 10 seconds was too short, but that 20 was too long. It's configurable on the connection.

domenicrosati: no worries

.eslintrc.js
```diff
@@ -55,7 +55,9 @@ module.exports = {
     'lines-between-class-members': 'off', // class members don’t need that space!
     'max-len': 'off', // let Prettier decide
     'object-curly-newline': 'off', // let Prettier decide,
-    "curly": ["error", "all"],
+    'curly': ['error', 'all'],
     'prettier/prettier': 'error',
+    'no-console': 'off',
```
dangodev: I actually prefer having the linter raise console usage. There’s been many times a `console.log()` was accidentally committed. Sure, it’s annoying to disable, but I personally think it’s helpful to say “this console.log is intentional and part of the app.” You can always `/* eslint-disable no-console */` at the beginning of a file, too, if it’s particularly console-heavy

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -29,33 +31,43 @@ export class ManifoldCredentials {
...
-      `${marketplace}/credentials/?resource_id=${this.resourceId}`,
-      withAuth(this.authToken)
+    const response = await this.restFetch<Marketplace.Credential[]>(
+      'marketplace',
+      `/credentials/?resource_id=${this.resourceId}`
     );
-    this.credentials = await response.json();
+
+    if (response instanceof Error) {
+      console.error(response);
```
dangodev: Should this `throw`?

.eslintrc.js
Minivera: Is there a way to have this error not be thrown for `console.error`?

Also, this was a warning before, so travis wouldn't fail.

dangodev: Ah. Is there a way to make it an error? I think we had it like that in other apps.

src/components/manifold-credentials/manifold-credentials.tsx
Minivera: Could be, though `console.error` will be replaced by whatever system we'll have to surface async errors to platforms, I saw this line as a placeholder of sorts. Would you prefer I switch all of them to throw?

.eslintrc.js
Minivera: Yes, and we can do `{ "allow": ["warn", "error"] }` for options!

src/components/manifold-credentials/manifold-credentials.tsx
dangodev: No, I really didn’t know! Was just asking. I’m fine with whatever works—`console.error` will log an error to Datadog or New Relic if they’re used.

dangodev:
Overall the code looks good to me! As soon as the builds pass I’d like to test out Storybook & the Docs examples.

domenicrosati:
i think the wait time needs to be longer

dangodev:
> i think the wait time needs to be longer

That sounds like TDD (timeout-driven development) to me. I’d like to avoid that if possible

src/components/manifold-connection/manifold-connection.tsx
domenicrosati: @davidleger95  mentioned that we should make these default quite a lot longer perhaps like a full minute or two?

domenicrosati:
to clarify on the gql query - it seems to be only that query that isn't behaving correctly and only when you refresh the page on the addons/new/some-provider page

src/components/manifold-connection/manifold-connection.tsx
Minivera: 15 seconds should be far more than enough to get the token from shadowcat, @sslotsky had suggested 5 or 10 seconds. I believe something is not working right with the `getAuthToken` function we pass to those endpoints.

I didn't have time to investigate that today, but I will tomorrow. I don't think increasing the wait time will do anything good except slow down any tests dependent on the timeout. I'll keep you posted.

Minivera:
The failing travis is because of timeouts that resolve in tests that have finished, I'll work on fixing that tomorrow.

For the timeout, I agree with @dangodev. This is an issue I'm also seeing in the rest endpoints and is something that needs investigating.

src/components/manifold-connection/manifold-connection.tsx
dangodev: > I don't think increasing the wait time will do anything good except slow down any tests dependent on the timeout.

This is the main point here—extending testing time will have a cascade effect of not only taking longer to ship versions, but also backing up our entire CI across projects if one repo hogs all the tasks.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -83,7 +99,7 @@ describe('<manifold-data-provision-button>', () => {
     it('will do nothing on a fetch error', async () => {
       const resourceLabel = 'new-resource';
 
-      fetchMock.mock(`${connections.prod.marketplace}/resources/?me&name=${resourceLabel}`, []);
+      fetchMock.mock(`${connections.prod.marketplace}/resources/?me&label=${resourceLabel}`, []);
```
dangodev: Do you need to `mockImplementation`? We’ve used this successfully in Sphericon before: https://github.com/manifoldco/sphericon/blob/edab29d039d400cdcb1f77c39ac9a0c311ba6dfa/src/components/BankInfoForm/__tests__/index.test.tsx#L128

(also we’re not using jest-fetch-mock for simplicity / flexibility)

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -23,9 +22,12 @@ interface ErrorMessage {
 export class ManifoldDataDeprovisionButton {
   @Element() el: HTMLElement;
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() connection?: Connection = connections.prod;
-  /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() authToken?: string;
+  @Prop() restFetch?: <T>(
+    service: keyof Connection,
+    endpoint: string,
+    body?: object,
```
dangodev: Personally I’m not a big fan of deviating from the `fetch()` API. This is difficult to remember.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -23,9 +22,12 @@ interface ErrorMessage {
...
+    options?: object
```
dangodev: This should be `RequestInit`

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
Minivera: This is what `fetch-mock` is doing, but in a much more elegant way. This is not `jest-fetch-mock`, but [fetch-mock](https://www.npmjs.com/package/fetch-mock). It works wonders.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -62,45 +64,46 @@ export class ManifoldDataDeprovisionButton {
...
 
-    const response = await fetch(
-      `${this.connection.gateway}/id/resource/${this.resourceId}`,
-      withAuth(this.authToken, {
-        method: 'DELETE',
-      })
+    const response = await this.restFetch(
+      'gateway',
+      `/id/resource/${this.resourceId}`,
+      {},
```
dangodev: Also, a suggestion: when a function reaches > 3 params, it’s usually a good idea to turn params into an object. Then memorization becomes easier, and you don’t have to have empty params like you have here. E.g.:

```
function restFetch({
  service: 'gateway',
  url: `/id/resource/${this.resourceId}`,
  options: { method: 'DELETE' },
})
```

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -23,9 +22,12 @@ interface ErrorMessage {
 export class ManifoldDataDeprovisionButton {
   @Element() el: HTMLElement;
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() connection?: Connection = connections.prod;
-  /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() authToken?: string;
+  @Prop() restFetch?: <T>(
```
dangodev: 💯! Love the generic here

src/components/manifold-data-has-resource/manifold-data-has-resource.tsx
```diff
@@ -1,17 +1,19 @@
...
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() connection?: Connection = connections.prod; // Provided by manifold-connection
-  /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() authToken?: string;
+  @Prop() restFetch?: <T>(
+    service: keyof Connection,
+    endpoint: string,
+    body?: object,
+    options?: object
+  ) => Promise<T | Error>;
```
dangodev: This should be moved into `/types/` since it’s used in multiple places. TypeScript _should_ catch mismatches, but why type the same thing over and over?

src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
```diff
@@ -26,21 +26,22 @@ interface ErrorMessage {
...
   @Prop() planId?: string = '';
   @Prop({ mutable: true }) productId?: string = '';
   @Prop() regionId?: string = globalRegion.id;
   @State() resourceId: string = '';
-  @Event({ eventName: 'manifold-manageButton-click', bubbles: true })
-  clickEvent: EventEmitter;
-  @Event({ eventName: 'manifold-manageButton-error', bubbles: true }) errorEvent: EventEmitter;
-  @Event({ eventName: 'manifold-manageButton-success', bubbles: true })
-  successEvent: EventEmitter;
+  @Event({ eventName: 'manifold-manageButton-click', bubbles: true }) click: EventEmitter;
```
dangodev: Just a note: this was specifically not named `click` because for every `@Event`, it’ll make an `onEvent` equivalent. So if we wanted to fire this specific event from the parent, we now wouldn’t be able to because Stencil’s generated `onClick` would conflict with the native element `onClick`.

Here, we still have the native `onClick`, and the Stencil, generated `onClickEvent` accessible should we need to access the event from its parent.

We may not need it here, but wanted to mention that since I spent hours figuring that out 😅

dangodev: The other names are fine, though, since there’s no native `onError` and `onSuccess` methods on elements

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
Minivera: The goal here is not to have an API identical to the `fetch()` api, but an utility function that builds the fetch options for us. 

In this case, I think it would not be productive to have the base `fetch(url, init)` api as we'd then need to export the connections and auth token from the `manifold-connection` tunnel to get that url and that would lead to a lot of reused code.

src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
Minivera: I had a lot of issues with the native `onClick` event in render with some components, as a developer, I expected the `onClick` to be the same as `manifold-manageButton-click`, since it give info when "the button is clicked".

dangodev: It won’t be–in Stencil-land, it’ll handle `<manifold-data-manage-button onClickEvent={myEvent()}>` like React would, and it’ll magically let you execute a child event in its parent context.

But it’ll handle `<manifold-data-manage-button onClick={}>` as something different—a click event on the actual custom element node, which is different than the child method.

Or maybe it will work, but in general Stencil discourages colliding with native event names like this. I believe this used to not work unless they changed something, but in general it’s probably good practice to avoid this because it will be confusing what’s happening (unlike React, Stencil doesn’t have synthetic events and you’re interacting directly with the browser, still)

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
dangodev: Ah. Hm. Well you’re right—we can still pass in the connection, but it’d still be nice to specify `body` inside `init` along with the headers. That way you can use TypeScript’s type system to help all of us not create malformed queries.

Because both are typed as `object`, I have a feeling this reduces confidence in our fetching overall.

dangodev: Again—not an issue for you specifically; I’m thinking more in the context of someone new coming in, making a requested change to UI that didn’t write this, and someone casually ✅ ing their PR, breaking our systems. This is **high, high, high-risk stuff** now, and I’d like for our system to catch things as much as possible.

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -38,9 +37,12 @@ interface ErrorMessage {
...
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() connection?: Connection = connections.prod;
-  /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() authToken?: string;
+  @Prop() restFetch?: <T>(
+    service: keyof Connection,
+    endpoint: string,
+    body?: object,
+    options?: object
+  ) => Promise<T | Error>;
```
dangodev: Another opportunity to share a type (maybe just `export` this from `restFetch` itself)?

src/components/manifold-connection/manifold-connection.tsx
Minivera: I managed to get it working with shadowcat resolving in less than a few hundred miliseconds, increasing the wait time won't be necessary.

domenicrosati: cool! I will pull down and have a look now

src/utils/restFetch.ts
```diff
@@ -0,0 +1,57 @@
...
+      ...withAuth(getAuthToken(), args.options),
+      body: JSON.stringify(args.body),
+    });
+
+    const body = await response.json();
+    if (response.status >= 200 && response.status < 300) {
+      return body;
+    }
+    if (response.status === 401) {
+      setAuthToken('');
```
domenicrosati: nice!

src/utils/graphqlFetch.ts
```diff
@@ -22,36 +23,54 @@ export interface GraphqlResponseBody<T> {
...
+  if (body.errors) {
+    body.errors.forEach((error: GraphqlError) => {
+      console.error(error.message);
+    });
+  }
 
-    return body;
-  };
+  if (body.errors && body.errors[0].type === 'unauthorized') {
+    // TODO trigger token refresh for manifold-auth-token
```
domenicrosati: nice one!

dangodev:
This PR broke our public-facing docs and our development environment for components. Fortunately we have a production environment for docs, so we were able for this PR to prevent breaking documentation for our client.

Marketplace doesn’t load for either:

- https://ui.manifold.now.sh/?path=/story/marketplace--default
- http://manifold-ui-stage.herokuapp.com/components/manifold-marketplace/?theme=default

There are 2 preview links attached to this PR. For future reference it’s a good idea to check and make sure they both work before approving / merging something. That’s why they’re there.

dangodev:
When this gets released it should at least be a minor version, because it breaks implementation of our web components as they’re sometimes used.

PR #317, opened 2019-07-30
Improve marketplace scrollspy

dangodev:
## Reason for change
`IntersectionObserver` is great for many things, and we all thought it would be great for scrollspy, but after a few months we realized it’s not. It’s inconsistent with user-expected behavior, doesn’t run at 60FPS, and has some browser-specific quirks (we’ve had issues in Firefox specifically). It’s more intended for lazyloading than scrollspies.

This replaces it with the amazing `@reach/observe-rect` library, which I’m a fan of because:
- it’s so lightweight it’s basically free
- it uses `RequestAnimationFrame` so it’s browser-throttled to 60FPS the way it should be
- it automatically memoizes to reduce updates
- using a community library means we don’t have to maintain / test this part of our code (similar to `IntersectionObserver`

**Before: active category skips and jumps around**

**After: every section is scrolled to properly**

## Testing
Test scrolling through `<manifold-marketplace>` either in Storybook or the docs site

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -199,9 +199,9 @@ export namespace Components {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
+    'regionId'?: string;
     /**
     * The label of the resource to provision
     */
@@ -1292,9 +1292,9 @@ declare namespace LocalJSX {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
+    'regionId'?: string;
     /**
     * The label of the resource to provision
     */
--- a/src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
+++ b/src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
@@ -5,7 +5,7 @@
   --grid-m: 1.5rem;
 
   /* theme */
-  --link-color: var(--manifold-text-color-secondary);
+  --link-color: var(--manifold-text-color-secondary, var(--manifold-color-gray60));
   --link-color-hover: var(--manifold-color-primary);
   --link-color-active: var(--link-color-hover);
   --heading-color: var(--manifold-grayscale-60);
@@ -77,9 +77,15 @@
     color: var(--link-color-active);
   }
 
-  &:hover,
-  &:focus {
+  &:hover {
     color: var(--link-color-hover);
+  }
+
+  &:focus:focus-visible {
+    color: var(--link-color-active);
+  }
+
+  &:not(:focus-visible):focus {
     outline: 0;
   }
 }
@@ -98,6 +104,8 @@
   left: 0;
   z-index: var(--manifold-layer-hover);
   grid-column: 1 / -1; /* always start new row, and span full row */
+  display: flex;
+  align-items: center;
   box-sizing: border-box;
   width: 100%;
   margin-top: 0;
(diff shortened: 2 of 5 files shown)


src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -163,27 +165,48 @@ export class ManifoldMarketplaceGrid {
...
+    // Update active category
+    this.activeCategory = currentCategory;
   };
 
-  private observeCategory = (el?: HTMLElement) => {
-    if (el) {
-      this.observer.observe(el);
+  private updateFilter = (e: KeyboardEvent): void => {
+    if (e.srcElement) {
+      this.filter = (e.srcElement as HTMLInputElement).value;
```
dangodev: This diff is a bug because I removed `observeCategory`. This method didn’t change

src/components/manifold-template-card/manifold-template-card.css
```diff
@@ -40,6 +40,8 @@
 }
 
 .heading {
+  display: flex;
+  align-items: center;
```
dangodev: Fixes icon alignment

src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
```diff
@@ -98,6 +104,8 @@
   left: 0;
   z-index: var(--manifold-layer-hover);
   grid-column: 1 / -1; /* always start new row, and span full row */
+  display: flex;
+  align-items: center;
```
dangodev: Fixes icon alignment

src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
```diff
@@ -77,9 +77,15 @@
...
-  &:focus {
+  &:hover {
     color: var(--link-color-hover);
+  }
+
+  &:focus:focus-visible {
+    color: var(--link-color-active);
+  }
+
+  &:not(:focus-visible):focus {
```
dangodev: Changed the behavior, now focus doesn’t change the color on desktop. It’s weird seeing a category stay highlighted while you scroll away.

src/components.d.ts
```diff
@@ -1292,9 +1292,9 @@ declare namespace LocalJSX {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
+    'regionId'?: string;
```
davidleger95: i thought I had these changes in my PR, maybe I forgot to push 🤷‍♂

src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
```diff
@@ -5,7 +5,7 @@
   --grid-m: 1.5rem;
 
   /* theme */
-  --link-color: var(--manifold-text-color-secondary);
+  --link-color: var(--manifold-text-color-secondary, var(--manifold-color-gray60));
```
davidleger95: should this use one of the `--manifold-grayscale` variables?

dangodev: Yeah that’d probably be more themable you’re right

davidleger95: also I don't think this fallback will ever be used (or is it?). I think `--manifold-text-color-secondary` falls back to `--manifold-text-color`

src/components.d.ts
dangodev: 🤷‍♂

src/components/manifold-marketplace-grid/manifold-marketplace-grid.css
dangodev: Oh you’re probably right. I was wondering why this wasn’t working. I’ll just revert this.

PR #322, opened 2019-07-31
Add @logger decorator to all components

sslotsky:
## Reason for change

This integrates the eslint rule that requires all components to decorate their render methods with `@logger()`, and updates all components to comply.

## Testing

ESLint should pass! Try checking out the branch and removing the `@logger()` decorator from any component, and then ESLint should fail!

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -199,9 +199,9 @@ export namespace Components {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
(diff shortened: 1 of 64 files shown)


src/components.d.ts
```diff
@@ -1292,9 +1292,9 @@ declare namespace LocalJSX {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
+    'regionId'?: string;
```
sslotsky: Someone must have missed these on a previous PR

PR #324, opened 2019-07-31
Fix free badges on manifold-marketplace

dangodev:
## Reason for change
The “free“ badge on products either sometimes doesn’t display, or requires more API fetches than it needs to (products appear in different categories, so if a product appears multiple times it’ll send one API request per appearance).

What’s more, it breaks when you search by name.

This changes how “free“ is fetched, making it more data-efficient, and it also persists across searching.

**Before**

**After**

## Testing
1. Make sure free tags appear on marketplace (Storybook or Docs)
2. Try searching
3. Make sure free tags are still there during/after clearing search

### DIFF
--- a/src/components/manifold-marketplace/manifold-marketplace.tsx
+++ b/src/components/manifold-marketplace/manifold-marketplace.tsx
@@ -1,4 +1,4 @@
-import { h, Component, Prop, State, Element } from '@stencil/core';
+import { h, Component, Prop, State, Element, Watch } from '@stencil/core';
 
 import { Catalog } from '../../types/catalog';
 import Tunnel from '../../data/connection';
@@ -28,10 +28,15 @@ export class ManifoldMarketplace {
   @Prop() products?: string;
   /** Template format structure, with `:product` placeholder */
   @Prop() templateLinkFormat?: string;
+  @State() freeProducts: string[] = [];
   @State() parsedExcludes: string[] = [];
   @State() parsedFeatured: string[] = [];
   @State() parsedProducts: string[] = [];
   @State() services: Catalog.Product[] = [];
+
+  @Watch('services') servicesUpdated() {
+    this.fetchFreeProducts();
+  }
 
   componentWillLoad() {
     this.parseProps();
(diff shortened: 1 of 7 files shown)


src/components/manifold-service-card/manifold-service-card.spec.ts
```diff
@@ -21,33 +21,24 @@ describe('<manifold-service-card>', () => {
...
 
-  it('formats links correctly', () => {
-    const serviceCard = new ManifoldServiceCard();
-    serviceCard.productLabel = Product.body.label;
-    serviceCard.productId = Product.id;
-    serviceCard.productLinkFormat = '/product/:product';
-
-    expect(serviceCard.href).toBe(`/product/${Product.body.label}`);
-  });
-
```
dangodev: This test was moved to the E2E test—before it was testing the function, but now we’re testing if the `href` actually exists on the render link (we no longer test implementation details)

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -63,24 +69,18 @@ export class ManifoldServiceCard {
     return this.productLinkFormat.replace(/:product/gi, this.productLabel);
   }
 
-  async fetchProduct(productLabel?: string, productId?: string) {
+  async fetchProduct({ id, label }: { id?: string; label?: string }) {
```
dangodev: This was converted to named params so you could do `fetchProduct({ label: 'lodgna' })` rather than `fetchProduct(undefined, 'lognda')`

src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -57,6 +62,40 @@ export class ManifoldMarketplace {
...
+
+            if (Array.isArray(plansResp) && plansResp.find(plan => plan.body.free === true)) {
+              freeProducts.push(id);
+            }
+            resolve();
+          })
+      )
+    );
+
+    this.freeProducts = freeProducts; // don’t update state until all requests finish
```
davidleger95: should we request a filter for the graphql api so we can filter by cost?

just thinking for future optimization

dangodev: Yeah it’s something I’ve asked for (repeatedly, for a while now), that we may be getting from a DO request. Fingers crossed 🤞

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -15,44 +15,50 @@ export class ManifoldServiceCard {
...
+  @Watch('product') productChange(newProduct: Catalog.Product) {
+    if (newProduct) {
+      this.fetchIsFree(); // if product has changed, re-fetch free
+    }
+  }
+
   @Watch('skeleton') skeletonChange(newSkeleton: boolean) {
-    if (!newSkeleton) {
-      this.fetchProduct(this.productLabel, this.productId);
+    if (newSkeleton !== true) {
```
Minivera: Eslint complains about those unsimplified conditions, can we go back to the old version?

dangodev: Yeah that’s fine. I was being paranoid and in HTML didn’t want `newSkeleton = ''` to be interpreted as `true`, but now that I think about it, I think that’s expected behavior anyway. But in general I try to be more explicit about potential HTML coercion

dangodev: Updated!

dangodev:
@Minivera would you prefer this be merged before or after your PR? I’m fine with either—whichever is easier for you

Minivera:
@dangodev sorry I missed you comment, after seems like the best course of action :)

PR #325, opened 2019-07-31
Send productName to platforms through 'manifold-marketplace-click' event emitter

batgirl:
## Reason for change
This PR passes the product name through the event emitter in `manifold-marketplace-click` so it can be handled by platforms.

In the Render dashboard, this enables the "Choose" button on the New Addon modal to show the product name instead of showing the product label. Related to https://github.com/manifoldco/dashboard-render/pull/19

## Testing
Load the Render Dashboard > Go to `Addons` > Click `New Addon` and then select a product > View the modal with product information and see the product name on the "Choose" button.

Screen Shot 2019-07-31 at 12 55 14 PM


src/components.d.ts
```diff
@@ -199,9 +199,9 @@ export namespace Components {
     */
     'productLabel'?: string;
     /**
-    * Region to provision (complete name), omit for all region
+    * Region to provision (ID)
     */
-    'regionName'?: string;
+    'regionId'?: string;
```
batgirl: I don't understand how this was picked up by `components.d.ts` - doesn't seem related to the changes I made. Should I put it back how it was?

sslotsky: Running the `dev` task causes these to be generated, so I think a previous PR probably changed this component definition without the `dev` task running. As a result, more than one of us has accidentally introduced this change because our dev servers _were_ running. 

Maybe just make sure your branch is up to date, kill the dev task if it's already running, and then restart it to make sure?

batgirl: So `npm run build` does not build the generated components, `npm run dev` does?

Also: you're saying I _do_ want `dev` to be running while I push the branch? It definitely wasn't before.

Minivera: `npm run dev` runs `npm run build` 😄

batgirl: Ok (for transparency to make sure I understood correctly) I rebased master from my branch and nothing changed, ran `npm run dev` and nothing changed, and I still see `regionId` and no `regionName` in `component.d.ts` and have nothing new to commit.

batgirl:
Yes sorry - docs and changelog?

Minivera:
Both would be great! Thanks

src/components.d.ts
Minivera: It's very possible a merge got in without the author running `npm run build` beforehand. Stencil knows what's it doing so we can ignore that I think.

davidleger95: That was my mistake, sorry. It seems I forgot to push the updated component types before merging my PR yesterday

batgirl:
@dangodev added the docs - and tried to add some coverage, but it wasn't enough :( I'm not sure how else to test the click event beyond what I added. Is there anything obvious I'm missing?

docs/docs/components/manifold-service-card.md
```diff
@@ -75,13 +75,13 @@ when it updates. To listen to those events, add an event listener either on
 the component itself, or `document`:
 
 ```js
-document.addEventListener('manifold-marketplace-click', ({ detail: { productLabel } }) => {
-  alert(`You clicked the card for ${productLabel}`);
```
dangodev: let’s maybe keep `productLabel` (sorry) because it is useful for URLs, etc. Maybe add both?

dangodev:
> tried to add some coverage, but it wasn't enough

coverage schmoverage this is good enough

docs/docs/components/manifold-service-card.md
batgirl: Is there an example of a use case for a `product label` I can add? I didn't want to run into the same problem where someone thinks `productLabel` is going to be a good name to display now that we have `productName`.

dangodev: product label is good for navigation: 
```
document.addEventListener('manifold-marketplace-click', ({ detail: { productLabel } }) => {
  window.location = `/product/${productLabel}`;
}
```

product name is good for displaying, like you pointed out.

Maybe just rework the example to use all 3? e.g.:

```js
document.addEventListener('manifold-marketplace-click', ({ detail } }) => {
  console.log(detail); // { productLabel: "jawsdb-mysql", productName: "JawsDB MySQL", productId: "234w1jyaum5j0aqe3g3bmbqjgf20p" } 
```

dangodev: Or something else that you feel is more readable (I find it helpful to run `npm run docs` and write the docs while looking at them, rather than trying to write docs in your IDE)

batgirl: Oh nice! I hadn't played with that. Thanks, I like the example that shows the detail in more.. detail :)

PR #326, opened 2019-07-31
Improve handling of token and expiry

sslotsky:
## Reason for change

1. The expiry should be delimited by a value that doesn't appear in JWT tokens. We chose `|`.
2. We should not use our local storage as backup. Doing so causes us to set the value in local storage even when a platform decides to control it themselves. This creates a false reassurance that the security of the token is in their hands.

### DIFF
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -22,3 +22,20 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
     },
   };
 }
+
+export function isExpired(token: string) {
+  try {
+    const [, expiry] = token.split('|');
+    const numericExpiry = parseInt(expiry, 10);
+    if (Number.isNaN(numericExpiry)) {
+      // We have no expiry information, or it's badly formatted. Consider this expired.
+      return true;
+    }
+
+    const d = new Date(numericExpiry * 1000);
+    return d < new Date();
+  } catch {
+    // Not a valid unix timestamp. Consider this expired.
+    return true;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/utils/auth.ts
```diff
@@ -22,3 +22,20 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
     },
   };
 }
+
+export function isExpired(token: string) {
+  try {
+    const [, expiry] = token.split('|');
```
domenicrosati: nice

src/utils/auth.ts
```diff
@@ -22,3 +22,20 @@ export function withAuth(authToken?: string, options?: RequestInit): RequestInit
...
+    const numericExpiry = parseInt(expiry, 10);
+    if (Number.isNaN(numericExpiry)) {
+      // We have no expiry information, or it's badly formatted. Consider this expired.
+      return true;
+    }
+
+    const d = new Date(numericExpiry * 1000);
+    return d < new Date();
```
domenicrosati: yep!

sslotsky: Wait a second is this backwards? Lol I think it might be backwards.

sslotsky: Wait no. It's not. I am. LOL

domenicrosati: no your good! i read it as the expiry is less than the current time so its fine

dangodev: Not a big deal, but maybe a better name than `d` would help?

sslotsky: Yeah, it's just logic confusion. I tend to expect expect functions that return `true` to be _passing_ flags and functions that return `false` to be _failing_ flags. But if the question you're asking is `isExpired` then a `false` indicates that the token is usable, which in a way feels like a passing flag.

domenicrosati:
I suppose you could say
Demo'd to @domenicrosati

PR #328, opened 2019-08-01
Add better release documentation

dangodev:
## Reason for change
Screen Shot 2019-08-01 at 09 48 18

## Testing
See the **🚀Deploying** section in [src/readme.md](https://github.com/manifoldco/ui/blob/dangodev/release-docs/src/readme.md#-deploying)


src/readme.md
```diff
@@ -149,40 +149,117 @@ hot reload, however it will break if you don’t add a `path` in
...
+
+💁 Remember: **don’t use `-latest` as a tag!** That’s the only reserved tag on npm.
+
+#### Releasing locally
+
+For faster publishing, or to avoid Travis, you can choose to publish from
+your computer by running:
+
+```bash
+npm run publish
```
markjamieson: This doesn't quite feel right to me but I'm having a hard time articulating why. I'm aware that you're not advocating skipping Travis on anything but pre-releases but I'd be curious if anyone else has concerns.

dangodev: Yeah. This is just extra info; I’m OK with removing it from the documentation and this just being a secret option in case of emergencies. Maybe just remove it, and if you know you know (you’d probably figure it out by reading the Travis task anyway)?

src/readme.md
```diff
@@ -149,40 +149,117 @@ hot reload, however it will break if you don’t add a `path` in
...
-requests the flag (e.g.: `npm i --save @manifoldco/ui@rc` or `npm i --save @manifoldco/ui@alpha`).
+<p align="center">
+  <img src="../.github/semver.svg" alt="npm semver" width="600" height="235" />
+</p>
 
-### Stable release
+| Increment | Description                                                                                                                                                                                                             |
+| :-------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Patch** | There are **no breaking changes** in this release; this is just a bugfix or minor quality improvement that can safely be pushed to every platform.                                                                      |
+| **Minor** | There are **some breaking changes** in this release that are well-documented in the [CHANGELOG][changelog]. Large bugs, or bigger improvements are shipped in this version, but the overall structure remains the same. |
```
sslotsky: I don't think there should be any breaking changes in a minor release. It should be new features only. But this is only true once you've reached a version 1.0.0. Before that time, it's considered to be unstable and breaking changes can occur at essentially any time. This is my understanding anyway.

src/readme.md
sslotsky: If we decide to keep it, it might be worth clarifying that you have to create the new tag before `npm run publish` can work. Using `npm version <version-number-without-the-v>` is how I would normally do this, but I don't feel like we've ironed out the details of how this should work. E.g. this will commit a change to the package.json and create a new tag and I don't think it's clear whether or not we should push those.

src/readme.md
```diff
@@ -149,40 +149,117 @@ hot reload, however it will break if you don’t add a `path` in
...
+
+Our [CHANGELOG][changelog] is important to keep up-to-date for dependabot,
+and other CI tools.
+
+**Only once the CHANGELOG is updated may you merge your PR.**
+
+### Step 4: final release
+
+1. Decide what the version will be (see [anatomy of a semver tag][anatomy] above).
+1. Pretend we’ve decided `v1.2.3` will be the new version.
```
sslotsky: The pretending part isn't really a step in the process. Maybe it would read better like this?

### Step 4: final release

1. Decide what the version will be (see [anatomy of a semver tag][anatomy] above). The following steps assume that you've chosen `v1.2.3` for your new version.
1. Publish `v1.2.3-rc.0` to npm, and **test it fully**

src/readme.md
```diff
@@ -149,40 +149,117 @@ hot reload, however it will break if you don’t add a `path` in
...
 
-To publish to npm, tag it in Git with a valid [npm-semver][npm-semver].
+Not every PR will warrant a release—it’s fine to create & merge PRs into
+`master` with the intent that multiple will comprise a release later. But for
+PRs that will result in a release, follow the following steps:
 
-### Unstable release
+<ol>
+  <li>
+    Tag a <strong>prerelease</strong> tag in GitHub with a valid <strong>semver</strong> like so:
```
jessifoo: Would it help to point out how to create a pre-release, by going to the main page of the repo, clicking on releases, and then draft a new release, etc.? Or is this specified somewhere else already?

dangodev: Ah good question. I won’t explain it, but I can link the page to releases in GitHub. Hopefully readers will connect 2 & 2

src/readme.md
dangodev: > I don't think there should be any breaking changes in a minor release. It should be new features only.

So you’d say **patch** is bugs-only; **minor** is features-only? Neither are breaking?

I hesitate to only have major be breaking changes, because say we change a CSS variable. That is a breaking change, but is that enough to bump from `1.0.0` to `2.0.0`?

sslotsky: We're not at version 1 yet, so a breaking change can happen at any time. But it's widely understood that breaking changes should only happen for major releases. Let's make sure we have our CSS variables the way we want them before we call the project stable :smiley_cat:

dangodev: semver.org says you’re right! 😄 Will update. Thanks for the feedback!

src/readme.md
```diff
@@ -149,40 +149,106 @@ hot reload, however it will break if you don’t add a `path` in
...
+- [ ] Are the [docs][docs] up-to-date?
+- [ ] Do any [Storybook][storybook] stories need to be written?
+- [ ] Do any [Happo tests][happo] need to be written?
 
+If everything looks good and works as expected, **don’t merge your PR yet**!
+Update the CHANGELOG (below).
+
+### Step 3. CHANGELOG
+
+Our [CHANGELOG][changelog] is important to keep up-to-date for dependabot,
```
Minivera: Maybe also link to the docs changelog?

dangodev: Oh hm. I didn’t notice that was duplicated. Let’s set up a script to copy that over, rather than relying on people to do that manually.

Minivera: There are some differences between the two, mostly in the frontmatter

PR #329, opened 2019-08-01
Fix docs and storybook

Minivera:
Work is related to manifoldco/engineering#8893

## Reason for change
The docs will now use a mocked fetch function that will catch all the content we want to be fixed and not require an auth-token and let through other calls. It uses a special plugin that preloads all our catalog data on build and will make sure our docs are never down due to manifold being down.

Storybook has be changed back the old story where you had to set the `manifold_api_token` manually in localstorage. Make sure to always have a production login token from dashboard in you localstorage.

## Testing
Test using the docs and storybook, both should work as expected.

For storybook, you will need to set a this localStorage value `manifold_api_token` to a prod dashboard login token.


dangodev:
Screen Shot 2019-08-01 at 11 56 33

Hm. This still isn’t working for me (at least on the preview link)

dangodev:
Likewise, docs is still not working (`manifold-plan-selector`):

Screen Shot 2019-08-01 at 11 58 26

dangodev:
This is the behavior I’m seeing (data not loading on hard refresh, plans being marked as “free”):

dangodev:
> This is the behavior I’m seeing (data not loading on hard refresh, plans being marked as “free”):
> 
>

dangodev:
Let’s get the fix in for the other components, then I’m happy to ✅ this PR and we can :shipit:

dangodev:
One last fix (for tomorrow): on the **JawsDB Custom Plan** plan, the cost API is no longer firing when you update the values. It just stays at $33 no matter what the value.

This may be an issue from my PR when I converted it to `restFetch`, but maybe something to look into.

dangodev:
Also ignore the CircleCI failure for now. I’m testing using that for deployments in another branch, and that will fail till it merges. I’m happy to adult it through if necessary.

stories/connectionDecorator.js
```diff
@@ -1,6 +1,8 @@
 export const manifoldConnectionDecorator = storyFn => `
   <manifold-connection>
-    <manifold-auth-token/>
+    <manifold-auth-token token="${localStorage.manifold_api_token}|${new Date(
+  Date.now() + 6.04e8
+).getTime()}"/>
```
dangodev: I was about to ask what this is doing but I see this is for Storybook. WFM but maybe a comment here would be helpful?

src/utils/restFetch.ts
```diff
@@ -26,7 +26,7 @@ export const createRestFetch = ({
   const url = `${endpoints[args.service]}${args.endpoint}`;
   const start = new Date();
 
-  while (!getAuthToken() && start.getTime() - new Date().getTime() <= wait) {
+  while (!getAuthToken() && new Date().getTime() - start.getTime() <= wait) {
```
dangodev: Why the change here, and in `graphqlFetch`?

Minivera: The condition was backwards so it actually led to an infinite loop

dangodev: That makes sense, but could this be a little more readable? i.e. could this be abstracted into a separate function here and in `graphqlFetch`? `tokenStillValid()` or something? That way a) this logic has a name for what it’s doing, and b) we have a test for this

src/state/connection.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class ConnectionState {
+  /** _(optional)_ Specify `env="stage"` for staging */
+  env: 'local' | 'stage' | 'prod' = 'prod';
+  /** _(optional)_ Wait time for the fetch calls before it times out */
+  waitTime: number = baseWait;
+
+  authToken?: string;
+
+  setEnv = (newValue: 'local' | 'stage' | 'prod') => {
```
dangodev: Minor point: this could be abstracted into a `type env = 'local' | 'stage' | 'prod'` within this file so it can be shared with `ConnectionState` as well.

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -15,45 +12,23 @@ export class ManiTunnel {
...
-            setAuthToken: this.setAuthToken,
-            endpoint: connections[this.env].graphql,
-            wait: this.waitTime,
-          }),
-        }}
-      >
-        <slot />
-      </Tunnel.Provider>
-    );
+    return <slot />;
```
dangodev: 💯

src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -15,45 +12,23 @@ export class ManiTunnel {
...
-  get accessToken() {
-    if (this.authToken) {
-      const [token] = this.authToken.split('|');
-      return token;
+  componentWillLoad() {
+    if (this.setEnv) {
+      this.setEnv(this.env);
+    }
+    if (this.setWaitTime) {
+      this.setWaitTime(this.waitTime);
```
dangodev: I know you’ve written a ton of tests already, but is there a good test we can write here, to ensure this sticks around? This feels critical to everything working, and feels a little brittle (or untouchable).

docs/src/pages/index.tsx
```diff
@@ -63,6 +89,208 @@ export const query = graphql`
...
+            body {
+              location
+              name
+              platform
+              priority
+            }
+          }
+        }
+      }
+    }
```
dangodev: 👍

docs/src/components/Example.tsx
```diff
@@ -39,7 +39,18 @@ interface ExampleProps {
...
+      dangerouslySetInnerHTML={{
+        __html: `
+          <manifold-connection>
+            <manifold-auth-token token="1234|${new Date(
+              Date.now() + 6.04e8
+            ).getTime()}" oauth-url="/oauth"></manifold-auth-token>
+            ${html}
+          </manifold-connection>`,
+      }}
+    />
```
dangodev: 👍

docs/plugins/gatsby-source-mayflower/.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+      name: 'Run Linters'
+      script: npm run lint
+    - stage: test
+      name: 'Run tests'
+      script: npm run test
+cache: npm
+branches:
+  only:
+    - master
+    - "/^v([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:\\-(rc\\.[0-9]+)*)?$/"
```
dangodev: Does this need to be in here?

dangodev:
We’ll have to come up with a solution for Storybook now, but at least the docs are back! Thank you.

PR #334, opened 2019-08-02
Handle responses that don't have content

sslotsky:
## Reason for change

The deprovision request was causing an error in UI because it sends a 202 with no content in the response, but the client code is assuming that there will be a JSON body. 

This change ensures that a 202 response with no body will not get all explody.

## Testing

Two ways to test.

1. Pull down the branch and comment out the code that was added to `restFetch.ts`, then run the test for `manifold-data-deprovision-button`. The test should fail. Then uncomment the code and the test should pass.
2. Try this out in Render Dashboard and make sure you can deprovision.

### DIFF
--- a/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
+++ b/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
@@ -130,8 +130,7 @@ describe('<manifold-data-deprovision-button>', () => {
 
     it('will trigger a dom event on successful deprovision', async () => {
       fetchMock.mock(`${connections.prod.gateway}/id/resource/${Resource.id}`, {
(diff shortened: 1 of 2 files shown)


src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -130,8 +130,7 @@ describe('<manifold-data-deprovision-button>', () => {
 
     it('will trigger a dom event on successful deprovision', async () => {
       fetchMock.mock(`${connections.prod.gateway}/id/resource/${Resource.id}`, {
-        status: 200,
-        body: Resource,
+        status: 202,
```
Minivera: 👍

PR #335, opened 2019-08-02
Show how tests can be improved significantly by using the DOM

Minivera:
Stencil actually gives you the full dom API when running tests with `newSpecPage`, this means that we can run tests against dom nodes and events rather than the web component instance.

This shows how that can be used with two components, one using events (Sadly a quirk in stencil makes those tests a bit annoying) and one using HTML comparison.


src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
```diff
@@ -145,82 +120,96 @@ describe('<manifold-data-sso-button>', () => {
...
+
+      element.addEventListener('manifold-ssoButton-click', onClick);
+      element.addEventListener('manifold-ssoButton-success', onSuccess);
+      expect(element.querySelector('button')).toBeDefined();
+      // @ts-ignore
+      element.querySelector('button').click();
+      await page.waitForChanges();
+      // TODO: wait for changes does not wait for all events to finish emitting.... WHYYYYYYY
+      // TODO: Wait on this: https://github.com/ionic-team/stencil/issues/1781
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
Minivera: Had to create a bug report for this: https://github.com/ionic-team/stencil/issues/1781

If you don't await anything in a function that emits events, those events are not awaited and the jest mocks are saying they haven't run. To see why this is so infuriating, try changing line 125 to `const onSuccess = jest.fn(() => console.log('Hello from success'));`

sslotsky: This is very sad :disappointed:

src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
```diff
@@ -145,82 +120,96 @@ describe('<manifold-data-sso-button>', () => {
...
+      const root = page.root as HTMLElement;
+      element.resourceLabel = Resource.body.label;
+      root.appendChild(element);
+      await page.waitForChanges();
+
+      element.addEventListener('manifold-ssoButton-click', onClick);
+      element.addEventListener('manifold-ssoButton-success', onSuccess);
+      expect(element.querySelector('button')).toBeDefined();
+      // @ts-ignore
+      element.querySelector('button').click();
```
sslotsky: I'm having mixed feelings about using `@ts-ignore` for this. I understand that the alternative is lots of nesting. Would be nice if the `expect` call on line 135 told TS what it needs to know.

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -59,10 +59,10 @@ export class ManifoldServiceCard {
   }
 
   get href() {
-    if (!this.productLinkFormat || !this.productLabel) {
+    if (!this.productLinkFormat || !this.product) {
```
sslotsky: Are the changes in this feel needed to improve the tests?

src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
Minivera: Yeah 😞

src/components/manifold-service-card/manifold-service-card.tsx
Minivera: No, those are actually bug fixes shown by the improved tests!

src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
sslotsky: I thought about this for a while, and I think it's safer to do the nesting. If somebody deletes your assertion on line 135, you no longer have a guarantee that the button is there. The compiler is there to protect us from things like these, I think we should let it do its job!

dangodev:
@Minivera sorry for this but you may have to rebase off latest `master` (`git fetch && git rebase origin/master && git push --force --force-with-lease`)

PR #337, opened 2019-08-05
Rework how the resources list work with operations

Minivera:
Work is related to manifoldco/engineering#8832

## Reason for change
We now use the code from the ZEIT hackahton, slightly modified to fit our needs. This is still FAR from perfect, but the list will now properly show provisioning and resizing operations if they exist. It will also show the data we want when deprovisoning.

The optimal next step right now would be to implement a GraphQl endpoint and rework components that use the resources rest endpoints to use GraphQl and not worry about operations anymore.

## Testing
Try testing with the render dashboard, it should show the proper behavior.

### DIFF
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
@@ -14,8 +14,17 @@ interface EventDetail {
 
 const AVAILABLE = 'available';
 const PROVISIONING = 'provision';
-const DEGRADED = 'degraded';
+const RESIZING = 'resize';
+const DEPROVISION = 'deprovision';
 const OFFLINE = 'offline';
+
+const statusToText = {
+  [AVAILABLE]: 'Available',
+  [PROVISIONING]: 'Provisioning',
+  [RESIZING]: 'Resizing',
+  [DEPROVISION]: 'Deprovisioning',
+  [OFFLINE]: 'Offline',
+};
 
 @Component({
   tag: 'manifold-resource-card-view',
@@ -56,16 +65,32 @@ export class ManifoldResourceCardView {
     return this.resourceLinkFormat.replace(/:resource/gi, this.label);
   }
 
-  get status(): string {
+  get status() {
     if (
       this.resourceStatus &&
       (this.resourceStatus === AVAILABLE ||
         this.resourceStatus === PROVISIONING ||
-        this.resourceStatus === DEGRADED)
+        this.resourceStatus === RESIZING ||
+        this.resourceStatus === DEPROVISION)
     ) {
       return this.resourceStatus;
     }
     return OFFLINE;
+  }
+
+  get statusText() {
+    switch (this.resourceStatus) {
+      case AVAILABLE:
+        return statusToText[AVAILABLE];
+      case PROVISIONING:
+        return statusToText[PROVISIONING];
+      case RESIZING:
+        return statusToText[RESIZING];
+      case DEPROVISION:
+        return statusToText[DEPROVISION];
+      default:
+        return statusToText[OFFLINE];
+    }
   }
 
   async fetchResourceId(resourceLabel: string) {
@@ -122,7 +147,7 @@ export class ManifoldResourceCardView {
         <div class="status-box">
           <div class="status" data-status={this.status}>
             <div class="inner" itemprop="status">
-              {this.status[0].toUpperCase() + this.status.slice(1)}
+              {this.statusText}
             </div>
           </div>
         </div>
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
@@ -68,10 +68,7 @@ export class ManifoldResourceCardView {
   get status() {
     if (
       this.resourceStatus &&
-      (this.resourceStatus === AVAILABLE ||
-        this.resourceStatus === PROVISIONING ||
-        this.resourceStatus === RESIZING ||
-        this.resourceStatus === DEPROVISION)
+      [AVAILABLE, PROVISIONING, RESIZING, DEPROVISION].includes(this.resourceStatus)
     ) {
       return this.resourceStatus;
     }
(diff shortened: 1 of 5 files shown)


src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -118,51 +119,111 @@ export class ManifoldResourceList {
...
+          const resource = resources.find((res: RealResource) => opBody.resource_id === res.id);
+
+          if (resource) {
+            const product = productsResp.find(
+              (prod: Catalog.Product): boolean => prod.id === resource.body.product_id
+            );
+
+            // Ignoring because we set `this.resource` to an empty array above
+            // @ts-ignore
+            this.resources.push({
```
sslotsky: Perhaps a better option here would be to initialize a local `resources` variable to an empty array, and set `this.resources = resources` when you're all done.

sslotsky: I see you already have a variable named `resources` but with a bit of renaming my comment still stands :sweat_smile:

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
```diff
@@ -56,18 +65,34 @@ export class ManifoldResourceCardView {
...
 
-  get status(): string {
+  get status() {
     if (
       this.resourceStatus &&
       (this.resourceStatus === AVAILABLE ||
         this.resourceStatus === PROVISIONING ||
-        this.resourceStatus === DEGRADED)
+        this.resourceStatus === RESIZING ||
+        this.resourceStatus === DEPROVISION)
```
dangodev: Change this to `[AVAILABLE, PROVISIONING, RESIZING, DEPROVISION].includes(this.resourceStatus)` instead. Easier to read; harder to mess up

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
```diff
@@ -56,18 +65,34 @@ export class ManifoldResourceCardView {
...
+      case PROVISIONING:
+        return statusToText[PROVISIONING];
+      case RESIZING:
+        return statusToText[RESIZING];
+      case DEPROVISION:
+        return statusToText[DEPROVISION];
+      default:
+        return statusToText[OFFLINE];
+    }
+  }
```
dangodev: Why do we need a function for this? Why can’t we use `statusToText[status] || statusToText.offline` in places?

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -118,51 +119,111 @@ export class ManifoldResourceList {
...
+              id: resource.id,
+              label: resource.body.label,
+              name: resource.body.name,
+              logo: product && product.body.logo_url,
+              status: ManifoldResourceList.opStateToStatus(operation),
+            });
+            return;
+          }
+          if (operation.body.type !== 'provision') {
```
dangodev: Maybe this should be a constant somewhere?

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -118,51 +119,111 @@ export class ManifoldResourceList {
...
+        this.resources.push({
+          id: resource.id,
+          label: resource.body.label,
+          name: resource.body.name,
+          logo: product && product.body.logo_url,
+          status: 'available',
+        });
+      });
+
+      this.resources.sort((a, b) => a.label.localeCompare(b.label));
```
dangodev: Be careful! `.sort()` will mutate the original reference, and Stencil may not pick up on updates. Perhaps `this.resources = [...this.resources].sort(…` is what you want instead?

src/components/manifold-resource-list/manifold-resource-list.tsx
dangodev: +1 to Sam’s comment. When you’re mutating Stencil’s state like this, you’re telling it you want to update, but it’s not complete yet. Let’s get everything ready, and only update Stencil state at the very end.

Minivera: Ohhhhhhh, good point

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -118,51 +119,111 @@ export class ManifoldResourceList {
...
+              break;
+            case 'resize':
+              opBody = operation.body as Provisioning.resize;
+              break;
+            case 'transfer':
+              opBody = operation.body as Provisioning.transfer;
+              break;
+            case 'deprovision':
+              opBody = operation.body as Provisioning.deprovision;
+              break;
```
dangodev: I think could be simplified by skipping the assignment if `operation.body.type` isn’t one of the specified types, no?

I’m leery of using `switch` just for TypeScript types—there might be a better way.

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
Minivera: Typescript does not necessarily know that `status` is a key of `statusToText`

dangodev: Right, but that’s because at runtime we can’t be sure of that either. Previously, we just let the API decide what the status was, and then passed it through. Why are we changing that, so that now we have to maintain our own library of statuses in addition to the API?

batgirl:
Testing this in Render Dash, it makes a huge difference in the way a provisioning resource loads! No more waiting with nothing happening and I don't notice any flickers.

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
dangodev: I’d prefer it if you used only either the dictionary `statusToText` at top, or just returned strings in the `switch`—either one.

If you’re concerned about keys of `statusToText`, then I believe removing the dynamic keys from the object (`[AVAILABLE]`) will enable TypeScript to treat it as an object with a defined set of keys.

Minivera: I want to use the constants as keys to remove the need for having multiple instances to change if those ever change. 

Also, our statuses were not the same as the statuses that gateway may return for resources, which is why I changed it. However, for some annoying reason, gateway returns the work `provision` as the status rather than `provisioning`. So we have to translate.

PR #338, opened 2019-08-05
 Add a timeout on the resource container to keep fetching until valid

Minivera:
Work is related to manifoldco/engineering#8832

## Reason for change
This will enable the container to fetch a resource that is still provisioning or is having an operation run on it. It will keep fetching until "done" to ensure that the container can handle cases like being rendered on a provisioning resource.

## Testing
Test with the render dashboard after provisioning a resource to see it in action.

### DIFF
--- a/docs/docs/advanced/resource-details.md
+++ b/docs/docs/advanced/resource-details.md
@@ -45,6 +45,9 @@ a `manifold-resource-container` component with access to the resource label.
   <manifold-resource-credentials></manifold-resource-credentials>
 </resource-mock-resource>
 
+## Refetching until the resource is valid
+The `refetch-until-valid` property on the `resource-container` can be used to force the container to refetch the resource until it has found one and the status of that resource is available. This is useful for reloading the page until the resource has finished resizing or provisioning for example.
+
 ## The resource credentials
 
 The [`manifold-credentials`](/components/credentials) component can be used inside the container without any attribute by
--- a/src/components/manifold-resource-container/manifold-resource-container.spec.ts
+++ b/src/components/manifold-resource-container/manifold-resource-container.spec.ts
@@ -79,7 +79,7 @@ describe('<manifold-resource-credentials>', () => {
   it('will refetch the resource after load if given an invalid resource', async () => {
     const resourceLabel = 'test-resource';
     // @ts-ignore
-    window.setTimeout = jest.fn(call => call());
+    page.win.setTimeout = jest.fn(call => call());
 
     fetchMock
       .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, {})
@@ -100,7 +100,7 @@ describe('<manifold-resource-credentials>', () => {
   it('will refetch the resource after load if it received an error', async () => {
     const resourceLabel = 'test-resource';
     // @ts-ignore
-    window.setTimeout = jest.fn(call => call());
+    page.win.setTimeout = jest.fn(call => call());
 
     fetchMock
       .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, 404)
--- a/src/components/manifold-resource-container/manifold-resource-container.spec.ts
+++ b/src/components/manifold-resource-container/manifold-resource-container.spec.ts
@@ -94,6 +94,27 @@ describe('<manifold-resource-credentials>', () => {
     expect(fetchMock.called(`${connections.prod.gateway}/resources/me/${resourceLabel}`)).toBe(
       true
     );
-    expect(window.setTimeout).toHaveBeenCalled();
+    expect(window.setTimeout).toHaveBeenCalledTimes(1);
+  });
+
+  it('will refetch the resource after load if it received an error', async () => {
+    const resourceLabel = 'test-resource';
+    // @ts-ignore
+    window.setTimeout = jest.fn(call => call());
+
+    fetchMock
+      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, 404)
+      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, GatewayResource);
+
+    const root = page.root as HTMLElement;
+    element.resourceLabel = resourceLabel;
+    element.refetchUntilValid = true;
+    root.appendChild(element);
+    await page.waitForChanges();
+
+    expect(fetchMock.called(`${connections.prod.gateway}/resources/me/${resourceLabel}`)).toBe(
+      true
+    );
+    expect(window.setTimeout).toHaveBeenCalledTimes(1);
   });
 });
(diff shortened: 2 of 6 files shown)


docs/docs/advanced/resource-details.md
```diff
@@ -45,6 +45,9 @@ a `manifold-resource-container` component with access to the resource label.
   <manifold-resource-credentials></manifold-resource-credentials>
 </resource-mock-resource>
 
+## Refetching until the resource is valid
+The `refetch-until-valid` property on the `resource-container` can be used to force the container to refetch the resource until it has found one and the status of that resource is available. This is useful for reloading the page until the resource has finished resizing or provisioning for example.
```
sslotsky: Thank you for making this a single line! :smile:

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -12,21 +12,33 @@ export class ManifoldResourceContainer {
...
+  componentWillLoad() {
+    return this.fetchResource(this.resourceLabel);
+  }
+
+  componentDidUnload() {
+    clearTimeout(this.timeout);
+  }
+
+  fetchResource = async (resourceLabel?: string) => {
+    if (!this.restFetch || !this.resourceLabel) {
```
sslotsky: Now I find myself confused as to why `resourceLabel` is optional. I know it's been that way for a while, but I don't understand why.

src/components/manifold-resource-container/manifold-resource-container.spec.ts
```diff
@@ -20,4 +38,62 @@ describe('<manifold-resource-credentials>', () => {
...
+    );
+  });
+
+  it('will refetch the resource after load if given an invalid resource', async () => {
+    const resourceLabel = 'test-resource';
+    // @ts-ignore
+    window.setTimeout = jest.fn(call => call());
+
+    fetchMock
+      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, {})
```
sslotsky: If there's no resource matching the label, shouldn't the API give us a 404 or something?

Minivera: It does yeah, this line is simply to allow two calls to the API, one where an invalid resource (One where the state is not `available`) is returned so it calls the setTimeout, then one that is valid so it won't have to call it again.

src/components/manifold-resource-container/manifold-resource-container.tsx
Minivera: stencil needs the props to either be optional or have a default value. If we were to give it a default value, then it would try to fetch something with that default value, which would require this check either way. 🤷‍♂

sslotsky: > stencil needs the props to either be optional or have a default value.

I think that's just an eslint rule that we made, and it's not true for strings. It should only fail for props that can't be assigned via HTML attributes.

sslotsky: https://github.com/manifoldco/eslint-plugin-stencil#restrict-required-props

https://github.com/manifoldco/eslint-plugin-stencil/blob/master/index.js#L62

src/components/manifold-resource-container/manifold-resource-container.spec.ts
sslotsky: Ah, gotcha. Does it make sense to have a check where we get a 404 first? Presumably we have the resource label while provision is still in progress right?

Minivera: Sure would, that would cover the `if (response is Error)` case too

src/components/manifold-resource-container/manifold-resource-container.tsx
Minivera: Huh... I did not know.

I would still leave it as optional though, I think it makes it clearer that assign it a default value.

sslotsky: You don't need a default value in this case. I think the prop can be required, with no default value.

sslotsky: The rule in question does not apply to strings and boolean. They can be required without specifying defaults.

All other types are not legal html attributes. That's why they must either be optional or have defaults. The reasoning is that you should be able to use our web components from HTML without a runtime error.

Minivera: What I mean is that, if the developer using this component does not specify a `resource-label`, the property would be initialized with an empty string (Or null, not sure) even if I specified it as required. They don't have typescript to tell them not to outside of stencil.

However, in that case, the compile wouldn't tell me that I should probably check the value of `this.resourceLabel` before starting to check. That's why I think I should leave it and this is probably why optional strings are everywhere in UI.

src/components/manifold-resource-container/manifold-resource-container.spec.ts
```diff
@@ -94,6 +94,27 @@ describe('<manifold-resource-credentials>', () => {
...
       true
     );
-    expect(window.setTimeout).toHaveBeenCalled();
+    expect(window.setTimeout).toHaveBeenCalledTimes(1);
+  });
+
+  it('will refetch the resource after load if it received an error', async () => {
+    const resourceLabel = 'test-resource';
+    // @ts-ignore
+    window.setTimeout = jest.fn(call => call());
```
sslotsky: I'm wondering if it's possible to use `page.win` for this instead, and if it would be an improvement at all.

Minivera: I can try, not sure if that would make the test any way better 🤔

src/components/manifold-resource-container/manifold-resource-container.tsx
sslotsky: > They don't have typescript to tell them not to outside of stencil.

Hmm. The component definitions are supposed to expose this information, so if they're using TypeScript at all, it should require them to add the property. But admittedly we've had difficulties getting the types right with `JSX.IntrinsicElements`. And if they're not using TypeScript, then indeed this is a strong possibility. However it does kind of follow by extension that every component prop needs to be considered optional, even for React components. That's sort of painful but perhaps it's reality.

Just to be clear, I'm not arguing we should change anything here. It just stood out to me so I wanted to hash it out. Thanks for going over this with me.

src/components/manifold-resource-container/manifold-resource-container.spec.ts
Minivera: It works 😄

sslotsky: I think I was mainly wondering if it would allow us to skip the `ts-ignore`. I'm seeing a lot of that recently and I'd prefer not to ignore TS. Sometimes it might be easier, in which case a solution that makes it just as easy to not ignore it is preferable, but I don't think that a little extra legwork is usually a good reason to ignore TS.

Minivera: `ts-ignore` here is not for window, but actually for `call => call()`. Call is typed as a `TimerHandle` which apparently cannot be called as a function when mocked like this.

sslotsky: OH. Fascinating!

PR #341, opened 2019-08-06
Plan Selector free plans filtering

dangodev:
## Reason for change
We need to add free plan filtering that satisfy the following requirements:

- [ ] No changes are required on the user end end (i.e. the list should be able to be updated without changing any attributes)
- [ ] It’s a temporary shim until it’s supported by the API

Screen Shot 2019-08-06 at 15 48 49

## Testing
Test the `<manifold-plan-selector>` component with the new `platform` attribute, and see if the plan list filters (free plans only—LogDNA, Mailgun, etc.).

_Note: until #349 merges, you’ll have to manually authenticate into Storybook to see the plan selector component:_

- Log into dashboard.manifold.co
- In your JS Console, run `localStorage.getItem('manifold_api_token')`. **Copy** that value (including the surrounding `""`, if there are any).
- Go back to the Storybook preview link from this PR. In that page, open a JS Console, and run `localStorage.setItem('manifold_api_token', MYTOKEN)`, pasting in the value from dashboard.manifold.co instead of `MYTOKEN` (add surrounding quotes if they were missing before)
- Refresh. You should see everything load now.

### DIFF
--- a/docs/docs/components/manifold-plan-selector.md
+++ b/docs/docs/components/manifold-plan-selector.md
@@ -12,7 +12,7 @@ example: |
 Display the plans for a product.
 
 ```html
-<manifold-plan-selector product-label="jawsdb-mysql" />
+<manifold-plan-selector product-label="jawsdb-mysql"></manifold-plan-selector>
 ```
 
 You can find the `:product` label for each at
@@ -53,6 +53,12 @@ The following events are emitted:
 | `manifold-planSelector-change` | Fires whenever a user makes a change.                                                                                      | `planID`, `planLabel`, `planName`, `productLabel`, `features` |
 | `manifold-planSelector-load`   | Identical to `-update` above, but this fires once on DOM mount to set the initial state (i.e. user hasn’t interacted yet). | `planID`, `planLabel`, `planName`, `productLabel`, `features` |
 
+## Free plans only
+
+```html
+<manifold-plan-selector free-plans></manifold-plan-selector>
+```
+
 ## Regions
 
 Most of our products are regionless, but some plans let users specify the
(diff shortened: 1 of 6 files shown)


davidleger95:
Should we hide the plan selector on the left if there's only a single plan to show?

src/utils/plan.ts
```diff
@@ -295,6 +298,16 @@ export function planSort(plans: Catalog.ExpandedPlan[]): Catalog.ExpandedPlan[]
...
+
+  // TODO: return only free plans
+  if (options && options.free === true) {
+    return plans.filter(
+      ({ body }) =>
+        body.free === true || body.defaultCost === 0 || (!body.defaultCost && body.cost === 0)
+    );
+  }
+
+  return sorted;
```
davidleger95: Not that we need to worry about it now, but would a customizable plan with a base cost of $0 be shown here? Should it be shown? if there are optional paid features, we wouldn't want to include those (due billing restrictions), correct?

dangodev: Yeah good question. As for JawsDB, if it was free, the `defaultCost` would be `0`. That’s the difference between `defaultCost` and `cost`, with `defaultCost` being the cheapest sum of all features, and `cost` being the base cost of the plan, with features not included (for JawsDB, `defaultCost` is $35 as it should be, and `cost` is $33 because it doesn’t include the extra $2 that storage costs)

dangodev:
> Should we hide the plan selector on the left if there's only a single plan to show?

I think not; this is a temporary change, and long-term consumers will almost always show multiple plans. So even if we hid it, we’d just un-hide it very soon

> are we hiding paid products altogether?

No; that’s a function of `manifold-marketplace`. We could add that functionality, but for now our consumers aren’t using that component so it’s fine. They’re managing the product list, so the only thing they don’t manage will be plan lists (nor should they!)

docs/docs/components/manifold-plan-selector.md
```diff
@@ -53,14 +53,22 @@ The following events are emitted:
...
 
 Most of our products are regionless, but some plans let users specify the
 region. In this case, you may optionally want to order the regions with the
 `regions` attribute:
 
 ```html
-<manifold-plan-selector regions="aws-eu-west-1,aws-eu-west-2,aws-us-east-1,aws-us-east-2" />
+<manifold-plan-selector
+  regions="aws-eu-west-1,aws-eu-west-2,aws-us-east-1,aws-us-east-2"
+></manifold-plan-selector>
```
dangodev: Docs in general should really have closing tags. React auto-closes for you, hence the confusion, but in our docs we should have closing tags, and yes it actually does matter in HTML

PR #346, opened 2019-08-07
Add the terms of service if provided to the product page component

Minivera:
Work is related to manifoldco/engineering#8916

## Reason for change
Some providers have terms of services that need to be displayed on the product details component for legal reasons. This adds those terms if provided.

Capture d’écran 2019-08-07 à 10 22 34

## Testing
Test with the docs or storybook. Aiven cassandra is a product with terms.

### DIFF
--- a/src/components/manifold-product-page/manifold-product-page.tsx
+++ b/src/components/manifold-product-page/manifold-product-page.tsx
@@ -1,5 +1,5 @@
 import { h, Component, Prop } from '@stencil/core';
-import { arrow_up_right, book, life_buoy } from '@manifoldco/icons';
+import { arrow_up_right, book, life_buoy, file_text } from '@manifoldco/icons';
 import { Catalog } from '../../types/catalog';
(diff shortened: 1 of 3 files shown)


src/components/manifold-product-page/manifold-product-page.tsx
```diff
@@ -81,6 +89,19 @@ export class ManifoldProductPage {
                       <manifold-icon class="external-link-icon" icon={arrow_up_right} margin-left />
                     </a>
                   </div>
+                  {terms.provided && (
```
davidleger95: So this is a boolean? seems strange. I'd expect if the terms were not provided `terms` would just be undefined. 🤷‍♂

no change requested, just stating my curiosity.

Minivera: I have no idea why this is the way it is, but if you don't have a terms URL, that boolean tells you.

PR #347, opened 2019-08-07
Update restFetch to report errors

sslotsky:
## Reason for change

Errors that occur in API calls need to be reported to platforms. This will be emitted as a custom event named `manifold-error`.

Getting this right involved tightening up the types for `createRestFetch`, which had consequences for every component that calls `restFetch` and every corresponding test.

## Testing

Make sure Storybook components that fetch data are still working. Components displaying catalog data should continue to work without an API token, and components that display resources or any other private data should still work when a token is set in `localStorage.manifold_api_token`

### DIFF
--- a/src/components/manifold-credentials/manifold-credentials.tsx
+++ b/src/components/manifold-credentials/manifold-credentials.tsx
@@ -51,7 +51,7 @@ export class ManifoldCredentials {
       endpoint: `/resources/?me&label=${resourceLabel}`,
     });
 
-    if (!Array.isArray(resources) || !resources.length) {
+    if (!resources || !resources.length) {
       console.error(`${resourceLabel} resource not found`);
       return;
     }
--- a/src/components/manifold-credentials/manifold-credentials.tsx
+++ b/src/components/manifold-credentials/manifold-credentials.tsx
@@ -37,11 +37,6 @@ export class ManifoldCredentials {
       endpoint: `/credentials/?resource_id=${this.resourceId}`,
     });
 
-    if (response instanceof Error) {
-      console.error(response);
-      return;
-    }
-
     this.credentials = response;
     this.loading = false;
   };
@@ -51,16 +46,10 @@ export class ManifoldCredentials {
       return;
     }
 
-    const response = await this.restFetch<Marketplace.Resource[]>({
+    const resources = await this.restFetch<Marketplace.Resource[]>({
       service: 'marketplace',
       endpoint: `/resources/?me&label=${resourceLabel}`,
     });
-
-    if (response instanceof Error) {
-      console.error(response);
-      return;
-    }
-    const resources: Marketplace.Resource[] = response;
 
     if (!Array.isArray(resources) || !resources.length) {
       console.error(`${resourceLabel} resource not found`);
--- a/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
+++ b/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
@@ -175,13 +175,15 @@ describe('<manifold-data-deprovision-button>', () => {
       const instance = page.rootInstance as ManifoldDataDeprovisionButton;
       instance.error.emit = jest.fn();
 
-      await instance.deprovision();
-
-      expect(fetchMock.called(`${connections.prod.gateway}/id/resource/${Resource.id}`)).toBe(true);
-      expect(instance.error.emit).toHaveBeenCalledWith({
-        message: 'ohnoes',
-        resourceLabel: 'test',
-        resourceId: Resource.id,
+      return instance.deprovision().catch(() => {
+        expect(fetchMock.called(`${connections.prod.gateway}/id/resource/${Resource.id}`)).toBe(
+          true
+        );
+        expect(instance.error.emit).toHaveBeenCalledWith({
+          message: 'ohnoes',
+          resourceLabel: 'test',
+          resourceId: Resource.id,
+        });
       });
     });
 
--- a/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
+++ b/src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
@@ -175,6 +175,7 @@ describe('<manifold-data-deprovision-button>', () => {
       const instance = page.rootInstance as ManifoldDataDeprovisionButton;
       instance.error.emit = jest.fn();
 
+      expect.assertions(2);
       return instance.deprovision().catch(() => {
         expect(fetchMock.called(`${connections.prod.gateway}/id/resource/${Resource.id}`)).toBe(
           true
--- a/src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
+++ b/src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
@@ -54,18 +54,12 @@ export class ManifoldDataManageButton {
       return;
     }
 
-    const response = await this.restFetch<Marketplace.Resource[]>({
+    const resources = await this.restFetch<Marketplace.Resource[]>({
       service: 'marketplace',
       endpoint: `/resources/?me&label=${resourceLabel}`,
     });
 
-    if (response instanceof Error) {
-      console.error(response);
-      return;
-    }
-    const resources: Marketplace.Resource[] = response;
-
-    if (!Array.isArray(resources) || !resources.length) {
+    if (!resources || !resources.length) {
       console.error(`${resourceLabel} product not found`);
       return;
     }
@@ -94,36 +88,38 @@ export class ManifoldDataManageButton {
       req.features = this.features;
     }
 
-    const response = await this.restFetch<Gateway.Resource>({
-      service: 'gateway',
-      endpoint: `/id/resource/${this.resourceId}`,
-      body: req,
-      options: {
-        method: 'PATCH',
-        headers: { 'Content-Type': 'application/json' },
-      },
-    });
+    try {
+      const response = await this.restFetch<Gateway.Resource>({
+        service: 'gateway',
+        endpoint: `/id/resource/${this.resourceId}`,
+        body: req,
+        options: {
+          method: 'PATCH',
+          headers: { 'Content-Type': 'application/json' },
+        },
+      });
 
-    if (response instanceof Error) {
+      if (response) {
+        const planName = response.plan && response.plan.name;
+        const success: SuccessMessage = {
+          message: `${this.resourceLabel} successfully updated to ${planName}`,
+          resourceLabel: this.resourceLabel || '',
+          resourceId: this.resourceId,
+          planName,
+          features: response.features,
+        };
+        this.success.emit(success);
+      }
+    } catch (e) {
       const error: ErrorMessage = {
         resourceLabel: this.resourceLabel || '',
         resourceId: this.resourceId,
-        message: response.message,
+        message: e.message,
         features: this.features,
       };
       this.error.emit(error);
-      return;
+      throw error;
     }
-
-    const planName = response.plan && response.plan.name;
-    const success: SuccessMessage = {
-      message: `${this.resourceLabel} successfully updated to ${planName}`,
-      resourceLabel: this.resourceLabel || '',
-      resourceId: this.resourceId,
-      planName,
-      features: response.features,
-    };
-    this.success.emit(success);
   }
 
   @logger()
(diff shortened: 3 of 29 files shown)


src/utils/restFetch.ts
```diff
@@ -16,47 +17,57 @@ interface RestFetchArguments {
...
 
     // Sometimes messages are an array, sometimes they aren’t. Different strokes!
     const message = Array.isArray(body) ? body[0].message : body.message;
-    return new Error(message);
-  } catch (e) {
-    return e;
-  }
-};
+    // This is an expected error, it doesn't need to be reported
+    throw new Error(message);
```
sslotsky: @Minivera (and @davidleger95 may also be interested) this is based on your description here:

> This function should be able to handle specific error results from our APIs and trigger a DOM event on any status that is not expected by UI.

So I think this means for example that we're expecting certain errors like `404` etc, and we don't need to report these ones because they should be handled somewhere (usually in the components that call them), right? 

And, I think that `fetch` throws for `5xx` errors (the truly exceptional ones) but doesn't throw for the `4xx` errors, so the `5xx` errors should be handled by the `.catch()` block that I chained off of `.fetch()` here https://github.com/manifoldco/ui/pull/347/files#diff-3a691cbad385fe5b116a49c20dfbc7c0R47

One change worth noting here is that it will throw errors now instead of returning them, but I don't actually think that will negatively impact the behavior of the components as they're written today. The ones that handle errors check if the response is an error and then log it out to `console.error`, and then return. Now they'll just get an exception, which will also appear in the console, and execution of the callback will stop just the same. On the other hand, calls that aren't currently checking for error responses will now be more visible if they fail, since some seem to be optimistically setting response data in the component, e.g.

```js
    const response = await this.restFetch<Marketplace.Credential[]>({
      service: 'marketplace',
      endpoint: `/credentials/?resource_id=${this.resourceId}`,
    });

    this.credentials = response;
    this.loading = false;
```

src/utils/errorReport.ts
```diff
@@ -0,0 +1,5 @@
+export function report<Detail>(detail: Detail) {
+  console.error(detail); // report error (Rollbar, Datadog, etc.)
+  const evt = new CustomEvent('manifold-error', { bubbles: true, detail });
```
Minivera: This should be documented in the docs so users reading them can find out how to handle them properly.

src/utils/restFetch.ts
Minivera: What I meant originally by this is that all errors should be properly reported and should never lead to a blank state. 

For errors we can expect (And even recover from), we should report them and then act to make sure the user has some feedback regarding that error. It can be something like showing a toast with a description of what happened so they won't stare at a loading component for an hour or even restarting the fetch if we think we can recover it.

For errors we can't expect, when I discussed it with design, the idea was to report it and show a "Something unexpected happened" message in a toast on the component that errored out. If we're going the route where the platform gets the message, they need to have a way to either listen for that event on the component itself or know from which component it was triggered if they listen from the document.

sslotsky: Reporting all errors is fine, I can make that change.

For expected errors, I think we can usually do better than a toast. I talked with Nick and we're going to want to do an audit of all expected errors from all components and design UI state for it.

For unexpected errors like `5xx`, when talking with Nick we didn't think it was that important for the error to show up near the component it originated from. It's an exceptional case and something like a toast in the upper right hand corner is normal for stuff like that. So we can do something like that (not as part of this PR though) and send the event which will let the platform do something like remove UI from the DOM entirely and display a generic error message that fits with their design principles. "Try reloading!" or something. But I agree the event needs to be documented.

As for telling platforms which component it ultimately came from, I see that as the ideal, but I also think that sending them the `response` which tells them what the request URL was is probably good enough. If we really do need to solve the problem of specifying which component it came from, we should look at it in a separate PR. Let's iterate on this because it's already getting big.

Minivera: My issue with this method for 500 errors is that the entire page probably should be nuked if one plan's cost turned out to return a 500. A platform is likely to want to handle those errors gracefully and knowing where that 500 error came from programmatically is, in my opinion, very important.

I do agree this can be done in another PR and we can work together on an issue for that, but I don't think that a toast in the right hand corner is a common response to this kind of situation. I'd agree with you if we were providing page wide widgets that can break down if an individual component has a failure, but our web components are very modular, uncoupled and some are even invisible to the user. 

What happens if a platform adds the `manifold-auth-token` to all their pages and, on a completely unrelated page, if 500 for some reason? Then suddenly the platform user sees a "Try reloading" toast on the side even though the 500 has no incidence on the page they're currently viewing.

sslotsky: > if one plan's cost turned out to return a 500

To me that seems very unlikely. If you get a 500 from a plan cost call, it points to a problem that we should expect to see repeated. Something is seriously wrong with the backend code. I think we can come up with lots of "what if" scenarios but we need to focus on problems that are likely first.

Minivera: From my experience with the backend, individual 500 happen as often if not more than a 500 that can kill an entire page.

sslotsky: That seems worth looking into :sweat_smile: It would seem to indicate some unsafe code where the possibility of failure might not be handled on the backend. It's a different conversation altogether but gathering some stats on that and examining the causes seems prudent.

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -51,17 +46,11 @@ export class ManifoldCredentials {
...
       endpoint: `/resources/?me&label=${resourceLabel}`,
     });
 
-    if (response instanceof Error) {
-      console.error(response);
-      return;
-    }
-    const resources: Marketplace.Resource[] = response;
-
     if (!Array.isArray(resources) || !resources.length) {
```
sslotsky: I can change this to 

```js
if (!resources || !resources.length) {
```

I did this elsewhere but I didn't get every location. Might come back for it.

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -37,11 +37,6 @@ export class ManifoldCredentials {
       endpoint: `/credentials/?resource_id=${this.resourceId}`,
     });
 
-    if (response instanceof Error) {
-      console.error(response);
```
dangodev: Should we keep these? These do report to Rollbar & New Relic

dangodev: Granted, I just started reading this PR so I may be surprised later

sslotsky: The `console.error` will happen during the `restFetch` call now when the error actually occurs.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -175,13 +175,15 @@ describe('<manifold-data-deprovision-button>', () => {
...
-        resourceId: Resource.id,
+      return instance.deprovision().catch(() => {
+        expect(fetchMock.called(`${connections.prod.gateway}/id/resource/${Resource.id}`)).toBe(
+          true
+        );
+        expect(instance.error.emit).toHaveBeenCalledWith({
+          message: 'ohnoes',
+          resourceLabel: 'test',
+          resourceId: Resource.id,
+        });
```
dangodev: Will this test pass now if `catch()` isn’t called?

src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
```diff
@@ -94,36 +88,38 @@ export class ManifoldDataManageButton {
...
+
+      if (response) {
+        const planName = response.plan && response.plan.name;
+        const success: SuccessMessage = {
+          message: `${this.resourceLabel} successfully updated to ${planName}`,
+          resourceLabel: this.resourceLabel || '',
+          resourceId: this.resourceId,
+          planName,
+          features: response.features,
+        };
```
dangodev: I don’t think we’re testing success output (or error output) in this component. Is that something we can add?

src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
```diff
@@ -326,12 +326,12 @@ describe('<manifold-data-provision-button>', () => {
...
-      expect(fetchMock.called(`${connections.prod.gateway}/resource/`)).toBe(true);
-      expect(instance.error.emit).toHaveBeenCalledWith({
-        message: 'ohnoes',
-        resourceLabel: 'test',
+      return instance.provision().catch(() => {
+        expect(fetchMock.called(`${connections.prod.gateway}/resource/`)).toBe(true);
+        expect(instance.error.emit).toHaveBeenCalledWith({
+          message: 'ohnoes',
+          resourceLabel: 'test',
+        });
```
dangodev: Same question here as the other test—will this pass if `catch()` isn’t called?

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
sslotsky: Nope! `instance.deprovision()` will throw an exception that isn't handled, so the `expect` calls wouldn't even occur if they weren't wrapped in a `catch` block.

src/components/manifold-data-manage-button/manifold-data-manage-button.tsx
sslotsky: Yeah, lots of these `if (response) {` blocks are uncovered. These are the tests I was suggesting I'd like to add in our Slack conversation.

src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
sslotsky: The test will fail without the `catch()`

src/utils/restFetch.spec.ts
```diff
@@ -53,16 +53,16 @@ describe('The fetcher created by createRestFetch', () => {
       body: {},
     });
 
-    await fetcher({
+    return fetcher({
       endpoint: '/test',
       service: 'marketplace',
+    }).catch(() => {
+      expect(fetchMock.called('path:/v1/test')).toBeTruthy();
+      expect(setAuthToken).toHaveBeenCalledWith('');
```
dangodev: So in my paranoia, I tested this. This test will successfully pass if someone touched the `restFetch` component and broke error handling there. So this test now successfully ensures that `restFetch` is handling errors, but if that breaks this will pass too because `catch()` is never called (a test will pass if `expect()` isn’t called)

src/utils/errorReport.ts
dangodev: +1—would like to have this added to the **Errors** doc, at least

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
dangodev: Cool you’re right here—seems to throw! But some other places (noted) the `catch()` won’t fire if something breaks. It’d be nice if our tests threw if our errors, uh… errored

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -43,27 +43,26 @@ export class ManifoldResourceContainer {
...
+      });
 
-    if (response instanceof Error) {
+      if (this.refetchUntilValid && (!response || response.state !== 'available')) {
+        this.timeout = window.setTimeout(() => this.fetchResource(this.resourceLabel), 3000);
+      }
+
+      this.resource = response;
+      this.loading = false;
+    } catch (error) {
```
dangodev: This change didn’t seem to be tested; I’d like to have a notice if this error handling broke

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -102,103 +102,94 @@ export class ManifoldResourceList {
...
-          label: resource.body.label,
-          name: resource.body.name,
-          logo: product && product.body.logo_url,
-          status: 'available',
-        });
-      });
-
-      this.resources = resources.sort((a, b) => a.label.localeCompare(b.label));
+        this.resources = resources.sort((a, b) => a.label.localeCompare(b.label));
+      }
```
dangodev: Same here—out of curiosity I tried reverting all your changes locally to see what would happen, and all the tests passed. Can you add a test here to ensure this stays?

src/utils/restFetch.spec.ts
dangodev: For clarity: I just removed the try/catch in `restFetch` but didn’t touch this test suite. It ran as if everything was working.

sslotsky: Ahh, you know what? I was looking for some kind of "expect to throw error" functionality and didn't find any on the first pass. But I just looked again and [it definitely exists](https://jestjs.io/docs/en/expect#tothrowerror). That's what we need here.

PR #349, opened 2019-08-07
Fix catalog auth

dangodev:
## Reason for change
A [previous, large PR](#313) accidentally committed code that required all catalog components to authenticate. This broke [our dev server](https://ui.manifold.now.sh) as well as our docs (until they were mocked). This also broke partner integration, as these can no longer be used on marketing sites.

This doesn’t change the functionality of `restFetch`, but instead just doesn’t send the `authorization` header with a bad token.

This relates a bit to https://github.com/manifoldco/engineering/issues/8907, but doesn’t really solve the authenticated issue.

## Testing
Inspect **the Storybook (ZEIT now deployment) link in this PR**, and confirm all the catalog components (public-facing) work: `manifold-marketplace`, `manifold-product`, `manifold-plan-selector`, etc. etc.

⚠️ No seriously really look through Storybook please

### DIFF
--- a/src/utils/restFetch.ts
+++ b/src/utils/restFetch.ts
@@ -27,17 +27,22 @@ export const createRestFetch = ({
   const url = `${endpoints[args.service]}${args.endpoint}`;
   const start = new Date();
 
-  while (!getAuthToken() && !hasExpired(start, wait)) {
+  // TODO: catalog should ALWAYS be able to fetch WITHOUT auth if needed,
+  // but this prevents the ability for it to auth altogether. We need both!
+  const isCatalog = args.service === 'catalog';
+
+  while (!getAuthToken() && !hasExpired(start, wait) && !isCatalog) {
     // eslint-disable-next-line no-await-in-loop
     await new Promise(resolve => setTimeout(resolve, 2000));
   }
 
-  if (!getAuthToken()) {
+  if (!getAuthToken() && !isCatalog) {
     return new Error('No auth token given');
   }
   try {
+    const token = isCatalog ? undefined : getAuthToken();
     const response = await fetch(url as string, {
-      ...withAuth(getAuthToken(), args.options),
+      ...withAuth(token, args.options),
       body: JSON.stringify(args.body),
     });
 
--- a/src/utils/restFetch.ts
+++ b/src/utils/restFetch.ts
@@ -31,18 +31,20 @@ export const createRestFetch = ({
   // but this prevents the ability for it to auth altogether. We need both!
   const isCatalog = args.service === 'catalog';
 
-  while (!getAuthToken() && !hasExpired(start, wait) && !isCatalog) {
-    // eslint-disable-next-line no-await-in-loop
-    await new Promise(resolve => setTimeout(resolve, 2000));
+  if (!isCatalog) {
+    while (!getAuthToken() && !hasExpired(start, wait)) {
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise(resolve => setTimeout(resolve, 2000));
+    }
+
+    if (!getAuthToken()) {
+      return new Error('No auth token given');
+    }
   }
 
-  if (!getAuthToken() && !isCatalog) {
-    return new Error('No auth token given');
-  }
   try {
-    const token = isCatalog ? undefined : getAuthToken();
     const response = await fetch(url as string, {
-      ...withAuth(token, args.options),
+      ...withAuth(getAuthToken(), args.options),
       body: JSON.stringify(args.body),
     });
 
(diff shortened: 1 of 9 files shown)


sslotsky:
I.... do not understand how this works :sweat_smile: 

It looks like the `restFetch` is supposed to throw an error if `getAuthToken()` doesn't return anything after the wait time is up. Why isn't it throwing an error here? :thinking:

Minivera:
Same, if the auth token is empty or invalid, it should throw here: https://github.com/manifoldco/ui/blob/master/src/utils/restFetch.ts#L35.

The only thing I can see happening here is if we have an auth token, but for some reason its value is `"undefined"`.

dangodev:
> I.... do not understand how this works 😅

Storybook doesn’t use the `<manifold-auth-token>` component, so the `getToken()` function will be undefined. If that component is present, even for catalog, then it’ll wait for the token to make the call. But if it’s missing, then it’ll just make the call (but during the refactor, it would send `Bearer undefined` to catalog, which was the issue).

We really, really, really, need to keep catalog components working without auth. We, along with all our partners, need to be able to use these on marketing sites.

Minivera:
Yes, but I don't think this here is the right spot to do that. If `getAuthToken` returns undefined because the `manifold-auth-token` component is missing, then the rest fetch function should always fail with the error sam mentioned. Even worse, we'll have to wait 15 seconds for anything to happen because of the timeout.

I think the work needs to happen in `restFetch` to give the ability to bypass the token check if the endpoint is meant to be public.

sslotsky:
> We really, really, really, need to keep catalog components working without auth

Agree 100%! I actually had planned to take this on myself, @Minivera and I had a plan for it. We can sync up on this today if you like!

stories/manifold-marketplace.stories.js
```diff
@@ -1,11 +1,9 @@
 import { storiesOf } from '@storybook/html';
 
 import markdown from '../docs/docs/components/manifold-marketplace.md';
-import { manifoldConnectionDecorator } from './connectionDecorator';
 
 storiesOf('Marketplace', module)
   .addParameters({ readme: { sidebar: markdown } })
-  .addDecorator(manifoldConnectionDecorator)
```
dangodev: Based on our convo in Slack, I removed auth from Storybook to verify that this is working. I did this for `manifold-marketplace`, `manifold-product`, and `manifold-plan-selector`.

But I left it for `manifold-product`, since I know that one is primarily authenticated? Open to feedback there.

src/utils/restFetch.ts
```diff
@@ -27,17 +27,22 @@ export const createRestFetch = ({
...
     // eslint-disable-next-line no-await-in-loop
     await new Promise(resolve => setTimeout(resolve, 2000));
   }
 
-  if (!getAuthToken()) {
+  if (!getAuthToken() && !isCatalog) {
     return new Error('No auth token given');
   }
   try {
+    const token = isCatalog ? undefined : getAuthToken();
```
dangodev: Without this last check, it would no longer wait, but still try to send _something_ as the auth token. This is disabled for now.

src/utils/restFetch.ts
```diff
@@ -27,17 +27,22 @@ export const createRestFetch = ({
   const url = `${endpoints[args.service]}${args.endpoint}`;
   const start = new Date();
 
-  while (!getAuthToken() && !hasExpired(start, wait)) {
+  // TODO: catalog should ALWAYS be able to fetch WITHOUT auth if needed,
+  // but this prevents the ability for it to auth altogether. We need both!
+  const isCatalog = args.service === 'catalog';
+
+  while (!getAuthToken() && !hasExpired(start, wait) && !isCatalog) {
```
sslotsky: This reads a little funny, having `!isCatalog` in a loop condition. Even though it will work, I think it would make more sense like this:

```ts
if (!isCatalog) {
  while (...) {
```

stories/manifold-marketplace.stories.js
sslotsky: I never knew about this decorator but I love it.

src/utils/restFetch.ts
sslotsky: What was `getAuthToken()` returning? I thought it would either return `undefined` or a valid auth token, is that not the case?

src/utils/restFetch.ts
dangodev: Agreed. I’ll change.

src/utils/restFetch.ts
dangodev: You’re right this is dumb. I did this before removing `<manifold-auth-token>` from the story like I should have. I removed this.

PR #351, opened 2019-08-08
Add gatsby-source-graphql

davidleger95:
## Reason for change

Add the ability to source content from the manifold GraphQL API.

I was going to convert a component as well in this PR as an example but it sounds like we might be holding off on that due to slowness in the GraphQL API. Is that correct? Just inferring that from the product page conversion being closed. maybe @dangodev can clarify?

How will we integrate this into docs ? Once components are converted to GraphQL, instead of having example components fetch their own data, use the `*-view` component for display and use a Gatsby `staticQuery` to load data at build time and pass that to the components.

## Testing

when running the docs in dev mode, go to `/___graphql`. the entire GraphQL API should be exposed via on the `manifoldApi` node. 

Example:

```graphql
{
  manifoldApi {
  # Query the Manifold API here
    products (first: 10) {
      edges {
        node {
          displayName
          logoUrl
        }
      }
    }
  }
}
```

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -72,5 +72,20 @@ module.exports = {
         ],
       },
     },
+    {
+      resolve: 'gatsby-source-graphql',
+      options: {
+        typeName: 'ManifoldAPI',
+        fieldName: 'manifoldApi',
+        // Url to query from
+        url: 'https://api.manifold.co/graphql',
+        // HTTP headers
+        headers: {
+          authorization: `Bearer ${process.env.MANIFOLD_API_TOKEN}`,
+        },
+        // Additional options to pass to node-fetch
+        fetchOptions: {},
+      },
+    },
   ],
 };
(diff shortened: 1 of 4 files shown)


dangodev:
> it sounds like we might be holding off on that due to slowness in the GraphQL API. Is that correct?

Yes; we need GraphQL to be faster before using it in our web components. But in regards to GraphQL, I think getting things set up early is great, _but_ in this scenario I have thoughts (below)

> How will we integrate this into docs ?

Honestly, the mocking should go away eventually. That was all a hack to get the catalog components working without auth. There’s a bigger problem of the catalog components not working on marketing pages.

In that light, I think piping all the data through Gatsby is probably not what we want, long-term? The components should fetch their own data. The fact that we had to mock them was probably a red flag more than anything. So mocking GraphQL in this PR… I’m not sure if we’ll need this?

davidleger95:
So we should close this PR then @dangodev? :P

dangodev:
> So we should close this PR then @dangodev? :P

I hate to say it, but I think it’s best 😞. But feel free to keep the branch or code around as we may need this in the future! Just not for our UI components

davidleger95:
no problem, sometimes things change :)

PR #352, opened 2019-08-08
Cleaning up some things 🧹 

davidleger95:
## Reason for change

- ~derovision~ deprovision
- remove duplication of "Manifold UI" in page title
- fix HTML tags in docs examples

### DIFF
--- a/docs/src/components/Page.tsx
+++ b/docs/src/components/Page.tsx
@@ -13,7 +13,7 @@ interface PageProps {
 function Page({ currentPage, links }: PageProps) {
   return (
     <>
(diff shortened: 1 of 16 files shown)


docs/src/components/Page.tsx
```diff
@@ -13,7 +13,7 @@ interface PageProps {
 function Page({ currentPage, links }: PageProps) {
   return (
     <>
-      <SEO title={`${currentPage.frontmatter.title} | Manifold UI`} />
```
dangodev: 👍

PR #353, opened 2019-08-08
Fix the resource card loading indicator

Minivera:
Work is related to manifoldco/engineering#8930

## Reason for change
It will be less wobbly now, it really didn't like to stay in place the little bugger!

## Testing
Test with storybook

### DIFF
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view.css
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view.css
@@ -154,7 +154,7 @@
     width: 0.75rem;
     height: 0.75rem;
     margin-right: 0.75rem;
-    transform-origin: 50% 40%;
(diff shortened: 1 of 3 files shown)


src/components/manifold-resource-card-view/manifold-resource-card-view.css
```diff
@@ -154,7 +154,7 @@
     width: 0.75rem;
     height: 0.75rem;
     margin-right: 0.75rem;
-    transform-origin: 50% 40%;
```
dangodev: 🤔 🤔 🤔 🤔 🤔 🤔 🤔

dangodev: That seemed like a magic number to me

PR #354, opened 2019-08-08
Fix the service card title dynamic size

Minivera:
Work is related to manifoldco/engineering#8934

## Reason for change
The dynamic sizing the the service card's first row made it so that the description would often look misaligned when products with different a number of lines for the description were put next to each other. This adds a fixed width for the first row of the service card.

Capture d’écran 2019-08-08 à 14 01 56

## Testing
Test with storybook, a new story for the marketplace called "curated" shows the behavior we wanted to fix.


davidleger95:
This breaks with longer names (ZeroSix). 

Does this solve the issue?

```css
grid-template-rows: auto 1fr auto;
align-items: start;
```

Screen Shot 2019-08-08 at 4 33 06 PM

Minivera:
I'll try!

PR #355, opened 2019-08-08
Fix the service card loading to keep loading when fetching if free

Minivera:
Work is related to manifoldco/engineering#8935

## Reason for change
The service card would render then start loading if the product is free, which led to the free badge jumping around a lot. Products will now load in different intervals, but will always be loaded with their badges on the first render.

## Testing
Test with the marketplace story on storybook.

---

# Drew’s Changes

## Reason for change
I wanted to satisfy Guillaume’s original ticket, but I had a few concerns:
- We now have 3 (THREE!) service card components doing basically the same work. Responsibilities aren’t clear between them.
- Guillaume’s work solved an issue, but the components were already were too complex and this PR didn’t help. It was very hard to follow what component was doing what
- In addition to all the above, and some much-needed refactoring of `<manifold-service-card>`, I noticed something: **we actually had broken the service card links**, and had no tests to catch them. This is a real problem!

Hopefully you can see my thinking with this—I wanted to complete the ticket, but felt that **us overcomplicating very simple components led directly to a breakage without any of us knowing**. Too many components, too much debt, not enough E2E tests.

I know that creating another PR in the midst of everything else going on may contribute to the problem, but hopefully the fact that many of my changes were test additions / improvements it’ll give us a little more assurance. A legitimate question to this would be “why not in another PR?” And I actually started the work in another PR. But I wanted to keep most of Guillamue’s changes, and the work related to this touched all the same spots anyway, so it felt like a natural addition (don’t worry, Guillaume—I kept your changes in this PR aside from the ones that delayed loading, which we can discuss later).

Anyway, here are the changes I made:

### New E2E Tests

We weren’t doing nearly enough E2E testing. I added tests to `<manifold-marketplace-grid>` that now tests `<manifold-service-card-view>` and `<manifold-template-card>`. In this way we get all the benefits of unit tests on the service card, and ensure we don’t break integrations as we had done (unit tests aren’t enough!).

Screen_Shot_2019-08-12_at_23_13_17

In other areas, existing tests were added to be more reliable, for example we were testing `[data-*]` attributes in places, rather than actual properties passed to the components. Now we’re testing actual properties passed, because there’s no guarantee those things will always be the same.

While these changes aren’t 100% foolproof, they’re a step in the right direction.

### New Spec Tests

We also weren’t testing that the `click` event is firing from the service card. Now we are (this didn’t seem to be broken, but it could have just as easily broken without any of us knowing, again, because it wasn’t tested).

### Simplified structure

The following changes were made to these components:

- `<manifold-service-card>` now fetches data, and contains no markup.
- `<manifold-service-card-view>` now contains markup, and fetches no data.
- `<manifold-marketplace>` now handles loading state, and no markup
- `<manifold-marketplace-grid>` no longer handles any loading state or data, and is now a pure view component

Again, I realize it’s a lot, but these components are so tightly-connected it was hard to break up the PR into multiple steps.

## Testing
Test `<manifold-marketplace>` and `<manifold-service-card>` in Storybook. Look for the following things:

- [ ] Is the “free” bag popping in after load?
- [ ] Are components loading quickly?
- [ ] Does the link work in the default marketplace component now?


happoRender.js
```diff
@@ -0,0 +1,6 @@
+export default component => `
+  <manifold-connection>
+      <manifold-auth-token token="something|${new Date(Date.now() + 6.04e8).getTime()}"/>
+      ${component}
+    </manifold-connection>
```
davidleger95: nitpick, but indentation isn't aligned

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -26,6 +26,7 @@ export class ManifoldServiceCard {
   @Prop() preserveEvent?: boolean = false;
   @Prop() skeleton?: boolean = false;
   @State() loading: boolean = false;
+  @State() free: boolean = false;
```
davidleger95: wondering why we need a prop and state for this

happoRender.js
sslotsky: We should submit this happo fix as another PR since it's blocking everyone else

sslotsky: https://github.com/manifoldco/ui/pull/356

src/components/manifold-service-card/manifold-service-card.tsx
Minivera: Because the prop was used as both a prop and the state, it was very hard to check if the prop was set externally or internally. This meant that deciding if we should fetch the plans was a big pain. Having both separated makes this easier and makes a bit more sense for the dev I think.

dangodev: I personally find this more confusing. I’m not sure which should take priority when. Generally when you separate the same thing into props & state, it just ends up as a potential source of conflict as soon as someone else touches this

dangodev: To clarify: I think what you did in this PR makes sense, but I think the larger structure doesn’t (not your fault). I think having `manifold-service-card` be a “smart” component is where the issue lies.

It’d be nice to remove all the fetching from the `manifold-marketplace` child altogether, and use the “dumb“ version. And have this free fetching in the “smart” version that only gets used as a standalone. Overall I think we need to improve our smart vs dumb separation, because there’s some brittle logic on what should fetch, when.

Minivera: Due to how watches and props work in stencil, it was impossible for me to know if the `isFree` property (When mutable) was unset externally. This led to issue like the manifold-marketplace not setting the `isFree` prop on service cards and those cards then trying to fetch the plans to see if the service was free, even though we already knew.

I can rename the props if necessary, but the separation is sadly needed from what I could observe.

See it like this: The Prop sets the base state and then the sate takes over, the prop is never used anywhere else.

dangodev: Proposal: can we just use `manifold-service-card-view` as a child of `manifold-marketplace-grid`? I think that’d address a lot of my concerns.

dangodev: And also: this is 100% my fault originally! I added the free fetching in the service card as a check, but that was before the child view was added. So this isn’t on you; I just want to course-correct from the current path.

src/components/manifold-service-card/manifold-service-card.e2e.ts
```diff
@@ -11,41 +10,4 @@ describe('<manifold-service-card>', () => {
...
-        elm.isFree = true;
-        elm.product = props.product;
-      },
-      { product: Product }
-    );
-    await page.waitForChanges();
-
-    const tag = await page.find('manifold-service-card-view >>> [data-tag="free"]');
-    expect(tag).not.toBeNull();
-  });
```
dangodev: This is tested in the spec page now (and doesn’t work now that we’re waiting to fetch)

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -67,10 +67,10 @@ export class ManifoldServiceCard {
 
   componentWillLoad() {
     if (this.skeleton || this.product) {
-      return null; // if skeleton UI or it’s passed a product, don’t fetch anything
+      return; // if skeleton UI or it’s passed a product, don’t fetch anything
     }
 
-    return this.fetchProduct({ id: this.productId, label: this.productLabel });
```
Minivera: Should we always return the promise when fetching something?

dangodev: Oh it’s actually really important we _not_ return the promise; otherwise the loading state never shows. You should do either one—either return the promise and get rid of the loading state, or never return the promise and keep the loading state.

In general I much prefer the loading state because users see something rendered and loading much, much sooner. It improves the perceived performance of our components.

dangodev: I changed this back because I was assuming by rearranging the components this was no longer needed

Minivera: We talked about loading states with @seriousben, @sslotsky and @davidleger95 and one thing that came up was that out loading state weren't that great for platforms, came with a lot of issues and increased the perceived lag for pages. Though that is something we'll have to discuss in person.

dangodev: That’s fine! But let’s please talk about that, come to a decision, and enforce it universally and not sneak changes like this into PRs. If we were to go this route we should be stripping out all the loading states & skeleton UIs from this component tree.

src/components/manifold-marketplace-grid/manifold-marketplace-grid.e2e.ts
```diff
@@ -52,192 +56,326 @@ describe('<manifold-marketplace-grid>', () => {
...
-      (elm: any, props: any) => {
-        elm.services = props.services;
-      },
-      { services }
-    );
-    await page.waitForChanges();
-    const search = await page.findAll('manifold-marketplace-grid >>> [type="search"]');
-    expect(search).toHaveLength(1);
+  describe('v0 API', () => {
+    it('[excludes] accepts exclusion list of products', async () => {
```
dangodev: Now the E2E tests are renamed to clearly indicate which property is testing. I’d like to be able to look through the test output, and quickly see we’re testing all a components’ props in a meaningful way.

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -213,12 +200,16 @@ export class ManifoldMarketplaceGrid {
...
       isFeatured={this.featured && this.featured.includes(product.body.label)}
-      isFree={this.freeProducts.includes(product.id)}
-      product={product}
+      isFree={this.freeProducts && this.freeProducts.includes(product.id)}
+      logo={product.body.logo_url}
+      name={product.body.name}
+      preserveEvent={this.preserveEvent}
+      productId={product.id}
+      productLabel={product.body.label}
+      productLinkFormat={this.productLinkFormat}
```
dangodev: We weren’t passing `productLinkFormat` or `preserveEvent` to the service cards, which was breaking `manifold-marketplace`. Tests were added to prevent this in the future.

src/components/manifold-marketplace-grid/manifold-marketplace-grid.e2e.ts
dangodev: No existing tests were harmed in the process 🙂 (only renamed)

src/components/manifold-service-card-view/manifold-service-card-view.css
```diff
@@ -29,38 +29,43 @@
...
+  opacity: 0.6;
 
   @media (--viewport-m) {
-    grid-template-columns: var(--manifold-product-logo-l) auto;
-    padding: 0.875rem;
+    margin-top: 0.375rem;
   }
 }
 
-.card-wrapper {
```
dangodev: We were adding/removing classes that were duplicating styles. I deduped styles and centralized everything to still work with Render (hopefully), but made editing styles easier to follow.

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -137,54 +103,38 @@ export class ManifoldServiceCard {
...
   render() {
-    return !this.skeleton && !this.loading && this.product ? (
-      <a
-        role="button"
-        itemscope
-        itemtype="https://schema.org/Product"
-        itemprop="url"
-        href={this.href}
-        onClick={this.onClick}
-        style={{ textDecoration: 'none' }}
```
dangodev: No more `!important` inline styles

docs/docs/changelog.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fixed the service card loading the free badge after rendering, which caused a jumpy UI.
```
sslotsky: Does this file get automatically copied over from the `CHANGELOG.md` at the project root?

src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -114,6 +116,8 @@ export class ManifoldMarketplace {
 
   @logger()
   render() {
+    const isLoading = !this.services.length || !this.freeProducts; // wait for free calls to finish
```
sslotsky: I'm not sure I understand how `freeeProducts` is being used. Where does this ultimately come from?

docs/docs/changelog.md
dangodev: Yup!

src/components/manifold-marketplace/manifold-marketplace.tsx
dangodev: Ultimately we should be able to get this from GraphQL, but for now we can only get free products by querying _all_ the plan data. `manifold-marketplace` is the one that fetches all this, and passes it to children.

The main goal of this PR was to have marketplace wait until everything is loaded before exiting the loading state.

src/components/manifold-service-card-view/manifold-service-card-view.e2e.tsx
```diff
@@ -0,0 +1,74 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+/* eslint-disable no-param-reassign, @typescript-eslint/no-explicit-any */
+
+// Things tested in other places:
+// - link formatting: manifold-marketplace-grid.e2e.tsx
+// - click events: manifold-service-card.spec.ts
+
+describe('<manifold-service-card-view>', () => {
```
sslotsky: How many of these scenarios could just be happo screenshots?

src/components/manifold-service-card-view/manifold-service-card-view.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+        ></manifold-service-card-view>`,
+    });
+    const { doc, root } = page;
+    doc.addEventListener('manifold-marketplace-click', mockClick);
+    const link = root && root.shadowRoot && root.shadowRoot.querySelector('a');
+    if (!link) {
+      return;
+    }
+    link.click();
+    expect(mockClick.mock.calls[0][0].detail).toEqual({
```
sslotsky: I'm pretty sure I saw a better way to do this with `expect.objectContaining`

https://jestjs.io/docs/en/expect#expectobjectcontainingobject

src/components/manifold-service-card-view/manifold-service-card-view.e2e.tsx
dangodev: 🤔 what would your preference be? Should they overlap, or just be moved to VRT entirely?

sslotsky: In general I'd say that if Happo can capture what we're trying to test, let's just do it with Happo. Basically, if we're just testing the output of some component, the main disadvantage of doing it with e2e tests is that they can break if we change an implementation detail. For instance, if we changed the `data-testid` or a component tag name or something like that. 

If we write a Happo test, the first time we generate the screenshot, we examine it to see if it displays our button or skeleton or whatever else. And from then on, Happo will alert us only if that changes, and won't care if the name of the component or data attribute or any other detail happened to change.

dangodev: Cool. I’m down. I’ll remove the tests I moved to Happo.

I do have some reservations about someone just blindly approving changes without inspecting them, but overall I think you’re right—this is safer and easier. In general I’d like to encourage more Happo testing as a great E2E option.

sslotsky: That's a fair point about not inspecting the diffs. The approval button is a powerful thing and shouldn't be taken lightly. I'm optimistic in this case though. When I recall Jest snapshots, the problem wasn't _just_ that the diffs were easy to dismiss, but also that it was difficult to tell if a change would really impact the visual output of the component, and diffs were frequent and noisy. We do wind up with some spurious diffs in VRT, but in general I think diffs are far less frequent, and the ability to compare visually when you do have a diff is game changing.

dangodev: new tests 🎉 
Screen Shot 2019-08-13 at 09 43 39

dangodev: I couldn’t test the `manifold-resource-product` properly because of the state tunnel (it kept un-setting the data I was giving it, and showed the loading state every time) 😕. But at least this will tell us if it’s not loading the service card beneath.

dangodev:
I removed my changes requested, but it’d be nice if someone else is able to ✅ this PR

src/components/manifold-service-card-view/manifold-service-card-view.e2e.tsx
sslotsky: Oh damn! Yeah I've definitely been bitten by state tunnels when I'm trying to test stuff in Happo and it's one of the things that gives me a lot of concern over the long term viability of state tunnels. They don't seem very active over there despite known issues that they've acknowledged need addressing.

PR #360, opened 2019-08-09
Update the auth-token event name and document how it works

Minivera:
## Reason for change
This changes the `manifold-auth-token` event name from whatever stencil chose to a fixed `manifold-token-received`. This will also ensure that this event can bubble up to the document rather than stay locked to the component.

This PR also updates the docs to better document the auth token component with the new shadowcat behavior.

## Testing
Test with the docs


docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Auth token provider
+path: /component/auth-token
+---
+
+# Auth Token Provider
+
+Data component that allows a developer to receive or provide the required bearer auth token for accessing data locked (🔒)
```
sslotsky: For `.md` files I prefer to keep paragraphs in a single line and view them with word wrap on. It makes them much easier to edit.

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
...
+behind authentication. The component will render the [shadowcat](https://github.com/manifoldco/shadowcat) authentication
+iframe and attempt to log the currently logged in user for your platform using oauth. See the shadowcat documentation for more information.
```
sslotsky: I think "attempt to log" should be changed either to "attempt to log in" or "attempt to authenticate"

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
...
+
+```html
+<manifold-auth-token></manifold-auth-token>
+```
+
+The component can be placed anywhere in the DOM tree as long as it exists within a `<manifold-connection>` component.
+
+## Receiving the token
+The component makes no decision as to how you should save to token on your side. As such, when the token is received from the iframe,
```
sslotsky: Should be "save the token"

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
...
+and event is triggered that will give you the token. The token is automatically injected into the connection once received,
```
sslotsky: Should be "an event is triggered"

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
...
+
+The component can be placed anywhere in the DOM tree as long as it exists within a `<manifold-connection>` component.
+
+## Receiving the token
+The component makes no decision as to how you should save to token on your side. As such, when the token is received from the iframe,
+and event is triggered that will give you the token. The token is automatically injected into the connection once received, 
+but subsequent requests will need the token to prevent another oauth request.
+
+```js
+document.addEventListener('manifold-token-received', ({ detail: { token } }) => {
```
sslotsky: @davidleger95 I believe in the past you've advised against using the past tense for event names. I don't have an issue with this event name but thought you might like to weigh in.

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,37 @@
...
+  // create a cookie or localstorage value with the token
+});
+```
+
+## Setting the cached token
+
+The component can receive a token previously saved from an oauth request to speed up all requests made by our components.
+If this token is provided, the oauth request will still happen in order to refresh the token, but any fetch calls
+happening in our web components will not wait for that oauth request to finish.
```
sslotsky: Do you think we should add a note about expired tokens at all?

docs/docs/components/manifold-auth-token.md
davidleger95: yes, i did :) html events are all present tense. `click`, `change`, `hover`, `load`, etc. so it would be preferable to use `receive-manifold-token` (or would it be `manifold-receive-token`? naming is hard).

Here's the example I'm looking at: https://github.com/manifoldco/ui/blob/6a4e5d011ee8f8111f87ab77ed57c6f23bce4376/src/components/manifold-resource-rename/manifold-resource-rename.tsx#L13

davidleger95: Looking at more examples I think `manifold-token-receive` (closer to what you have) fits our naming conventions best

docs/docs/components/manifold-auth-token.md
Minivera: Hum... Good question.

docs/docs/components/manifold-auth-token.md
Minivera: All our docs are written like this, many editor strangely don't support word wrap for markdown.

davidleger95:
I'll let @sslotsky put the ✅on this. He had more change requests than I did

docs/docs/components/manifold-auth-token.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Setting the cached token
+
+The component can receive a token previously saved from an oauth request to speed up all requests made by our components.
+If this token is provided, the oauth request will still happen in order to refresh the token, but any fetch calls happening
+in our web components will not wait for that oauth request to finish.
+
+```html
+<manifold-auth-token token="new-token"></manifold-auth-token>
+```
```
dangodev: ”Question: should we have an “Authentication” guide under ”Advanced” in the docs? I think having both would be great. It can largely be the same information, but I would wager that almost every reader would click on an “Authentication” link as a guide, but it would take them a while to find the auth token component, and connect the dots that’s what they should use for auth.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the resource card loading indicator - for a lack of a better word - wobbling around like its life depended on it.
 - Fixed the title of the `service-card` taking a dynamic amount of space and making the description look misaligned.
 
+### Changed
+- Changed the event name for the `manifold-auth-token` component from the stencil auto-generated name to `manifold-token-received` and documented that event.
```
sslotsky: Is this event name still accurate?

Minivera: Nope!

docs/docs/advanced/authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+The component can be placed anywhere in the DOM tree as long as it exists within a
+`<manifold-connection>` component.
+
+## Receiving the token
+
+The component makes no decision as to how you should save the token on your side. As such, when the
+token is received from the iframe, an event is triggered that will give you the token. The token is
+automatically injected into the connection once received, but subsequent requests will need the
+token to prevent another oauth request.
```
sslotsky: This last sentence is a little confusing, and you've explained setting the cached token in a later section. Maybe here we should say something more like this:

> The token will be stored in the connection for use in subsequent API requests, but this event gives you the opportunity to save the token to prevent delays on the next page load [as described in another section](#setting-the-cached-token).

src/components.d.ts
```diff
@@ -1077,7 +1077,7 @@ declare namespace LocalJSX {
   }
   interface ManifoldAuthToken extends JSXBase.HTMLAttributes<HTMLManifoldAuthTokenElement> {
     'oauthUrl'?: string;
-    'onManifoldOauthTokenChange'?: (event: CustomEvent<any>) => void;
+    'onManifold-token-received'?: (event: CustomEvent<any>) => void;
```
sslotsky: I think you might need a build for this to reflect the changed event name.

sslotsky:
Also looks like the `now` deployment failed.

src/components.d.ts
dangodev: Fixed

dangodev:
@sslotsky I know Guillaume said “test in the docs“ but I’m not really sure what that means. Would you be OK with pushing this through yourself? I don’t know if I have context to approve this PR (nothing jumps out at me)

sslotsky:
> @sslotsky I know Guillaume said “test in the docs“ but I’m not really sure what that means. Would you be OK with pushing this through yourself? I don’t know if I have context to approve this PR (nothing jumps out at me)

@dangodev Maybe this just means we make sure the documentation looks good in the context of the docs page?

docs/docs/advanced/authentication.md
```diff
@@ -0,0 +1,69 @@
...
+
+The component can be placed anywhere in the DOM tree as long as it exists within a
+`<manifold-connection>` component.
+
+## Receiving the token
+
+The component makes no decision as to how you should save the token on your side. As such, when the
+token is received from the iframe, an event is triggered that will give you the token. The token will be
+stored in the connection for use in subsequent API requests, but this event gives you the opportunity to
+save the token to prevent delays on the next page load [as described in the next section](#setting-the-cached-token).
```
sslotsky: Sooo this link doesn't work in the docs page. @dangodev do you know if there's any way to get the markdown links within the docs page to point to another section within the same page? I thought that normally in markdown the headings have an id that's created by dasherizing the text. But it seems our headings don't have IDs at all.

If we can't link by id, what do you think would be the best way to point users to another section where some information is displayed in further detail?

dangodev: > normally in markdown the headings have an id that's created

That’s actually not a function of markdown; that’s a nice-to-have thing that platforms like GitHub generate on top.

We looked into doing this, and there are easy-to-use libraries that pull this off. But the problem is scrolling in Gatsby—it’s a well-known issue that as soon as the page loads, Reach Router scrolls to the top, and there’s not a way (that I could find) that doesn’t scroll back to top on page load (there’s supposed to be a Gatsby API for this that comes up when you search it, but that seems to be broken).

dangodev: > If we can't link by id, what do you think would be the best way to point users to another section

Just naming the section, possibly **In Bold** is helpful. In general you want to avoid relative terms like “next section” when writing; it’s all subject to change.

sslotsky: Troof

sslotsky: So for example,

> prevent delays on the next page load as described in **Setting the cached token**.

??

sslotsky: Although the section name is also subject to change :joy:

dangodev:
Also suggestion: is there a test where we double-check the name of the event? Event names are so brittle, it’d be nice to have each event be tested independently, that way we know there are no typos in the names.

sslotsky:
Good idea, I'll add one

docs/docs/changelog.md
```diff
@@ -15,14 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed the service card loading the free badge after rendering, which caused a jumpy UI.
 - Simplied `<manifold-service-card>` (data) and `<manifold-service-card-view>` (“dumb” view)
-
-### Fixed
```
sslotsky: Hmm, I don't think I intended to remove this? Might have been a funky rebase.

PR #361, opened 2019-08-12
Mark plan cost API as public to prevent auth token polling

sslotsky:
## Reason for change

The plan selector currently won't work when not authenticated because the calls to get the plan cost are polling for the auth token. But this is a public API, so it should not stop the plan selector from working.

## Testing

Make sure the plan selector works in storybook with no authentication.

### DIFF
--- a/src/utils/plan.ts
+++ b/src/utils/plan.ts
@@ -320,6 +320,7 @@ export function planCost(
   return restFetch<Gateway.Price>({
     service: 'gateway',
     endpoint: `/id/plan/${planID}/cost`,
+    isPublic: true,
     body: { features },
     options: {
       method: 'POST',
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1030,23 +1030,6 @@
         "@manifoldco/gql-zero": "^0.2.0",
         "@manifoldco/shadowcat": "^0.1.5",
         "@stencil/state-tunnel": "^1.0.1"
-      },
-      "dependencies": {
-        "@manifoldco/gql-zero": {
-          "version": "0.2.0",
-          "resolved": "https://registry.npmjs.org/@manifoldco/gql-zero/-/gql-zero-0.2.0.tgz",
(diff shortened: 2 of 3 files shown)


sslotsky:
@Minivera @dangodev check the Happo diff :thinking: 

I'm not sure why the custom plan for Jaws DB had a "Free" badge on it before, but it seems like what shows up in the new screenshot is correct. Want to make sure I'm not reading this wrong.

dangodev:
> I'm not sure why the custom plan for Jaws DB had a "Free" badge on it before, but it seems like what shows up in the new screenshot is correct. Want to make sure I'm not reading this wrong.

Ah yeah—I’d like to know as well 🤔.

Unrelated, but why are the plans out of order? They should be in descending order of cost, with custom at the end.

sslotsky:
> They should be in descending order of cost, with custom at the end.

@dangodev The sorting happens in `manifold-plan-selector`, which isn't being used here. We are feeding the API data directly into the `manifold-active-plan`.

src/utils/plan.ts
```diff
@@ -320,6 +320,7 @@ export function planCost(
   return restFetch<Gateway.Price>({
     service: 'gateway',
     endpoint: `/id/plan/${planID}/cost`,
+    isPublic: true,
```
dangodev: Since the logic was changed in `restFetch`, we should probably add this to the catalog components as well

dangodev: Actually this may be fine as-is, so long as we cover one scenario:
1. User logs in, has auth token set
2. User logs out, auth token persists in localStorage?
3. User navigates to marketing site. They should still be able to view catalog components that don’t try and authenticate

sslotsky: The `isCatalog` check is still there in `restFetch`, right? So catalog endpoints will always be public, but this allows us to declare other endpoints as public too, which we need for some Gateway URLs.

dangodev: Oh I missed the `isCatalog ||` bit! I see.

PR #362, opened 2019-08-12
Expose the credentials buttons colors for more customization

Minivera:
Work is related to manifoldco/engineering#8940

## Reason for change
This gives platforms the ability to customize the credentials buttons to their theme without needing too many CSS variables. The prop is passed through all the wrapper components to the credentials view component.

Capture d’écran 2019-08-12 à 10 23 22

## Testing
Test with the docs and set the two properties manually in the dom.

### DIFF
--- a/src/components/manifold-credentials-view/manifold-credentials-view-happo.ts
+++ b/src/components/manifold-credentials-view/manifold-credentials-view-happo.ts
@@ -2,7 +2,7 @@ import credentials from '../../spec/mock/cms-stage/credentials.json';
 import fromJSON from '../../spec/mock/fromJSON';
 
 export const credsHidden = () => {
-  const creds = document.createElement('manifold-resource-credentials-view');
+  const creds = document.createElement('manifold-credentials-view');
   creds.resourceLabel = 'test';
 
   document.body.appendChild(creds);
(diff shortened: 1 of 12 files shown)


dangodev:
What’s the need for this functionality?

Minivera:
Don't they already have full control using the themability of the manifold button itself? For the use case, see the linked issue.

dangodev:
I see a ticket for margin within the card, then at the end you give a suggestion of adding additional theming properties.

Reason I’m asking about this is because the more options we add like this, the more we have to support, and ensure they don’t break. This PR adds functionality and more long-term maintenance work for us, but doesn’t fully open up customization to platforms. Here, I‘d like to think of a clever way to **expose the button entirely** to platforms (i.e. they use their own React-controlled, React-styled button)—would that be possible, via event listeners or anything?

src/components/manifold-credentials-view/manifold-credentials-view-happo.ts
```diff
@@ -21,7 +21,7 @@ export const credsShown = () => {
 };
 
 export const resourceLoading = () => {
-  const creds = document.createElement('manifold-resource-credentials-view');
+  const creds = document.createElement('manifold-credentials-view');
```
dangodev: Let’s add some happo stories to test the button colors!

Minivera:
My issue might not be clear enough 🤔 It has two parts, one is about making the button customizable so render can have it empty rather than black. The other is about margin, which does not happen in the render dashboard.

I'd be much more happy with slots, I've been trying to not break the current behaviors of components too much so we don't suddenly remove the button when platforms are starting to use this.

PR #373, opened 2019-08-13
Add graphql query for product logos

davidleger95:
## Reason for change

- The REST endpoint is slow and gets a lot of unnecessary data. To speed this up, we're switching it to a GraphQL query which is much faster.

## Testing

- NOTE: this is currently blocked by auth not working properly for catalog data, so cannot test yet. :(
- Ensure the Resource List component loads properly with logos in Storybook. (Authenticated so will need a valid token)

### DIFF
--- a/src/components/manifold-resource-list/manifold-resource-list.tsx
+++ b/src/components/manifold-resource-list/manifold-resource-list.tsx
@@ -1,10 +1,12 @@
 import { h, Component, Prop, State, Element, Watch } from '@stencil/core';
+import { gql } from '@manifoldco/gql-zero';
 
 import { Marketplace } from '../../types/marketplace';
-import { Catalog } from '../../types/catalog';
 import { Provisioning } from '../../types/provisioning';
+import { ProductEdge, ProductConnection, Query } from '../../types/graphql';
 import Tunnel from '../../data/connection';
 import { RestFetch } from '../../utils/restFetch';
+import { GraphqlFetch } from '../../utils/graphqlFetch';
 import logger from '../../utils/logger';
 
 interface FoundResource {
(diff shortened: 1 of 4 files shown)


domenicrosati:
Unblocked!

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -108,15 +112,27 @@ export class ManifoldResourceList {
...
+            products(first: 500) {
+              edges {
+                node {
+                  label
+                  logoUrl
+                }
+              }
+            }
+          }
+        `,
```
dangodev: Note: with the PR just merged, we’ll have to add `isPublic: true` to this GraphQL query (as well as any other catalog-ish queries)

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -108,15 +112,27 @@ export class ManifoldResourceList {
     }
 
     if (Array.isArray(operationsResp)) {
-      const productsResp = await this.restFetch<Catalog.Product[]>({
-        service: 'catalog',
-        endpoint: `/products`,
+      const { data, errors } = await this.graphqlFetch<Query>({
```
dangodev: Let’s not wait for this to finish. We can achieve a nice speed boost by firing this query alongside the `restFetch` query in parallel. There are a couple ways to do this, either by waiting for the 2nd one but not the 1st, or using `Promise.all()`. But it’s a waste of milliseconds to defer the resource query (arguably the more important one) until after GraphQL when one doesn’t depend on the other.

src/utils/graphqlFetch.ts
```diff
@@ -23,6 +23,8 @@ export interface GraphqlResponseBody<T> {
   errors?: GraphqlError[];
 }
 
+export type GraphqlFetch = <T>(body: GraphqlRequestBody) => GraphqlResponseBody<T>;
```
dangodev: I think in my testing PR I added this, or created something similar. Rebase off `origin/master` to grab this

PR #374, opened 2019-08-13
Add missing coverage for restFetch consumption points

sslotsky:
## Reason for change

A previous PR for reporting errors from `restFetch` left a lot of blocks with no test coverage. This PR adds tests for most of those uncovered areas.

## Testing

CI should still pass! This is only test code.


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -6,7 +6,7 @@ import logger from '../../utils/logger';
 const baseWait = 15000;
 
 @Component({ tag: 'manifold-connection' })
-export class ManiTunnel {
```
dangodev: Goodnight, sweet prince

PR #376, opened 2019-08-13
Added: Better GraphQL error logging

dangodev:
## Reason for change
Updates `graphqlFetch` with the improvements @sslotsky added to `restFetch` in #347 for better error reporting

## Testing
Tests should pass.


src/utils/graphqlFetch.ts
```diff
@@ -7,11 +8,9 @@ interface CreateGraphqlFetch {
...
 }
 
-export interface GraphqlRequestBody {
-  query?: string;
-  mutation?: string;
-  variables?: object;
-}
+type graphqlArgs =
+  | { mutation: string; variables?: object; isPublic?: boolean }
+  | { query: string; variables?: object; isPublic?: boolean }; // require query or mutation, but not both
```
dangodev: Now when making a gql query, either `query` or `mutation` is required, but not both

src/utils/graphqlFetch.ts
```diff
@@ -23,56 +22,56 @@ export interface GraphqlResponseBody<T> {
   errors?: GraphqlError[];
 }
 
-export const createGraphqlFetch = ({
+export function createGraphqlFetch({
```
dangodev: All the changes in this file are mostly cosmetic, to align with `utils/restFetch`

src/utils/graphqlFetch.ts
```diff
@@ -23,56 +22,56 @@ export interface GraphqlResponseBody<T> {
...
+      method: 'POST',
+      headers: {
+        'content-type': 'application/json',
+        ...(getAuthToken() ? { authorization: `Bearer ${getAuthToken()}` } : {}),
+      },
+      body: JSON.stringify(request),
+    });
 
-  const body = await response.json();
+    // Don’t handle success; we always need data returned
```
dangodev: The only main difference is in `restFetch` we return `undefined` for a status code `> 201 && < 300`. But in gql we need everything, so we always return the response for `2xx`

src/utils/graphqlFetch.ts
```diff
@@ -23,56 +22,56 @@ export interface GraphqlResponseBody<T> {
...
+      throw new Error('Auth token expired');
+    }
 
-  return body;
-};
+    // Sometimes messages are an array, sometimes they aren’t. Different strokes!
+    report(response);
+    const message = Array.isArray(body) ? body[0].message : body.message;
+    throw new Error(message);
+  };
```
dangodev: We’re not really returning errors to the front-end like we should, but I think that’s a separate discussion.

src/utils/graphqlFetch.ts
```diff
@@ -23,56 +22,58 @@ export interface GraphqlResponseBody<T> {
...
-    body: JSON.stringify(requestBody),
-  });
+    const response = await fetch(endpoint, {
+      method: 'POST',
+      headers: {
+        'content-type': 'application/json',
+        ...(getAuthToken() ? { authorization: `Bearer ${getAuthToken()}` } : {}),
+      },
+      body: JSON.stringify(request),
+    });
```
sslotsky: In the `restFetch` version, there is a `.catch()` at the end of `fetch` to account for `5xx` errors. I would think we need that here too?

src/utils/graphqlFetch.ts
sslotsky: I might be open to either way here... :thinking: 

On one hand, it's probably legit to return the response, because your response type accounts for both `data` and `error`.

On the other hand, I think we want to force consumers to consider error scenarios, which we could probably do by setting it up such that if there's no data present, _there must be an error_. Making both `data` and `error` optional on the response type allows consuming code to act as though neither one exists. 

So I think there's two ways to accomplish that goal. One is to have this function return _either_ a GraphQL response body _or_ a GraphQL error, instead of an object that might contain either. The other is to have this function either retrurn a GraphQL response body or _throw_ the error. Throwing the error makes it more consistent with `restFetch`, but if we decided that it's better to return some error type for these scenarios, we'd probably want `restFetch` to do the same, again for consistency.

Thoughts?

src/utils/restFetch.ts
```diff
@@ -48,8 +48,7 @@ export function createRestFetch({
       }
     }
 
-    const url = `${endpoints[args.service]}${args.endpoint}`;
-    const response = await fetch(url as string, {
+    const response = await fetch(`${endpoints[args.service]}${args.endpoint}`, {
```
sslotsky: Lol, I'm at least glad that `as string` was removed, what in the actual heck was that doing there?

src/utils/graphqlFetch.ts
dangodev: So on second glance I realized we’re not returning `data`; the server returns that. I think it returns the `error` part as well.

Maybe let’s leave it like it is now, because like `5xx` there will be some errors that are network/code errors and not GraphQL errors. That’s what we’re handling here, and probably what we’re more concerned about anyway, yeah?

src/utils/restFetch.ts
dangodev: 🤷🏻‍♂️ I’m all for variables as names but to me this was noise, and yeah the forced typing was weird

src/utils/graphqlFetch.ts
sslotsky: Ah I see what you mean. Ok, let's just return here. We might wind up rethinking whether or not we throw `4xx` errors in `restFetch` too, but hopefully that will go away sooner than later and we won't have to worry about it.

src/utils/graphqlFetch.ts
dangodev: Nice…
( •_•)>⌐■-■ 
…`catch()` 
(⌐■_■)

src/utils/graphqlFetch.spec.ts
```diff
@@ -73,17 +76,36 @@ describe('The fetcher created by createGraphqlFetch', () => {
...
     fetchMock.mock(graphqlEndpoint, {
-      status: 200,
+      status: 500,
       body,
     });
 
-    const result = await fetcher({});
+    expect.assertions(2);
+    return fetcher({ query: '' }).catch(result => {
+      expect(fetchMock.called(graphqlEndpoint)).toBeTruthy();
```
sslotsky: I've noticed you frequently use `toBeTruthy()` on the fetch mock instead of `toBeTrue()`, any particular reason?

dangodev: Oh that was a copy + paste from other tests. I personally don’t like it. I’ll change.

src/utils/graphqlFetch.ts
```diff
@@ -23,56 +22,62 @@ export interface GraphqlResponseBody<T> {
...
+      report(response);
+      throw new Error('Auth token expired');
+    }
 
-  return body;
-};
+    // Sometimes messages are an array, sometimes they aren’t. Different strokes!
+    report(response);
+    const message = Array.isArray(body) ? body[0].message : body.message;
+    throw new Error(message);
```
sslotsky: Some questions about this

1. I thought we were just going to return the response now. Did I misinterpret how that conversation went?
2. Do we still either have a string or an array for GraphQL errors? I thought we were moving towards consistency. Can we look into this?
3. If we are going to throw (and this goes for the `restFetch` as well), I feel like just having the message isn't enough for callers to recover. They should have the status as well. So if we continue to throw, this might be a case for a custom error type, and if we're not going to throw, we might have to adjust the return types to allow for the ability to return a status.

dangodev: > Did I misinterpret how that conversation went?
Oh I think I misinterpreted your comment as “leave as-is and keep it similar to restFetch.” But maybe later today I’ll ping you with questions.
>Do we still either have a string or an array for GraphQL errors?
Those are always arrays, but they’re returned in `body.error`. I need to check on status codes for GQL errors though; I know statuses for GraphQL are just different.

dangodev:
@sslotsky OK I’m passing through most errors to the front-end, but reporting along the way. Also added tests making sure `graphqlFetch` is logging to the error reporting in a number of different ways. WDYT?

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,95 @@ interface CreateGraphqlFetch {
...
+        if (Array.isArray(message)) {
+          errors = message.map(err => ({ message: err })) as GraphqlError[];
+        } else {
+          errors = [{ message }] as GraphqlError[];
+        }
+      } else {
+        errors = [{ message: response.statusText }] as GraphqlError[];
+      }
+
+      report(errors);
```
dangodev: This will just always report errors in the GraphQL format: `[{ message: 'error1' }, { message: 'error2' }]`

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,95 @@ interface CreateGraphqlFetch {
...
+      report(e);
+      return Promise.reject(e);
     });
-  }
 
-  if (body.errors && body.errors[0].type === 'unauthorized') {
-    // TODO trigger token refresh for manifold-auth-token
-    setAuthToken('');
-  }
+    const body: GraphqlResponseBody<T> = await response.json();
```
dangodev: Should we `try / catch` around `response.json()`?

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,95 @@ interface CreateGraphqlFetch {
...
+
+    // handle unauthenticated error
+    if (response.status === 401) {
+      // TODO trigger token refresh for manifold-auth-token
+      setAuthToken('');
```
sslotsky: This needs to be reported too.

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,95 @@ interface CreateGraphqlFetch {
...
+    }
 
-  return body;
-};
+    // handle non-GQL responses from errors
+    if (!body.data && !Array.isArray(body.errors)) {
```
sslotsky: I'm really confused as to what case this is handling. Why would we expect to get a non-GQL repsonse from the GQL server unless it was some kind of `5xx` which is already handled?

src/utils/graphqlFetch.ts
sslotsky: Hmm. Well `response.json()` should for sure happen after the `401` check or any other exceptional cases. 

Once it's in the right spot, I wouldn't expect it to fail, but wrapping it probably couldn't hurt either.

src/utils/graphqlFetch.ts
dangodev: Hm. I was just wondering if our GraphQL endpoint would err at some level above, and return an unknown error message type.

I poked at our GraphQL endpoint, and even gave it weird stuff like the wrong encoding type, and it seemed to pretty consistently always return the same `{data: null, errors: […]}` format. Maybe this is overkill.

sslotsky: If it had some unexpected error it should just `500`, which we already have covered.

src/utils/graphqlFetch.ts
dangodev: That happens down below in `report(body.errors)`. Just tested an authenticated GraphQL response and it returns in that array just like everything else:

```
{"errors":[{"message":"User does not have permission to access the resource","path":["resources"]}],"data":{"resources":null}}
```

src/utils/graphqlFetch.ts
dangodev: Yeah you’re right. Removed this whole block—I was handling Swagger errors which don’t make sense here.

src/utils/graphqlFetch.ts
dangodev: `401` is returned by the GraphQL server; I guess I was just worried about the server blowing up. But I guess at that point we can’t stop our web components from doing the same, either 😅

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,86 @@ interface CreateGraphqlFetch {
...
+    if (response.status === 401) {
+      // TODO trigger token refresh for manifold-auth-token
+      setAuthToken('');
+      report(response); // report unauthenticated
+    }
 
-  return body;
-};
+    // handle non-GQL responses from errors
+    if (!body.data && !Array.isArray(body.errors)) {
```
sslotsky: I thought this whole block would be removed. What's the use case where we'd have a non-GQL error that isn't already handled by the `fetch().catch()`?

dangodev: Well, in case there’s a `500`, which wouldn’t be handled by the `catch()` above (that only throws if there’s a network error). From MDN:

> The Promise returned from `fetch()` **won’t reject on HTTP error status** even if the response is an HTTP 404 or 500

dangodev: I don’t know how to simulate a `500` on our GraphQL server so I wasn’t sure how it’d respond

sslotsky: Well shit. I thought 500 errors would throw but the documentation is pretty clear! We should probably try to figure out what will actually happen then, whether by talking to some backend folks or just reviewing the code. 

Having this code in there won't hurt anything, but ideally we don't have code (and tests) to handle things that won't happen. It's fine for now though.

src/utils/restFetch.ts
```diff
@@ -64,18 +63,18 @@ export function createRestFetch({
...
     }
 
+    const body = await response.json();
+    if (response.status >= 200 && response.status < 300) {
+      return body;
+    }
+
     // Sometimes messages are an array, sometimes they aren’t. Different strokes!
     report(response);
     const message = Array.isArray(body) ? body[0].message : body.message;
```
sslotsky: FYI https://app.zenhub.com/workspaces/engineering-56990653b8516b1b4867e8b2/issues/manifoldco/engineering/9027

src/utils/graphqlFetch.ts
```diff
@@ -7,72 +8,86 @@ interface CreateGraphqlFetch {
...
+        report(detail);
+        throw new Error(detail.message);
+      }
+    }
+
+    const response = await fetch(endpoint, {
+      method: 'POST',
+      headers: {
+        'content-type': 'application/json',
+        ...(getAuthToken() ? { authorization: `Bearer ${getAuthToken()}` } : {}),
```
sslotsky: This looks right to me, but in Storybook, I'm getting an error from the product logo query. For some reason it's sending this:

```
authorization: Bearer undefined
```

That results in this GraphQL error:

```
{"errors":[{"message":"unauthorized: Could not exchange token","path":["product"]}],"data":{"product":null}}
```

dangodev: Thanks for checking! I think that’s been broken for a while already, but forgot to investigate why. I think the auth token decorator was added. I’ll just remove it

dangodev: Yeah that was another component that was broken with all the authToken work, it seems. Adding `isPublic` to that query fixed it.

dangodev: Hey remember that time we built a system that only requested auth when a request was denied? That was awesome. Wish we could do something like that for web components, rather than failing before fetching.

sslotsky: > Hey remember that time we built a system that only requested auth when a request was denied?

I'm actually not sure what you mean here?

dangodev: The Sphericon login system—I was wondering if the web token auth could take a similar approach

sslotsky: Oh! I'm not sure how that system worked. I _think_ you're talking about reacting to 401 errors, which is something we still need to do for UI, but is also kind of independent from how we're handling our tokens. Hit me up if you want to have a chat about it.

PR #379, opened 2019-08-15
Remove fallback Stencil cache

dangodev:
## Reason for change
Sometimes an old build of Stencil would be cached, which is problematic for releases. This undoes that

## Testing
CircleCI should pass; we’ll know it worked on publish


.circleci/config.yml
```diff
@@ -29,7 +29,6 @@ commands:
       - restore_cache:
           keys:
             - npmcache-{{ checksum "package.json" }}
-            - npmcache-
```
dangodev: You know what? Forget the npm cache too while we’re at it—don’t want old versions of Shadowcat hanging around, either

sslotsky:
Hmm. I published an RC for this, and the code in the RC appears to be correct, but it's packaged differently which seems to break everything.

sslotsky:
```
sam@sam-ThinkPad-T480s:~/dashboard-render/dashboard$ ls node_modules/@manifoldco/ui/dist/
manifold  manifold.js
```

PR #380, opened 2019-08-15
Poll until resource rename complete

markjamieson:
## Reason for change

Polls resource endpoint until renamed resource is returned, then emits success event.

## Testing

New to Render work, so there's possibly a better way, but I...
1. Built UI locally
1. Copied `dist/` to `<render-dashboard>/node_modules/@manifoldco/ui/`
1. Provisioned a resource
1. Renamed the resource

### DIFF
--- a/src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
+++ b/src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
@@ -130,7 +130,7 @@ export class ManifoldDataRenameButton {
     const { restFetch } = this;
     await new Promise(resolve => {
       const interval = window.setInterval(async () => {
-        const renamedResource: [Marketplace.Resource] | undefined = await restFetch({
+        const renamedResource = await restFetch<Marketplace.Resource[]>({
           service: 'marketplace',
           endpoint: `/resources/?me&label=${this.newLabel}`,
         });
--- a/src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
+++ b/src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
@@ -126,6 +126,21 @@ export class ManifoldDataRenameButton {
       return Promise.reject(error);
     });
 
+    // Poll until rename complete
+    const { restFetch } = this;
+    await new Promise(resolve => {
+      const interval = window.setInterval(async () => {
+        const renamedResource: [Marketplace.Resource] | undefined = await restFetch({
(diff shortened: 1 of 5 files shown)


src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -126,6 +126,21 @@ export class ManifoldDataRenameButton {
       return Promise.reject(error);
     });
 
+    // Poll until rename complete
+    const restFetch = this.restFetch;
```
markjamieson: Does anybody know why I needed to do this? TS compiler did not like me trying to use `this.restFetch` inside this new Promise

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -126,6 +126,21 @@ export class ManifoldDataRenameButton {
       return Promise.reject(error);
     });
 
+    // Poll until rename complete
+    const { restFetch } = this;
+    await new Promise(resolve => {
+      const interval = window.setInterval(async () => {
+        const renamedResource: [Marketplace.Resource] | undefined = await restFetch({
```
dangodev: Do you need a type: `this.restFetch<Marketplace.Resource>` ?

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -126,6 +126,21 @@ export class ManifoldDataRenameButton {
       return Promise.reject(error);
     });
 
+    // Poll until rename complete
+    const { restFetch } = this;
```
markjamieson: I'm not really sure why this was necessary but TS didn't like me using `this.restFetch` inside this new Promise

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
markjamieson: Yes, just fixed that up a little bit

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
dangodev: Wonder if it’s a context thing 🤔

dangodev: Though seems like your solution still points to the same reference; it just hides the warning

PR #382, opened 2019-08-15
Fixed: improved loading state for <manifold-resource-list>

dangodev:
## Reason for change
Improves the animation & loading state of the resource list component. Handles the loading state internally rather than burdening the user with that.

resource-loading

## Testing
Go to the **Storybook** preview link, click on **Resource List 🔒**. Inspect the components there (may have to set `localStorageSet('manifold_api_token', 'TOKEN')`)

### DIFF
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
@@ -1,9 +1,9 @@
 export const skeleton = async () => {
   const card = document.createElement('manifold-resource-card-view');
-  card.label = 'my-resource';
+  card.label = 'loading';
   card.loading = true;
   card.logo = 'https://cdn.manifold.co/providers/logdna/logos/ftzzxwdr0c8wx6gh0ntf83fq4w.png';
-  card.name = 'my-resource';
+  card.name = 'loading';
 
   document.body.appendChild(card);
 
@@ -12,9 +12,11 @@ export const skeleton = async () => {
 
 export const logDNA = async () => {
   const card = document.createElement('manifold-resource-card-view');
-  card.label = 'my-resource';
+  card.label = 'custom-resource';
   card.logo = 'https://cdn.manifold.co/providers/logdna/logos/ftzzxwdr0c8wx6gh0ntf83fq4w.png';
-  card.name = 'my-resource';
+  card.name = 'logging';
+  card.resourceId = 'test';
+  card.resourceStatus = 'available';
 
   document.body.appendChild(card);
 
@@ -23,8 +25,10 @@ export const logDNA = async () => {
 
 export const logoPlaceholder = async () => {
   const card = document.createElement('manifold-resource-card-view');
-  card.label = 'my-resource';
-  card.name = 'my-resource';
+  card.label = 'custom-resource';
+  card.name = 'custom-resource';
+  card.resourceId = 'test';
+  card.resourceStatus = 'available';
 
   document.body.appendChild(card);
 
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
@@ -1,0 +1,32 @@
+export const skeleton = async () => {
+  const card = document.createElement('manifold-resource-card-view');
+  card.label = 'my-resource';
+  card.loading = true;
+  card.logo = 'https://cdn.manifold.co/providers/logdna/logos/ftzzxwdr0c8wx6gh0ntf83fq4w.png';
+  card.name = 'my-resource';
+
+  document.body.appendChild(card);
+
+  return card.componentOnReady();
+};
+
+export const loaded = async () => {
+  const card = document.createElement('manifold-resource-card-view');
+  card.label = 'my-resource';
+  card.logo = 'https://cdn.manifold.co/providers/logdna/logos/ftzzxwdr0c8wx6gh0ntf83fq4w.png';
+  card.name = 'my-resource';
+
+  document.body.appendChild(card);
+
+  return card.componentOnReady();
+};
+
+export const noLogo = async () => {
+  const card = document.createElement('manifold-resource-card-view');
+  card.label = 'my-resource';
+  card.name = 'my-resource';
+
+  document.body.appendChild(card);
+
+  return card.componentOnReady();
+};
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
@@ -10,7 +10,7 @@ export const skeleton = async () => {
   return card.componentOnReady();
 };
 
-export const loaded = async () => {
+export const logDNA = async () => {
   const card = document.createElement('manifold-resource-card-view');
   card.label = 'my-resource';
   card.logo = 'https://cdn.manifold.co/providers/logdna/logos/ftzzxwdr0c8wx6gh0ntf83fq4w.png';
@@ -21,7 +21,7 @@ export const loaded = async () => {
   return card.componentOnReady();
 };
 
-export const noLogo = async () => {
+export const logoPlaceholder = async () => {
   const card = document.createElement('manifold-resource-card-view');
   card.label = 'my-resource';
   card.name = 'my-resource';
--- a/src/components/manifold-resource-card-view/manifold-resource-card-view.css
+++ b/src/components/manifold-resource-card-view/manifold-resource-card-view.css
@@ -3,9 +3,9 @@
 @custom-media --viewport-l (min-width: 1080px);
 
 :host {
-  --manifold-product-logo-s: 0.5rem;
-  --manifold-product-logo-m: 1.25rem;
-  --manifold-product-logo-l: 1.5rem;
+  --manifold-product-logo-s: 1.5rem;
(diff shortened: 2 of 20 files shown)


src/components/manifold-resource-card-view/manifold-resource-card-view-happo.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const noLogo = async () => {
+  const card = document.createElement('manifold-resource-card-view');
+  card.label = 'my-resource';
+  card.name = 'my-resource';
+
+  document.body.appendChild(card);
+
+  return card.componentOnReady();
+};
```
dangodev: Added resource card view VRT tests

src/components/manifold-resource-card-view/manifold-resource-card-view.css
```diff
@@ -199,3 +89,47 @@
...
+
+  @media (--viewport-l) {
+    padding-left: 0;
+  }
+}
+
+.status {
+  grid-area: status;
+  padding-top: 0.25rem;
+}
```
dangodev: Fixes responsive bugs

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
```diff
@@ -135,15 +96,17 @@ export class ManifoldResourceCardView {
...
           {this.name || this.label}
         </h3>
-        <div class="status-box">
-          <div class="status" data-status={this.status}>
-            <div class="inner" itemprop="status">
-              {this.statusText}
-            </div>
-          </div>
+        <div class="status">
+          <manifold-resource-status-view size="xsmall" resourceState={this.resourceStatus} />
```
dangodev: We were duplicating quite a bit of status logic, checks, and styles across components. This consolidates all that into `<manifold-resource-status-view>`.

src/components/manifold-resource-card-view/manifold-resource-card-view.e2e.ts
```diff
@@ -1,45 +1,27 @@
...
-    expect(src).toBe(logo);
-  });
-
-  it('displays the given status', async () => {
-    const page = await newE2EPage({
-      html: `<manifold-resource-card-view resource-status="provision" resource-id="test" />`,
-    });
-    const el = await page.find('manifold-resource-card-view >>> [itemprop="status"]');
-
-    expect(el.innerText).toBe('Provisioning');
```
dangodev: Some tests were removed here, but replaced with Happo VRT tests. Coverage “went down,” but not really because these were all visual tests anyway.

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
sslotsky: Yeah, a lot happened in this component that shouldn't have happened. For example, it fetches. `-view` components aren't supposed to, but we (i.e. I) never documented that. At a glance it seems like this could just be a descendant of `manifold-resource-container` so it wouldn't have to worry about fetching.

sslotsky: Not that I'm saying you should change that in this PR but we ought to follow up.

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
```diff
@@ -135,15 +96,17 @@ export class ManifoldResourceCardView {
...
         </div>
         <div class="logo">
-          {this.logo && <manifold-lazy-image src={this.logo} alt={this.label} itemprop="image" />}
+          {this.logo ? (
+            <manifold-lazy-image src={this.logo} alt={this.label} itemprop="image" />
```
sslotsky: I've been meaning to ask if we ever gave more thought into using the native lazy images and getting rid of the intersection observer stuff.

src/components/manifold-resource-card/manifold-resource-card.tsx
```diff
@@ -18,51 +18,42 @@ export class ManifoldResourceCard {
...
+    this.fetchResource({ label: newlabel });
   }
 
   componentWillLoad() {
-    if (this.resourceId) {
-      this.fetchResourceId(this.resourceId);
-    } else if (this.label) {
-      this.fetchResourceLabel(this.label);
-    }
+    this.fetchResource({ id: this.resourceId, label: this.label });
```
sslotsky: I don't know when this component changed to being able to fetch by either ID or label, but I don't like it. It makes it unclear how it should be used. Do you know what the reasoning is for it?

src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -201,22 +201,30 @@ export class ManifoldResourceList {
...
+                label={resource.label}
+                name={resource.name}
+                logo={resource.logo}
+                resourceId={resource.id}
+                resourceStatus={resource.status}
+                resourceLinkFormat={this.resourceLinkFormat}
+                preserveEvent={this.preserveEvent}
+              />
+            ))
+          : [1, 2, 3, 4].map(() => (
```
sslotsky: Wasn't someone objecting to the fact that this shows a different number of cards than what will ultimately render? Was that concern alleviated in some other way? I've been thinking recently that it might be easier to defer loading state to consumers by leveraging `componentOnReady()` in conjunction with an async `componentWillLoad()`.

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
dangodev: I’ve been watching it. It’s Chrome-only right now. You can polyfill it, but the polyfill is as much or more code than intersection observer. Maybe when another browser picks it up we’ll use it.

src/components/manifold-resource-list/manifold-resource-list.tsx
dangodev: I heard that complaint before, but this is a technique that Netlify, Now, Facebook, GitBook, … all use to create placeholders. You can’t know how many things will appear before you fetch the things. You can’t fetch things before fetching things.

The only solid solution to this IMO is pre-rendering content, or showing nothing. But personally I’d rather show placeholders while content is loaded, and add a reasonable number close to what the average count will be.

dangodev: Also part of the problem was we were showing “loading“ text alongside placeholders, which breaks the illusion. When you have readable text, that elevates beyond placeholder into an actual resource that is being loaded. Hopefully now it’ll be perceived as a placeholder and nothing more, because it’s now purely abstract and not a blend of abstract + concrete.

src/components/manifold-resource-card/manifold-resource-card.tsx
dangodev: To me it’s clearer—we weren’t fetching resource IDs or labels; we’re fetching resources. We were also duplicating logic to do the same thing. When we move this to GraphQL, too, I think it’ll be combined anyway.

If we want to make TypeScript a little more robust, I can change the param to require either the `id` or `label` with a discriminated union type—maybe that’d help?

src/components/manifold-resource-card-view/manifold-resource-card-view.tsx
dangodev: Yeah I saw that fetching in the `view` but don’t want to change that just yet. Maybe later.

src/components/manifold-resource-card/manifold-resource-card.tsx
sslotsky: What I mean is that it's unclear from the perspective of someone who's going to drop the component on a page. Should I give it an ID or a label? The component allows me to supply both. What if the ID and label aren't for the same resource? It's not clear to me that the component will just ignore the label and use the ID.

And when I ask what the reasoning is for it, what I mean is, what's the use case? It used to only support labels, so if we updated it to support IDs, was there some use case that required us to fetch by ID instead? Why do we support it?

dangodev: Oh I see. Yeah I’m not sure there. It smells like there wasn’t long-term thinking there. I’m with you—flexibility is the enemy, and there should be a specific use-case.

Good news! I don’t think we’re using `<manifold-resource-card>` like… at all. Neither in UI nor Render. I think now is a great time to cut scope somewhere

sslotsky: Suh-weeeet! Would that be part of this PR or a follow-up?

src/components/manifold-resource-list/manifold-resource-list.tsx
sslotsky: Cool. I still plan to explore leveraging `componentOnReady()` and `async componentWillLoad()` in order to defer loading state to consumers. I think that is ultimately a good way to go. But this is fine with me as long as it's fine with Nick.

FWIW I'm anti skeleton screen. I feel like it's not fooling anyone. But that's just me :smile:

src/components/manifold-resource-card/manifold-resource-card.tsx
dangodev: If you could double-check with me it’s not being used anywhere (UI, or Render), I‘d like to remove the ID altogether from this component from fetching.

And if you could also confirm `<manifold-resource-card-view>` isn’t being used in Render at all (other than the `loading` state, which will go away), I can killl the fetching there, too.

sslotsky: Confirmed, they're not being used! That's a relief.

PR #390, opened 2019-08-16
Fixed: Flash of 'No services found' on marketplace grid

sslotsky:
## Reason for change

The marketplace grid was flashing the "No services found" message instead of displaying the skeleton. This tightens up the logic and makes sure that the `filteredServices` function returns the skeleton services when they're provided.

## Testing

Test this in dashboard-render and make sure you see the skeleton when loading the marketplace, and that the "No services found" message does not flash at any point. Also make sure to check Storybook and the docs site.

### DIFF
--- a/docs/docs/components/manifold-marketplace.md
+++ b/docs/docs/components/manifold-marketplace.md
@@ -14,38 +14,23 @@ A list of all Manifold services.
 <manifold-marketplace></manifold-marketplace>
 ```
 
-## Catalog curation
+## Catalog Curation
 
-Our marketplace offers 2 options, either to show all services excluding a
-few, or manually specifying the products that should be shown. For either
-option there is either the `excludes` or `products` options—use either, but
-not both in the same instance.
-
-#### Excludes
-
-Excludes will start with **all products**, but let you specify which ones
-you’d like to hide:
-
-```html
-<manifold-marketplace excludes="manifold-provider"></manifold-marketplace>
-```
-
-#### Products
-
-Conversely, manually specifying products will start with **no products**, and
-let you manually specify every product that should be shown.
+Specifying products will start with **no products**, and let you manually specify every product that
+should be shown.
 
 A common usecase is using `products` in conjuction with `hide-categories` (below).
 
 ```html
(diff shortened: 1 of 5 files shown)


src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -90,34 +89,17 @@ export class ManifoldMarketplaceGrid {
...
 
-    // Handle excludes
-    if (Array.isArray(this.excludes)) {
-      services = services.filter(
-        ({ body: { label } }) => this.excludes && !this.excludes.includes(label)
-      );
-    }
+    const services = this.products.length
+      ? this.services.filter(s => this.products.includes(s.body.label))
+      : this.services;
```
dangodev: Minor note: this reverses the ability to specify order of products, I think. Do we want to keep that?

src/components.d.ts
```diff
@@ -331,16 +327,15 @@ export namespace Components {
     'templateLinkFormat'?: string;
   }
   interface ManifoldMarketplaceGrid {
-    'excludes'?: string[];
```
dangodev: If this is being removed, docs should be updated

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
sslotsky: Ah, I didn't think of that. Was that explicitly requested? I can restore it if it matters.

docs/docs/components/manifold-marketplace.md
```diff
@@ -14,38 +14,23 @@ A list of all Manifold services.
...
+></manifold-marketplace>
 ```
 
 ## Hide categories
 
-Hiding the categories is a great way to have a more compact display. It’s
-recommended to use this if you’re only displaying a few products via
-`products` (above):
+Hiding the categories is a great way to have a more compact display. It’s recommended to use this if
+you’re only displaying a few products via `products` (above):
```
dangodev: Why the changes here? Different prettier version?

sslotsky: Probably the first time the file was saved since `proseWrap` was introduced.

PR #394, opened 2019-08-16
Add PR template checklist

dangodev:
## Reason for change
Adds a few checklist items to PR templates, specific to UI

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,8 +5,20 @@
 <!-- ✅ Linked to ZenHub issue               -->
 
 ## Reason for change
+
 <!-- What does this change, in plain language? -->
 <!-- Before/after screenshots may be helpful.  -->
 
 ## Testing
+
 <!-- For someone unfamiliar with the issue, how should this be tested? -->
+
+## Checklist
+
+<!-- are all the steps completed? -->
+
+- [ ] **CHANGELOG**: The **Unreleased** section of CHANGELOG was updated (and `npm run copy-changelog` was run)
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -5,8 +5,20 @@
...
 
 ## Testing
+
 <!-- For someone unfamiliar with the issue, how should this be tested? -->
+
+## Checklist
+
+<!-- are all the steps completed? -->
+
+- [ ] **CHANGELOG**: The **Unreleased** section of CHANGELOG was updated (and `npm run copy-changelog` was run)
```
batgirl: Wait hmm I don't see `npm run copy-changelog` in the docs re: the Changelog, I'm not sure when or why I would need to run it. Can that maybe be added to the README? And we're only making changes in the main Changelog, right? (For a while there were 2 of them.)

batgirl: I now see how my two questions are related 😄

dangodev: That’s a good question! Yeah we have to have 2, unfortunately, because of the way Gatsby works. It doesn’t like importing things above its working directory in `docs/`. But we also need a `CHANGELOG` in the root for things like dependabot, and standards, and stuff.

We were managing 2, but I made that script to copy the root one to `docs/`. Running it just updates Git, but that script gets run so if someone forgot to copy their changes, the one in root would be deployed.

Honestly I should have just stopped tracking the other one in Git; after explaining it all it’s apparent it’s bad. I’ll make a PR for that.

PR #396, opened 2019-08-16
Emit metrics events from restFetch and graphqlFetch

markjamieson:
## Reason for change

This PR emits events to record duration of graphql and REST fetch calls. Event is emitted from optional EventEmitter and falls back to emitting from `document`.

## Testing

I tested by adding an event listener to a render-dashboard page and using this version of UI in render-dashboard. I had trouble testing this in storybook but there might be a way.

### DIFF
--- a/docs/docs/advanced/performance.md
+++ b/docs/docs/advanced/performance.md
@@ -64,3 +64,12 @@ Stencil’s web components work for more browsers than even have web component
 support [via a polyfill][stencil-browsers]. It’s great to have built-in
 support, but sometimes webpack can accidentally bundle the polyfill when it
 doesn’t need to. We’ve found the best fix to be **code splitting** (above).
+
+## Metrics
+
+RTT of endpoint calls are calculated and emitted as events - either from
+`document` or from an `EventEmitter` supplied by the component calling the
+endpoint.
+
+A solution to listen for and log these events is currently left as an
+implementation detail on the platform side.
--- a/docs/docs/advanced/performance.md
+++ b/docs/docs/advanced/performance.md
@@ -39,6 +39,7 @@ big payoffs.
 [fmp]: https://developers.google.com/web/tools/lighthouse/audits/first-meaningful-paint
 [import]: https://webpack.js.org/guides/code-splitting/#dynamic-imports
 [stencil-browsers]: https://stenciljs.com/docs/browser-support
+[rtt]: https://developer.mozilla.org/en-US/docs/Glossary/Round_Trip_Time_(RTT)
 
 ## Other Gotchas
 
@@ -67,9 +68,15 @@ doesn’t need to. We’ve found the best fix to be **code splitting** (above).
 
 ## Metrics
 
-RTT of endpoint calls are calculated and emitted as events - either from
+[RTT][rtt] of endpoint calls are calculated and emitted as events - either from
 `document` or from an `EventEmitter` supplied by the component calling the
 endpoint.
 
 A solution to listen for and log these events is currently left as an
-implementation detail on the platform side.
+implementation detail on the platform side:
+
+```js
+document.addEventListener('manifold-rest-fetch-duration', { detail } => {
+  console.log(detail); // { endpoint: "/catalog/products", duration: 95 }
+});
+```
--- a/src/utils/graphqlFetch.spec.ts
+++ b/src/utils/graphqlFetch.spec.ts
@@ -269,12 +269,12 @@ describe('graphqlFetch', () => {
       });
 
       let event: CustomEvent | undefined;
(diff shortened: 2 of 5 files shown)


src/utils/graphqlFetch.ts
```diff
@@ -87,6 +88,21 @@ export function createGraphqlFetch({
       report(body.errors);
     }
 
+    const fetchDuration = new Date().getTime() - start.getTime();
```
dangodev: Does this work better than `performance.now()`?

src/utils/graphqlFetch.ts
```diff
@@ -87,6 +88,21 @@ export function createGraphqlFetch({
...
+        request,
+        duration: fetchDuration,
+      });
+    } else {
+      document.dispatchEvent(
+        new CustomEvent('graphql-fetch-duration', {
+          detail: { request, duration: fetchDuration },
+        })
+      );
+    }
```
dangodev: Nice! I like this flexibility. And I like that this doesn’t block the graphql request at all (we’re not `await`-ing anything)

dangodev:
@markjamieson are you able to add a test for these? Since these are just functions you should be able to get away with just Jest tests. You should be able to pass in a mock function for the emitter & test it gets called.

Also adding a section about this to the “Performance” markdown doc would be helpful too

src/utils/graphqlFetch.ts
dangodev: Sorry; I should have phrased that better. `performance.now()` is meant for this kind of performance benchmarking because it’s much more accurate. I’d prefer we use that here, but I did ask because I’m open to arguments if `Date()` is better in this scenario.

markjamieson: No, I just reached for something familiar and that happened to be `Date()`. I had looked at `performance` for `mark` and `measure` which I ultimately rejected, but I didn't realize there was a simple `now` function I can use here. Definitely a change I'll make.

docs/docs/advanced/performance.md
```diff
@@ -64,3 +64,12 @@ Stencil’s web components work for more browsers than even have web component
...
 doesn’t need to. We’ve found the best fix to be **code splitting** (above).
+
+## Metrics
+
+RTT of endpoint calls are calculated and emitted as events - either from
+`document` or from an `EventEmitter` supplied by the component calling the
+endpoint.
+
+A solution to listen for and log these events is currently left as an
+implementation detail on the platform side.
```
dangodev: Suggestions:
- for `RTT` can you link to an explanation (maybe this? https://developer.mozilla.org/en-US/docs/Glossary/Round_Trip_Time_(RTT)).
- maybe include some code snippets here to show how logging works, for them to copy + paste (e.g. “Events” here: https://ui.sandbox.manifold.co/components/manifold-marketplace)

src/utils/graphqlFetch.spec.ts
```diff
@@ -251,4 +252,53 @@ describe('graphqlFetch', () => {
...
+        body,
+      });
+
+      let event: CustomEvent | undefined;
+      addEventListener('graphql-fetch-duration', e => {
+        event = e as CustomEvent;
+      });
+
+      await fetcher({ query: '' });
+      event && event.detail && expect(event.detail.duration).toBeDefined;
```
dangodev: Let’s switch this around to have `event && event.detail` inside the `expect()`. Otherwise it won’t fire (if you conditionally bypass `expect`, the test will pass, unless you do `expect.assertions(1)`).

src/utils/graphqlFetch.spec.ts
```diff
@@ -251,4 +252,53 @@ describe('graphqlFetch', () => {
...
+        body,
+      });
+
+      let emitter: EventEmitter = {
+        emit: jest.fn(),
+      };
+
+      await fetcher({ query: '', emitter });
+      expect((emitter.emit as jest.Mock).mock.calls[0][0]).toMatchObject({
+        type: 'graphql-fetch-duration',
```
dangodev: Is there anything else we can expect in this call (event details or something?)

markjamieson: We can expect a duration but I don't know exactly what the duration will be in the test. I can test for its existence though.

dangodev: Oh good point. 👍

docs/docs/advanced/performance.md
markjamieson: Done

src/utils/graphqlFetch.spec.ts
markjamieson: Done

PR #401, opened 2019-08-20
Remove extra event emitters, add DOM tests

dangodev:
## Reason for change

We were firing the same events in certain instances from both parent and child, which was resulting in multiple components firing, say, a `success` event twice. Normally that wouldn’t be a huge problem, but when the parent fires it doesn’t have any of the data it needs.

That’s actually a huge problem, because if you have a `success` event, and you tell your application to redirect to `/addon/[resourceName]`, it’s going to redirect to `/addon/undefined` and other fun things like that.

Also, tests were rewritten to test the DOM + event name, rather than testing implementation details on the class. New tests were added when events weren’t being tested.

## Testing

Overall this is a fairly safe PR to merge as it only removes event emitting in 3 spots. It’s hard to test, but tests should pass. 90% of the changes were making the tests stronger by testing DOM clicks on buttons & waiting for events to fire, rather than testing class methods.

I did test this PR in Render Dashboard and it solves the intended problem of events firing twice.

## Checklist

### DIFF
--- a/docs/docs/data/manifold-data-rename-button.md
+++ b/docs/docs/data/manifold-data-rename-button.md
@@ -21,19 +21,23 @@ Set the CTA text by adding anything between the opening and closing tags:
 </manifold-data-provision-button>
 ```
 
-`slot` can be attached to any HTML or JSX element. To read more about slots, see [Stencil’s Documentation][stencil-slot]
+`slot` can be attached to any HTML or JSX element. To read more about slots, see [Stencil’s
+Documentation][stencil-slot]
 
 ## Events
 
 For validation, error, and success messages, it will emit custom events.
 
 ```js
-document.addEventListener('manifold-renameButton-click', ({ detail: { resourceLabel, newLabel } }) =>
-  console.info(`⌛ Renaming ${resourceLabel} to ${newLabel} …`)
+document.addEventListener(
+  'manifold-renameButton-click',
+  ({ detail: { resourceLabel, newLabel } }) =>
+    console.info(`⌛ Renaming ${resourceLabel} to ${newLabel} …`)
 );
 document.addEventListener(
   'manifold-renameButton-success',
-  ({ detail: { resourceLabel, newLabel } }) => alert(`${resourceLabel} renamed to ${newLabel} successfully!`)
+  ({ detail: { resourceLabel, newLabel } }) =>
+    alert(`${resourceLabel} renamed to ${newLabel} successfully!`)
 );
 document.addEventListener('manifold-renameButton-error', ({ detail }) => console.log(detail));
 // {message: "bad_request: bad_request: No plan_id provided", resourceid: "1234", resourceLabel: "my-resource", newLabel: "new-name"}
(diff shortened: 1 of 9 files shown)


src/components/manifold-resource-deprovision/manifold-resource-deprovision.tsx
```diff
@@ -19,9 +16,6 @@ export class ManifoldResourceDeprovision {
         resourceId={this.data && this.data.id}
         resourceLabel={this.data && this.data.label}
         loading={this.loading}
-        onManifold-deprovisionButton-click={this.click.emit}
-        onManifold-deprovisionButton-error={this.error.emit}
-        onManifold-deprovisionButton-success={this.success.emit}
```
dangodev: This was the main cause of the problem—we don’t need to force-emit child events from the parent.

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
```diff
@@ -168,6 +149,29 @@ export class ManifoldDataRenameButton {
...
+            // wait till interval has passed at least to continue
+            const diff = Math.round(performance.now() - start - pollInterval);
+            setTimeout(() => this.pollRename(), Math.max(diff, 0));
+          }
+        });
+      }
+      return reject();
+    });
+  }
+
```
dangodev: @markjamieson I was encountering some `this` problems in your PR in the test suite. I think it was unique to the test suite because it worked flawlessly in Render’s Dashboard, but I pulled it out to its own method also for clarity.

I made a minor change where I changed the interval to a timeout that only fires if it’s not done, because I imagined an instance where these would start to stack up if they didn’t resolve within `100ms`. In this way the calls don’t happen at perfect intervals anymore, but they won’t stack up.

I also dialed it down to `300` in case marketplace is slow, but that was based more on past experience than testing this exact call.

Anyways both were gut changes that didn’t need to be made, but wanted to get your thoughts on them. I’m 100% OK reverting any or all of it if there’s thinking I missed.

src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
```diff
@@ -120,120 +149,92 @@ describe('<manifold-data-sso-button>', () => {
...
 
-      element.addEventListener('manifold-ssoButton-error', onError);
-      const button = element.querySelector('button') as HTMLElement;
-      expect(button).toBeDefined();
-
-      button.click();
-      await page.waitForChanges();
-      // TODO: wait for changes does not wait for all events to finish emitting.... WHYYYYYYY
-      // TODO: Wait on this: https://github.com/ionic-team/stencil/issues/1781
-      await new Promise(resolve => setTimeout(resolve, 1000));
```
dangodev: I was able to take this test that took `1s` to execute each time and got it down to ~`9ms` by just waiting for the event to fire rather than a timeout. Also did that in other places as well.

src/components/manifold-resource-rename/manifold-resource-rename.tsx
```diff
@@ -23,14 +19,8 @@ export class ManifoldResourceRename {
...
         loading={this.loading}
         newLabel={this.newLabel}
-        onManifold-renameButton-click={this.click.emit}
-        onManifold-renameButton-invalid={this.invalid.emit}
-        onManifold-renameButton-error={this.error.emit}
-        onManifold-renameButton-success={this.success.emit}
       >
-        <manifold-forward-slot>
-          <slot />
-        </manifold-forward-slot>
```
dangodev: Also removed the forward slot—that’s only needed for named slots, not default ones.

docs/docs/data/manifold-data-rename-button.md
```diff
@@ -21,40 +21,43 @@ Set the CTA text by adding anything between the opening and closing tags:
...
+| :------------------------------ | :--------------------------------------------------: | :-------------------------------------------------------------------------------------------------------------------------- |
+| `manifold-renameButton-click`   |      `resourceId`, `resourceLabel`, `newLabel`       | Fires immediately when button is clicked. May be used to trigger a loading state, until `-success` or `-error` is received. |
+| `manifold-renameButton-success` | `message`, `resourceId`, `resourceLabel`, `newLabel` | Successful renaming. Returns a resource ID and resource Label as well as a message and the new name for the resource.       |
+| `manifold-renameButton-error`   | `message`, `resourceId`, `resourceLabel`, `newLabel` | Erred rename, along with information on what went wrong.                                                                    |
+| `manifold-renameButton-invalid` | `message`, `resourceId`, `resourceLabel`, `newLabel` | Invalid renaming, along with information on what went wrong.                                                                |
```
dangodev: Fixed typo in docs. The other changes were just Prettier reformatting.

src/components/manifold-data-rename-button/manifold-data-rename-button.spec.ts
```diff
@@ -120,135 +117,209 @@ describe('<manifold-data-rename-button>', () => {
...
       );
-      expect(instance.success.emit).toHaveBeenCalledWith({
-        message: `${Resource.body.label} successfully renamed`,
-        resourceLabel: Resource.body.label,
-        newLabel: 'test2',
-        resourceId: Resource.id,
+    });
+  });
+
+  describe('events', () => {
```
dangodev: Updated the rename events to test on DOM click rather than testing on the class instance.

src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
```diff
@@ -96,15 +94,46 @@ describe('<manifold-data-sso-button>', () => {
...
+      expect(element.querySelector('button')).toBeDefined();
+      // @ts-ignore
+      element.querySelector('button').click();
+      await page.waitForChanges();
+
+      expect(fetchMock.called(`${connections.prod.connector}/sso`)).toBe(false);
     });
+  });
+
+  describe('events', () => {
```
dangodev: Updated the sso events to test on DOM click rather than testing on the class instance.

src/components/manifold-data-rename-button/manifold-data-rename-button.tsx
markjamieson: Both changes look fine to me.

markjamieson:
> if you have a success event, and you tell your application to redirect to `/addon/[resourceName]`, it’s going to redirect to `/addon/undefined` and other fun things like that.

Is this something you've seen happen? Are there reliable repro steps in Render dashboard?

dangodev:
> Is this something you've seen happen? Are there reliable repro steps in Render dashboard?

Yeah it’s something I experienced while working in the Render Dashboard. The clearest evidence I have of it is right here: https://github.com/manifoldco/dashboard-render/blob/master/dashboard/src/components/addon/Addon.js#L146

```js
                          onSuccess={({ detail }) => {
                            if (detail.redirectUrl) {
                              setIsSubmitting(false);
                              Object.assign(document.createElement("a"), {
                                target: "_blank",
                                rel: "noopener noreferrer",
                                href: detail.redirectUrl
                              }).click();
                            }
                          }}
```

If you remove that `if`, then it will fire 2 events with one having `undefined` for `detail.redirectUrl`. Basically, we’re writing code around this bug rather than fixing it in UI.

But also it should follow that events only need to be emitted from a user event (in all 3, clicking a `<button>`). IMO there’s no reason why the action should be `user clicks button` > `component fires action` > `parent re-fires same action`. This PR takes out the last step.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -143,92 +139,150 @@ describe('<manifold-data-deprovision-button>', () => {
...
+      if (!button) throw new Error('button not found in document');
+
+      // listen for event and fire
+      const mockClick = jest.fn();
+      page.doc.addEventListener('manifold-deprovisionButton-click', mockClick);
+      button.click();
+
+      expect(fetchMock.called(`${connections.prod.gateway}/id/resource/${Resource.id}`)).toBe(true);
+      expect(mockClick).toHaveBeenCalledWith(
+        expect.objectContaining({
```
markjamieson: Hey, this is something I'm glad to have learned about today. I thought `toHaveBeenCalledWith()` required an exact args match.

dangodev: @sslotsky taught me about this one recently!

PR #404, opened 2019-08-20
added as-card prop

davidleger95:
## Reason for change

- fix layout issues for https://app.zenhub.com/workspaces/engineering-56990653b8516b1b4867e8b2/issues/manifoldco/engineering/8939

Screen Shot 2019-08-20 at 11 37 14 AM

## Testing

Check that card renders properly in "Resource Product Details" stories in Storybook.

## Checklist

### DIFF
--- a/src/components/manifold-resource-product/manifold-resource-product.tsx
+++ b/src/components/manifold-resource-product/manifold-resource-product.tsx
@@ -6,6 +6,7 @@ import logger from '../../utils/logger';
 
 @Component({ tag: 'manifold-resource-product' })
 export class ManifoldResourceProduct {
+  @Prop() asCard?: boolean = false;
   @Prop() data?: Gateway.Resource;
   @Prop() loading: boolean = true;
 
@@ -18,6 +19,7 @@ export class ManifoldResourceProduct {
         name={this.data.product.name}
         productId={this.data.product.id}
         productLabel={this.data.product.label}
+        borderDisabled={!this.asCard}
       >
         <manifold-forward-slot slot="cta">
           <slot name="cta" />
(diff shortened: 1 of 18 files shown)


src/components/manifold-resource-product/manifold-resource-product.tsx
```diff
@@ -6,6 +6,7 @@ import logger from '../../utils/logger';
 
 @Component({ tag: 'manifold-resource-product' })
 export class ManifoldResourceProduct {
+  @Prop() asCard?: boolean = false;
```
davidleger95: So this was documented but not implemented (or the implementation was removed). Was it removed for a reason? I know this is a bit of a hacky way of doing things, but for now it works. We'll have to figure out something more long-term because this is flexibility we're going to want for a lot of components. 

Hopefully my new theming API approach works out and we don't have to worry about this for much longer.

dangodev: Oh shoot I removed that because it was a hack for Render. That prop didn’t seem useful to me.

Maybe we should just remove the docs & storybook test?

dangodev: In case it’s helpful to know, Render is no longer using this prop. I tried (maybe in another component) to just add border & hover styles if `[href]` is set instead. Maybe a similar approach here would help?

davidleger95: that's what is currently implemented. should I just remove this and keep the changes for the button?

dangodev: That sounds good to me. Just make sure that `<manifold-resource-list>` still is able to generate links & keeps its hover styling.

davidleger95: done

PR #407, opened 2019-08-20
Add expiration test

dangodev:
## Reason for change

Wanted to test that this expiration token was working as expected.

## Testing

Tests should pass.

## Checklist


package.json
```diff
@@ -51,7 +51,8 @@
     "prepare:package": "npm run build && rm -rf pkg && mkdir pkg && cp -r dist pkg && cp package.json pkg/package.json",
     "publish": "npm run prepare:package && node scripts/publish",
     "storybook": "start-storybook -p 6006",
-    "test": "stencil test --spec --e2e --maxWorkers=2",
+    "test": "stencil test --spec --e2e --maxWorkers=2 && npm run test:tz",
+    "test:tz": "TZ=America/New_York npm run test:spec auth.spec && TZ=America/Los_Angeles npm run test:spec auth.spec && TZ=Europe/London npm run test:spec auth.spec && TZ=Asia/Tokyo npm run test:spec auth.spec",
```
dangodev: I’m on the fence about this timezone test, but kinda leaning in favor of it. Turns out we didn’t need it, but this may save some future snafus. Timezones in JavaScript are a bit of a minefield, and though we’re using timezoneless methods now, if someone none the wiser modified `auth.ts` and used a timezone-dependent method our tests would likely still pass.

config.yml
```diff
@@ -1,15 +1,12 @@
...
-        dependency_name: 'fs-extra' # For a CI script; doesn’t matter
       - match:
         dependency_name: 'jest' # Version-managed by Stencil
+        version_requirement: '*'
       - match:
         dependency_name: 'jest-cli' # Version-managed by Stencil
+        version_requirement: '*'
       - match:
         dependency_name: 'puppeteer' # Version-managed by Stencil
+        version_requirement: '*'
```
dangodev: Changed because dependabot seems to be installing dependencies that Stencil wants to overwrite (`jest`, `jest-cli`, `pupeeteer`). Trying to add this wildcard.

package.json
```diff
@@ -106,8 +107,8 @@
     "graphql": "^14.4.2",
     "happo-plugin-typescript": "^1.0.0",
     "happo.io": "^3.21.2",
-    "jest": "24.9.0",
-    "jest-cli": "24.9.0",
+    "jest": "24.8.0",
+    "jest-cli": "24.8.0",
```
dangodev: When running tests, Stencil automatically installed an older version it needed as part of a pre-check. Updated the dependabot config to try and get dependabot to ignore these.

PR #408, opened 2019-08-21
add cdn publish

domenicrosati:
🎉 

Once we merge this we can create some test tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,8 @@ jobs:
       - setup
       - happo/run_happo
       - save-build
-    
+  
+
   # Publish packages to js.cdn.manifold.co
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -153,6 +164,15 @@ workflows:
...
             - vrt
+      - cdn-publish:
+          context: gcp-cdn-auth
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v[0-9]+\.[0-9]+\.[0-9]+-.*$/
+          requires:
+            - npm-publish
```
dangodev: LGTM! To verify it’s working correctly it should have `dist/loader/index.mjs` and `/LICENSE.md` among a bunch of other files. But those are 2 to check everything got built & pushed properly.

PR #409, opened 2019-08-21
Add add-auth attribute for manifold-marketplace

dangodev:
## Reason for change

For components that use catalog, it’s difficult to determine whether or not they need to wait for auth. Since most consumers **aren’t** using it, let’s do that by default, but allow a `use-auth` prop that gets passed in for cases they need it. 

This was based on some discussions we had, and the logic around when to wait (or not) was overly complex. But by passing the responsibility to the consumer on what to use, we can be more certain we’re handling that usecase. As a result we can simplify the auth code in UI.

## Testing

- [ ] In Storybook, `manifold-marketplace` doesn’t use auth by default
- [ ] (eventually, we’ll remove the `catalog` check in restFetch so for now the “with auth“ story doesn’t authenticate, but the plan is once all these components use this, remove the catalog-specific check)

## Checklist

### DIFF
--- a/docs/docs/components/manifold-marketplace.md
+++ b/docs/docs/components/manifold-marketplace.md
@@ -63,10 +63,8 @@ You can add a “Featured” tag to select products by specifing a comma-separat
 
 ## Events
 
-This component emits
-[custom events](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) when it
-updates. To listen to those events, add an event listener either on the component itself, or
-`document`:
+This component emits [custom events][custom-events] when it updates. To listen to those events, add
+an event listener either on the component itself, or `document`:
 
 ```js
 document.addEventListener('manifold-marketplace-click', { detail } => {
@@ -98,3 +96,15 @@ also be turned into an `<a>` tag by specifying `product-link-format` and `templa
 custom resource templates.
 
 Note that this will disable the custom events unless `preserve-event` is passed as well.
+
+## Authentication
+
+By default, `<manifold-marketplace>` is a public component. However, when used in conjunction with
+[Authentication][auth], you can request to authenticate with `with-auth`:
+
+```html
+<manifold-marketplace with-auth></manifold-marketplace>
+```
+
+[auth]: /advanced/authentication
+[custom-events]: (https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
--- a/src/components/manifold-marketplace/manifold-marketplace.tsx
+++ b/src/components/manifold-marketplace/manifold-marketplace.tsx
@@ -25,6 +25,8 @@ export class ManifoldMarketplace {
   @Prop() productLinkFormat?: string;
   /** Comma-separated list of shown products (labels) */
   @Prop() products?: string;
+  /** Use auth? */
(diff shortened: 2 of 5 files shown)


src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -25,6 +25,8 @@ export class ManifoldMarketplace {
   @Prop() productLinkFormat?: string;
   /** Comma-separated list of shown products (labels) */
   @Prop() products?: string;
+  /** Use auth? */
+  @Prop() useAuth?: boolean = false;
```
davidleger95: is there any reason why we wouldn't always request without auth?

Only thing that comes to mind controlling the products that can be accessed by a user? or maybe a platform? but platform access is going to be handled by a different header, right?

docs/docs/components/manifold-marketplace.md
```diff
@@ -98,3 +96,15 @@ also be turned into an `<a>` tag by specifying `product-link-format` and `templa
...
 
 Note that this will disable the custom events unless `preserve-event` is passed as well.
+
+## Authentication
+
+By default, `<manifold-marketplace>` is a public component. However, when used in conjunction with
+[Authentication][auth], you can request to authenticate with `with-auth`:
+
+```html
+<manifold-marketplace with-auth></manifold-marketplace>
```
davidleger95: is `with-auth` the naming we want to use? Personally that naming I associate with wrapping a component in a function that injects props. like I'd expect: `withAuth(ManifoldMarketplace)`

alternative suggestion: `auth-required`

src/components/manifold-marketplace/manifold-marketplace.tsx
sslotsky: As I understand it, platforms may want to see certain hidden products and plans. For that, wouldn't they need a platform token?

dangodev: You do have a point—at the moment I’m not aware of any actual need for platforms to access hidden products. You’re right in that they will have a specific list curated for them, but none of it uses auth.

I think only times platforms would need auth would be in the provider role, when they’re a platform and they are also releasing services. But to Dave’s point, we’re just anticipating needs, and carrying through the `use-auth` work would be a lot of work now for mere anticipation.

Removing all auth for catalog now, and making this change later, would solve the original problem I was trying to solve (reduce the complexity around the wait/don’t wait logic), and be simple to change later.

docs/docs/components/manifold-marketplace.md
dangodev: Ah this was a typo. Was supposed to be `use-auth`

src/components/manifold-marketplace/manifold-marketplace.tsx
sslotsky: I thought Digital Ocean would need this, but I'm not opposed to waiting.

dangodev: Cool. In that case, I’ll close this PR, and we’ll just count on **always** skipping auth for catalog in web components (and remove the TODOs basically).

dangodev:
Closing; let’s keep all catalog requests public for now and revisit when needed (hopefully we’ll be on GraphQL by then)

PR #410, opened 2019-08-21
Fixed: Disable provision button while provisioning

sslotsky:
## Reason for change

The provision button currently remains enabled while a provision is in progress, which allows the user to request a provision multiple times. This change disabled the button during provisioning so that only one provision can occur.

This was solved in the simplest manner possible: adding a `provisioning` state variable to the component and setting it at the appropriate times. It would sure be nice to have query/mutation components that would allow us to avoid cluttering up all of our data components with things like loading state.

## Testing

Pull down the branch and try it in the Render dashboard. When a provision is in progress, the provision button should be disabled.

## Checklist

### DIFF
--- a/src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
+++ b/src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
@@ -290,18 +290,24 @@ describe('<manifold-data-provision-button>', () => {
         `,
       });
 
-      const instance = page.rootInstance as ManifoldDataProvisionButton;
-      instance.success.emit = jest.fn();
+      const success = jest.fn();
+      page.doc.addEventListener('manifold-provisionButton-success', success);
 
-      await instance.provision();
+      const button = page.doc.querySelector('button[type=submit]') as HTMLButtonElement;
+      button.click();
+      await page.waitForChanges();
 
       expect(fetchMock.called(`${connections.prod.gateway}/resource/`)).toBe(true);
-      expect(instance.success.emit).toHaveBeenCalledWith({
-        createdAt: GatewayResource.created_at,
-        message: 'test successfully provisioned',
-        resourceId: GatewayResource.id,
-        resourceLabel: GatewayResource.label,
-      });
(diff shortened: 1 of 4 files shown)


src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -204,7 +208,7 @@ export class ManifoldDataProvisionButton {
       <button
         type="submit"
         onClick={() => this.provision()}
-        disabled={!this.planId || !this.productId}
+        disabled={!this.planId || !this.productId || this.provisioning}
```
dangodev: I was curious if this disabled the `onClick` or not (if we needed to add a check in the function), but it seems to! Nice.

dangodev:
This looks good to me! But I’d like to have a test for the attribute before merging

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
sslotsky: Yeah, disabled buttons can't receive clicks.

src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
```diff
@@ -290,18 +290,24 @@ describe('<manifold-data-provision-button>', () => {
...
+      expect(success).toHaveBeenCalledWith(
+        expect.objectContaining({
+          detail: {
+            createdAt: GatewayResource.created_at,
+            message: 'test successfully provisioned',
+            resourceId: GatewayResource.id,
+            resourceLabel: GatewayResource.label,
+          },
+        })
+      );
```
dangodev: See this branch here: https://github.com/manifoldco/ui/blob/dangodev/use-auth/src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts#L143

Also I’d suggest deleting the tests that test the event emitter without the button click + event name in this test suite. Those just become hard to maintain over time because they don’t allow you to change event names, etc.

PR #414, opened 2019-08-21
use our cdn for documentation

domenicrosati:
we us js.cdn.manifold.co for our cdn now.
so lets replace the documentation with it!


domenicrosati:
blocked by having the latest syncing working which requires delete permissions.
also blocked by CORS policies being fixed

dangodev:
@domenicrosati this is good to merge preemptively! We’re not shipping things to prod right away just yet.

PR #415, opened 2019-08-21
use cache restore for stencil instead of setup

domenicrosati:
closes: https://github.com/manifoldco/engineering/issues/9007

## Reason for change

The configuration for pushing to our cdn isn't running in a container with npm support, needed for the setup step. Instead we can just restore the files cached from the npm-publish job so we can push the pkg file that is generated by npm-publish for distributing the ui build on our cdn.

## Testing

Unfortunately we will have to merge it and tag a prerelease to see.

## Checklist

n/a - circleci change

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,13 @@ commands:
             - pkg
           key: stencilcache-{{ .Revision }}
 
+  save-pkg-build:
+    steps:
+      - save_cache:
+          paths:
+            - pkg
+          key: npmpkgcache-{{ .Revision }}
+
   # npm install
   setup:
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -107,7 +107,9 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/project
-      - setup
+      - restore_cache: # Restore Stencil cache if exists
+          keys:
+            - stencilcache-{{ .Revision }}
```
domenicrosati: @dangodev unless we need the pkg file in stencilcache, perhaps we should factor it out? (for another PR)

sslotsky:
> Unfortunately we will have to merge it and tag a prerelease to see.

Can we tag a prerelease without merging?

domenicrosati:
No :( because the circle ci needs to be in

.circleci/config.yml
dangodev: It’s only cached for a Git hash (`.Revision`), so it should be blown away every time it changes

PR #417, opened 2019-08-21
Reset auth token if failed

dangodev:
## Reason for change

This re-calls the function if the auth token failed

TODO:
- [ ] tests

## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -24,10 +24,8 @@ export class ManifoldAuthToken {
   }
 
   setExternalToken(token?: string) {
-    if (token) {
-      if (!isExpired(token)) {
-        this.setAuthToken(token);
-      }
+    if (token && !isExpired(token)) {
+      this.setAuthToken(token);
     }
   }
 
--- a/src/utils/restFetch.ts
+++ b/src/utils/restFetch.ts
@@ -8,7 +8,7 @@ export interface CreateRestFetch {
   endpoints?: Connection;
   wait?: number;
   getAuthToken?: () => string | undefined;
-  setAuthToken?: (token: string) => void;
+  setAuthToken?: (token: string | null) => void;
 }
 
 interface RestFetchArguments {
@@ -68,9 +68,8 @@ export function createRestFetch({
 
     /* Handle expected errors */
     if (response.status === 401) {
-      setAuthToken('');
+      setAuthToken(null); // this will re-request a new token
       report(response);
-      throw new Error('Auth token expired');
     }
 
     const body = await response.json();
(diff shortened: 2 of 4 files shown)


src/utils/restFetch.ts
```diff
@@ -68,9 +68,9 @@ export function createRestFetch({
 
     /* Handle expected errors */
     if (response.status === 401) {
-      setAuthToken('');
+      setAuthToken(null);
       report(response);
-      throw new Error('Auth token expired');
+      return restFetch(args); // retry
```
sslotsky: Is there any logic in place to refresh the iframe when this happens? Or will that be in a separate PR?

dangodev: Oh good call. I’m a bit stumped on how to go about doing that. What would you suggest?

dangodev: You’re totally right—that’s the issue

sslotsky: Been scratching my head on that one for a while. One possibility is to have shadowcat listen for the event that's getting reported when the 401 happens. But one potential difficulty there is that we could get a bunch of 401 errors at the same time, and reloading the iframe a bunch of times in a row is definitely not ideal.

dangodev: Crazy idea: one old cachebusting trick is the ol’ arbitrary searchparam set by timestamp. e.g. `https://auth.manifold.co/oauth?t=123430938023`

We could do that, but it’d still have the problem of firing over and over again like you said.

We could clamp it to `5s` intervals or something, though and… yeah this is getting complicated 🤔

sslotsky: I considered clamping to a time interval as well. It might not be awful.

dangodev: Hm. Just tried dynamically creating an `iframe` in JS, and it sends one request every time you set the `src` even if you set it to the same thing over again. So we’d have to rely on Stencil’s `@Watch` decorator only allowing actual changes to come through (which I think it well)

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -24,10 +24,20 @@ export class ManifoldAuthToken {
...
 
   setExternalToken(token?: string) {
-    if (token) {
-      if (!isExpired(token)) {
-        this.setAuthToken(token);
-      }
+    if (token && !isExpired(token)) {
+      // if token not expired, set
+      this.setAuthToken(token);
+    } else if (this.oauthUrl) {
```
sslotsky: I don't think clearing the auth token will result in the `token` prop being changed. That prop is being set by platforms. In order to achieve the implementation you're aiming at, you might have to inject the token from the tunnel as well, and set a new `@Watch()` on that property that kicks off this logic when it changes from present to null.

dangodev: Oh shoot. You’re right again.

Ugh my brain is breaking trying to follow all this stuff.

dangodev: It’s just frustrating because I’m looking at a Render dashboard right now that definitely doesn’t work after only a few minutes, and refreshing doesn’t help. I suspect it’s the localStorage stuff we added.

dangodev:
Closing whatever was going on here… Best to start over again probably.

PR #418, opened 2019-08-21
add conditional sync to latest

domenicrosati:
enhances: manifoldco/engineering#9007

## Reason for change

we'd like @manifoldco/ui to point to our latest webcomponet build so that we can continue to use it in dev and examples.

## Testing

Yes! we can tag a prerelease and see this is not run. We will have to tag a release to see if the positive case is run but i can flip the bools for prerelease and see to it that works at least.

## Checklist

n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,10 @@ jobs:
   # Publish packages to js.cdn.manifold.co
   # this is always run after npm-publish so the ./pkg file will be available.
   cdn-publish:
+    parameters:
+      latest:
+        type: boolean
+        default: false
     docker:
       - image: google/cloud-sdk
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -104,15 +104,24 @@ jobs:
...
   # this is always run after npm-publish so the ./pkg file will be available.
   cdn-publish:
+    parameters:
+      latest:
+        type: boolean
+        default: false
     docker:
       - image: google/cloud-sdk
     steps:
+      - run: echo "latest is << parameters.latest >>"
```
domenicrosati: To Remove: testing artifact

dangodev:
One weird scenario is with npm you can back-publish versions, so technically it’d be possible to release a version older than the most current.

But that’s a rare scenario, and on top of that I think CircleCI accidentally makes that impossible (I tried and it wouldn’t let me)

PR #423, opened 2019-08-22
MVP for token refresh

sslotsky:
## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -11,12 +11,21 @@ export class ManifoldAuthToken {
   @Prop() setAuthToken: (s: string) => void = () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
+  @Prop() accessToken?: string;
   @Prop() oauthUrl?: string;
   @Event({ eventName: 'manifold-token-receive', bubbles: true })
   manifoldOauthTokenChange: EventEmitter;
 
   @Watch('token') tokenChange(newToken?: string) {
     this.setExternalToken(newToken);
+  }
+
+  @Watch('accessToken') internalTokenChange(newToken?: string) {
+    if (this.accessToken && !newToken && this.oauthUrl) {
+      const url = new URL(this.oauthUrl);
+      url.searchParams.set('ts', new Date().getTime().toString());
+      this.oauthUrl = url.href;
+    }
   }
 
   componentWillLoad() {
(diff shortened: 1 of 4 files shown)


src/utils/restFetch.ts
```diff
@@ -70,7 +70,7 @@ export function createRestFetch({
     if (response.status === 401) {
       setAuthToken('');
       report(response);
-      throw new Error('Auth token expired');
+      restFetch(args);
```
sslotsky: I guess this should `return restFetch(args)`

src/utils/graphqlFetch.ts
```diff
@@ -70,6 +70,7 @@ export function createGraphqlFetch({
       // TODO trigger token refresh for manifold-auth-token
       setAuthToken('');
       report(response); // report unauthenticated
+      graphqlFetch(args);
```
sslotsky: This should probably `return graphqlFetch(args)`

src/utils/restFetch.ts
dangodev: I think we need to `return` here so the promise will wait

src/utils/graphqlFetch.ts
dangodev: Same here: `return`?

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -19,6 +20,14 @@ export class ManifoldAuthToken {
...
   }
 
+  @Watch('accessToken') internalTokenChange(newToken?: string) {
+    if (this.accessToken && !newToken && this.oauthUrl) {
+      const url = new URL(this.oauthUrl);
+      url.searchParams.set('ts', new Date().getTime().toString());
+      this.oauthUrl = url.href;
+    }
+  }
+
```
dangodev: What’s going on here? Where is `accessToken` coming from? When does this get called?

dangodev: (to be clear, I’m just asking for a comment or something, because I’m not sure how to debug or follow this)

src/utils/restFetch.ts
dangodev: 👯

src/utils/graphqlFetch.ts
dangodev: 👯

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: It's injected by the connection. I'll add a comment above the `@Prop()` declaration similar to other injected props.

PR #428, opened 2019-08-22
Improve auth docs

dangodev:
## Reason for change

Our authentication section was written before the implementation, and is more general than instructional. This change adds example code in a more usable way, and explains the OAuth dance a little better.

## Testing

Spellcheck, etc.

## Checklist

### DIFF
--- a/docs/docs/advanced/authentication.md
+++ b/docs/docs/advanced/authentication.md
@@ -142,16 +142,16 @@ If the token given to the component is invalid, endpoints will return a 401 erro
 be removed from the `manifold-connection`. Use the [error handling capabilities](/advanced/errors)
 of our web components to detect and act on such errors.
 
-### Authenticated requests timeout
+### OAuth timeout
 
-Any requests requiring authentication - which are sent by components locked (🔒) behind
-authentication - will wait on a valid token for up to 15 seconds. If this component does not inject
-a token into the connection after that time, an authentication error will be thrown.
+Any requests requiring authentication will wait on a valid token for up to 15 seconds. If this
+component does not inject a token into the connection after that time, an authentication error will
+be thrown.
 
-This timeout duration can be customized on the `manifold-connection` component.
+This timeout duration can be changed on the `manifold-connection` component:
 
 ```html
-<manifold-connection wait-time="time-in-ms">
+<manifold-connection wait-time="10000">
   <!-- Application -->
 </manifold-connection>
 ```
--- a/docs/docs/advanced/authentication.md
+++ b/docs/docs/advanced/authentication.md
@@ -37,10 +37,9 @@ here’s a general summary of how it works with UI, and what to expect:
 
 ### Step 1: `<manifold-auth-token>`
 
-To make our side of it easy, we provide a special token—`<manifold-auth-token>`—that handles the
-first part of the OAuth dance. It should be placed inside the [Connection][connection] component. It
-may appear anywhere in your app (but the higher it the DOM tree it appears, the better, so it can
-load sooner):
+To make it easy, we provide a special component—`<manifold-auth-token>`—that handles the first part
+of the OAuth dance. It should be placed inside the [Connection][connection] component. It may appear
+anywhere in your app (but the higher it the DOM tree it appears, the better, so it can load sooner):
 
 ```html
 <manifold-connection>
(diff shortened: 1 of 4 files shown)


docs/docs/advanced/authentication.md
```diff
@@ -1,70 +1,184 @@
...
+</manifold-toast>
+
+Everything else not related to product data—user provisions, user resources, account
+credentials—need auth. The full auth docs can be found at [docs.manifold.co][authentication], but
+here’s a general summary of how it works with UI, and what to expect:
+
+### Step 1: `<manifold-auth-token>`
+
+To make our side of it easy, we provide a special token—`<manifold-auth-token>`—that handles the
+first part of the OAuth dance. It should be placed inside the [Connection][connection] component. It
```
batgirl: On a first read from the top, I noticed that this is the first mention of an `OAuth dance` and my brain wondered if I should know what that is referring to at this point. Maybe a tiny amount of context here would help with that?

`...handles the OAuth dance necessary for platform users to authenticate with Manifold.`

Also what do you mean by `our side` in this paragraph? Maybe `your side` would be better?

docs/docs/advanced/authentication.md
```diff
@@ -1,70 +1,184 @@
...
+  <manifold-auth-token></manifold-auth-token>
+  <!-- rest of your HTML -->
+</manifold-connection>
+```
+
+In addition to loading this as soon as possible, we also recommend **only loading this once** to
+prevent unnecessary token requests (another reason for only having this be near the root of your
+HTML).
+
+### Step 2: your app
```
batgirl: Maybe capitalize `Your`?

docs/docs/advanced/authentication.md
```diff
@@ -1,70 +1,184 @@
...
+This will then try to re-use an existing token, which can speed up the user experience by saving an
+OAuth dance.
+
+In the case of an expired token, the component will try, silently fail, request a new one, and
+trigger the `manifold-token-receive` event which should update anything listening without any
+interruption on the user side.
+
+## Events
+
+The following custom events are emitted to `document`. You can listen for events like so:
```
batgirl: Are these custom events always emitted to `document`? I thought you could technically add the event listener anywhere.

docs/docs/advanced/authentication.md
```diff
@@ -1,70 +1,184 @@
...
+### Token expiration
+
+For the `manifold-token-receive` event (**Events** section), the `expiry` is the Unix timestamp when
+the token will expire. On our end, we’ll automatically refresh tokens with no action on your part,
+and with no interruption to the user (other than a slightly-slowed-down request while the token
+refreshes).
+
+But for your needs, the `manifold-token-receive` event allows you to preemptively do anything you’d
+like to (such as, say, not passing the token to `<manifold-auth-token>` which could result in a
+minor performance boost for users by skipping what will be a `401` request before the token is
```
batgirl: This sentence is the only part that confuses me a little - not passing the token to `<manifold-auth-token>` will skip the `401` request before it's refreshed.. Could we say `response` instead of `request` here?

I guess my question as a user reading it is being unsure of what the alternative to passing the token to that component is, which is probably described above but I'm having trouble connecting the ideas. The answer is that the auth flow will work even if we don't pass a token to that component, right? And platforms can control whether they keep making requests that will `401` from inside the `manifold-token-receive` event listener?

docs/docs/advanced/authentication.md
dangodev: Heh I just re-read that sentence and it seemed weird. I like your suggestion!

docs/docs/advanced/authentication.md
dangodev: Yeah I was on the fence there—I’m using `Sentence case` consistently for all titles. Not sure if a colon constitutes a capitalization… it does look a bit weird but 🤷‍♂

docs/docs/advanced/authentication.md
dangodev: It bubbles, so yes. And you should be able to add the event listener on the `<manifold-auth-token>` component or any of its parents, all the way up to `document`. I just generally like `document` because it’s easy, and because querySelection can be messy depending on the setup (React or not, etc.)

docs/docs/advanced/authentication.md
dangodev: >  the auth flow will work even if we don't pass a token to that component, right?
Yup! Passing a token is merely recommended to not be fetching new tokens all the time.
> platforms can control whether they keep making requests that will 401 from inside the manifold-token-receive event listener?
No; that’s on us. But they can prevent that initial `401` by keeping track of when tokens expire, and knowing when to discard them. I didn’t really go into detail there because IMO that’s a micro-optimization that they can figure out on their own.

PR #430, opened 2019-08-22
Added: Update oauth URL to restart oauth when token is unset

sslotsky:
## Reason for change

Currently, if we receive a 401 status from the API, an unhandled exception will be thrown, and users will have to refresh the page to obtain a new auth token.

With this change, a 401 will cause the token to be refreshed, and the API call will retry.

## Testing

Testing is tricky because tokens take a while to expire. You might want to [watch this video](https://manifoldco.slack.com/archives/CDW445F53/p1566568237455700) where @domenicrosati demonstrates this working with fast-expiring tokens.

## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -14,7 +14,7 @@ export class ManifoldAuthToken {
   @Prop() subscribe: (s: Subscriber) => () => void = () => () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
-  @Prop() oauthUrl: string = 'https://login.manifold.co/signin/oauth/web';
+  @Prop() oauthUrl?: string = 'https://login.manifold.co/signin/oauth/web';
   @Event({ eventName: 'manifold-token-receive', bubbles: true })
   manifoldOauthTokenChange: EventEmitter;
 
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -11,7 +11,7 @@ export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() setAuthToken: (s: string) => void = () => {};
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() subscribe: (s: Subscriber) => () => void;
+  @Prop() subscribe: (s: Subscriber) => () => void = () => () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
   @Prop() oauthUrl: string = 'https://login.manifold.co/signin/oauth/web';
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -4,16 +4,21 @@ import { AuthToken } from '@manifoldco/shadowcat';
 import Tunnel from '../../data/connection';
 import logger from '../../utils/logger';
 import { isExpired } from '../../utils/auth';
+import { Subscriber } from '../../state/connection';
 
 @Component({ tag: 'manifold-auth-token' })
 export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() setAuthToken: (s: string) => void = () => {};
+  /** _(hidden)_ Passed by `<manifold-connection>` */
+  @Prop() subscribe: (s: Subscriber) => () => void;
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
-  @Prop() oauthUrl?: string;
+  @Prop() oauthUrl: string = 'https://login.manifold.co/signin/oauth/web';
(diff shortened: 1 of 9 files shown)


src/utils/graphqlFetch.ts
```diff
@@ -29,10 +30,14 @@ export type GraphqlFetch = <T>(args: graphqlArgs) => Promise<GraphqlResponseBody
...
   endpoint = 'https://api.manifold.co/graphql',
   wait = 15000,
+  retries = 0,
   getAuthToken = () => undefined,
   setAuthToken = () => {},
 }: CreateGraphqlFetch): GraphqlFetch {
-  return async function graphqlFetch<T>(args: graphqlArgs): Promise<GraphqlResponseBody<T>> {
+  return async function graphqlFetch<T>(
+    args: graphqlArgs,
+    attempts: number = 0
```
sslotsky: If there's one flaw here, it's that I can call `restFetch(args, attempts)` from consuming code. That would indicate that I've already attempted to fetch some number of times when in fact that's not true.

I could possibly fix this by removing the `return` statement from this function definition and making it local, and then returning an outer function that just kicks this one off.

dangodev: Ohh I forgot that it’s only initialized once. You’re right. I think this is fine

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -4,23 +4,41 @@ import { AuthToken } from '@manifoldco/shadowcat';
...
 @Component({ tag: 'manifold-auth-token' })
 export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() setAuthToken: (s: string) => void = () => {};
+  /** _(hidden)_ Passed by `<manifold-connection>` */
+  @Prop() subscribe: (s: Subscriber) => () => void = () => () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
-  @Prop() oauthUrl?: string;
+  @Prop() oauthUrl: string = 'https://login.manifold.co/signin/oauth/web';
```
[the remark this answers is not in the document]
dangodev: Screen Shot 2019-08-23 at 08 33 11

Built UI and imported it into Sphericon, and yeah it’s just reading the types from `components.d.ts`, so anything not optional must be passed as a prop.

I remember Terry running into this on www, where a prop wasn’t accidentally marked optional. It’s only required for top-level (consumer) components; internal components we can (and should) make things mandatory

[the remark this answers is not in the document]
sslotsky: I wonder if we should tweak our lint rule then. We have a rule where props that can't be used as HTML attributes need to be optional _or_ have a default value, but it seems like they should be optional every time?

src/utils/graphqlFetch.ts
sslotsky: It might be fine, but I could also do this:

```ts
  return async function(args: RestFetchArguments) {
    return restFetch(args, 0);
  };

  async function restFetch<T>(args: RestFetchArguments, attempts: number): Promise<T | Success> {
```

That will prevent the possibility of abuse.

src/components/manifold-auth-token/manifold-auth-token.tsx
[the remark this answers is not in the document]
dangodev: Well IMO that’d be a burden for us to deal with optional props for our internal components. I think in general we just need to have a clearer distinction between user components and internal components.

My personal (slightly-crazy) opinion: **all** components should be public-facing, and we should just deal with big components (or abstract things to functional components if it’s massive). So yes in that world all props would be optional, and everything would be documented.

That’s a little lofty, but IMO would reduce the mental overhead of imagining what’s consumer-facing and what’s not, since technically users _can_ use any component in the bundle, and that’s kinda the web component way™

sslotsky:
@dangodev I switched the `oauthUrl` prop back to being optional and added some tests.

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -4,23 +4,41 @@ import { AuthToken } from '@manifoldco/shadowcat';
...
 import logger from '../../utils/logger';
 import { isExpired } from '../../utils/auth';
+import { Subscriber } from '../../state/connection';
 
 @Component({ tag: 'manifold-auth-token' })
 export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
   @Prop() setAuthToken: (s: string) => void = () => {};
+  /** _(hidden)_ Passed by `<manifold-connection>` */
+  @Prop() subscribe: (s: Subscriber) => () => void = () => () => {};
```
dangodev: I think we’d benefit from these other props being optional, but we can tackle that later and I’m fine with merging this as-is. I’m sure there are other TypeScript problems we’ll hit when Sphericon picks back up again.

src/utils/restFetch.ts
```diff
@@ -97,5 +103,9 @@ export function createRestFetch({
     report(response);
     const message = Array.isArray(body) ? body[0].message : body.message;
     throw new Error(message);
+  }
+
+  return function(args: RestFetchArguments) {
+    return restFetch(args, 0);
```
dangodev: In awe of the pureness of this lad

PR #434, opened 2019-08-23
Delay transitions until component render

dangodev:
## Reason for change

**Before** (hide credentials is visible on load)

**After** (hide credentials is invisible on load AND animations are preserved)

## Testing
Minor addition

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "build:storybook": "npm run build && build-storybook && npm run copy-changelog",
     "bundlesize": "npm run build && npx bundlesize",
     "copy-changelog": "node scripts/copy-changelog",
-    "dev": "npm run copy-changelog && concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
+    "dev": "concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
     "docs": "npm run prepare:docs && cd docs && npm start",
     "generate:gql": "graphql-codegen --config codegen.yml",
     "generate:specs": "npm run generate:specs:catalog && npm run generate:specs:gateway && npm run generate:specs:marketplace && npm run generate:specs:provisioning && npm run generate:specs:connector",
--- a/src/components/manifold-credentials-view/manifold-credentials-view.tsx
+++ b/src/components/manifold-credentials-view/manifold-credentials-view.tsx
@@ -1,4 +1,4 @@
-import { h, Component, Prop, Event, EventEmitter, Element } from '@stencil/core';
+import { h, Component, Prop, Event, EventEmitter, Element, State } from '@stencil/core';
 import { eye, lock, loader } from '@manifoldco/icons';
 
 import { Marketplace } from '../../types/marketplace';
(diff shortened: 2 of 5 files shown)


src/components/manifold-credentials-view/manifold-credentials-view.tsx
```diff
@@ -24,6 +25,13 @@ export class ManifoldResourceCredentials {
     this.addListeners();
   }
 
+  componentDidLoad() {
+    // hack to prevent “Hide credentials” from being visible on load in Firefox
+    setTimeout(() => {
```
sslotsky: Sad. Do we _really_ have to?

dangodev: It’s not as “hacky” as it seems (OK I did say “hack”).

The reason for the set timeout is:
1. component loads, and animates the button, which then “flashes” as it’s hiding
2. this transition takes ~`150ms` after rendering
3. with a little padding, we do want to enable transitions, but only `150ms` after rendering

There’s not a way to handle this in CSS, because if you took off the transitions when it was closed, then the button wouldn’t smoothly animate out as it is doing. We honestly just want to prevent transitions from happening for the first _n_ milliseconds after initial render.

dangodev: In that light I would argue this is one of the rare instances `setTimeout` isn’t actually a hack, because we want to wait for the duration of an animation after an event happens

src/components/manifold-credentials-view/manifold-credentials-view.tsx
```diff
@@ -95,7 +103,11 @@ export class ManifoldResourceCredentials {
     );
 
     return [
-      <menu class="secrets-menu" data-showing={!!this.credentials}>
+      <menu
+        class="secrets-menu"
+        data-showing={!!this.credentials || undefined}
```
domenicrosati: classic

package.json
```diff
@@ -31,7 +31,7 @@
     "build:storybook": "npm run build && build-storybook && npm run copy-changelog",
     "bundlesize": "npm run build && npx bundlesize",
     "copy-changelog": "node scripts/copy-changelog",
-    "dev": "npm run copy-changelog && concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
+    "dev": "concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
```
dangodev: We don’t need this for dev anymore

src/components/manifold-credentials-view/manifold-credentials-view.tsx
davidleger95: I feel like there's probably a better way of doing this but my brain doesn't want to figure it out on a Friday afternoon 🙃I might take a look at this next time I have to do something in this component.

dangodev: There might be a way to do it with animations, but in general I like transitions because animations always have to reset to start/end positions but transitions don’t.

Example: https://codepen.io/dangodev/pen/RwbKEqL

AFAIK the only way to deal with transitions where the entry animation is different, is to add a class or something via JS (or take a class away). This is basically what that’s doing.

dangodev: Anyway that explains my reasoning—that said if anyone can accomplish this in a smarter way I’m 10000% in favor! I’ll be thinking about it as well.

PR #452, opened 2019-08-27
copy-credentials component

Minivera:
Work is related to manifoldco/engineering#9042

## Reason for change

Adds a `<manifold-copy-credentials>` component that will copy creds to the clipboard on click.

## Testing

- Storybook: test out the “Copy Credentials” component. Use **Knobs** to set the resource name (but put your own token in localStorage)
- Docs: review the docs; see if they make sense

## Checklist


src/components/manifold-data-sso-button/manifold-data-sso-button.spec.ts
```diff
@@ -221,15 +229,19 @@ describe('<manifold-data-sso-button>', () => {
...
       const resourceLabel = 'success-label';
       element.resourceLabel = resourceLabel;
       page.root.appendChild(element);
       await page.waitForChanges();
 
       const button = element.querySelector('button');
-      if (!button) throw new Error('button not found in document');
+      if (!button) {
+        throw new Error('button not found in document');
+      }
```
dangodev: IIRC just one other file had these changes. If we enforce it there, too, we can add the (mysteriously-) missing lint rule

Minivera: It was here before, but blame says I removed it? 🤔 https://github.com/manifoldco/ui/blame/b9ef628b8885a104e3de75f00b2abed9614f2466/.eslintrc.js#L58

dangodev:
So this successfully grabs credentials for me, but the clipboard isn’t working:

| Browser | Copy to Clipboard works |
|:--------|:--------------------------|
| Firefox 70 | 🚫 |
| Chrome 76  | ✅   |
| Safari 12.1.2 | 🚫 |

Should this be addressed in this PR, or a followup?

dangodev:
Also, this module is `0.2 KB`: https://www.npmjs.com/package/copy-text-to-clipboard. If this works, then I think it’s fine to add this as a dep (it’ll only be loaded for this one component, not globally).

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a typo in manifold-data-deprovision-button documentation (#457)
 - Upgraded `@manifoldco/shadowcat` to feature the latest security patches. (#477)
 
+### Added
+- Added a new `manifold-data-get-credentials-button` component to fetch the credentials of a resource on click. Those credentials are passed as a dom event.
```
Minivera: Might want to change this to reflect the recent changes

docs/docs/data/manifold-copy-credentials.md
```diff
@@ -0,0 +1,74 @@
+---
+title: 🔒 Copy Credentials
+path: /data/copy-credentials
+example: |
+  <manifold-copy-credentials resource-label="my-resource">
+    Copy
+  </manifold-copy-credentials>
+---
+
+# 🔒 Get credentials Button
```
Minivera:
```suggestion
# 🔒 Copy credentials Button
```

src/components/manifold-copy-credentials/manifold-copy-credentials.tsx
```diff
@@ -0,0 +1,132 @@
...
+interface ErrorDetail {
+  error: string;
+  resourceLabel?: string;
+}
+
+interface SuccessDetail {
+  resourceLabel?: string;
+}
+
+@Component({ tag: 'manifold-copy-credentials' })
```
Minivera: Any reason why this isn't a data component anymore?

dangodev: Good question! Unfortunately this is a long, boring explanation as to why there were data components in the first place. But here, I think the simplest answer is `<manifold-copy-credentials>` is the longest HTML element I’d like to type, personally.

In general, let’s prefer simplicity and straightforward naming to plaforms over consistency (ideally, adhering to the first will contribute to the latter automatically). We’ll have a consistency discussion very very soon.

src/components/manifold-copy-credentials/manifold-copy-credentials.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+      credentials.forEach(credential => {
+        if (!credential || !credential.node) {
+          throw new Error('credential node missing');
+        }
+        expect(html.includes(credential.node.key)).toBe(false);
+        expect(html.includes(credential.node.value)).toBe(false);
+      });
+    });
+  });
+});
```
dangodev: Overall I’m pretty happy with this test. I’m starting to land on a pattern:

- `v0 HTML`: HTML attributes we’ll need to review for breaking changes in v1. Failing a test here means breaking implementation in platforms.
- `v0 events`: same as `v0 HTML` but for events
- `security`: tests that come up during security audits (here, since it deals with credentials, we have quite a few).

dangodev: This test checks that:
- Things are copied to the clipboard (as best we can figure with a mocked clipboard library; it’s tough to test in true E2E because of the endpoint, but we may figure out a way eventually)
- credentials don’t appear in HTML or event data (to prevent rogue scripts from listening)
- event listeners with proper names and event details reach the `document`

.manifold-local.yml
```diff
@@ -1,2 +0,0 @@
-team: manifold
-project: manifold-ui-local
```
dangodev: This file was deleted because the project can be changed with `manifold run -p`. Moving files around can get weird.

package.json
```diff
@@ -31,7 +31,7 @@
     "build:storybook": "npm run build && build-storybook && npm run copy-changelog",
     "bundlesize": "npm run build && npx bundlesize",
     "copy-changelog": "node scripts/copy-changelog",
-    "dev": "concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
+    "dev": "concurrently 'stencil build --watch' 'rm -rf dist && wait-on dist/loader/index.cjs.js && manifold run -p manifold-ui-local -- npm run storybook'",
```
dangodev: The dev command was changed because Storybook was intermittently failing (Stencil doesn’t blow away the build folder, so Storybook would start booting up with one build, then it would swap midway and fail). This fixes that problem so Storybook should start every time now.

dangodev: Also I added a `-p` flag to set the project, rather than moving files around.

package.json
```diff
@@ -76,7 +76,8 @@
   "dependencies": {
     "@manifoldco/gql-zero": "^0.2.0",
     "@manifoldco/shadowcat": "^0.1.7",
-    "@stencil/state-tunnel": "^1.0.1"
+    "@stencil/state-tunnel": "^1.0.1",
+    "clipboard-polyfill": "^2.8.6"
```
dangodev: This uses a tiny, TypeScript-based library that’s well-maintained. We probably could write something ourselves, but I’d personally pay a few KB to not have to deal with clipboard debugging anymore (I think I speak for Gui, too 😅)

stories/manifold-copy-credentials.stories.js
```diff
@@ -0,0 +1,25 @@
...
+storiesOf('Copy credentials 🔒', module)
+  .addDecorator(withKnobs)
+  .addDecorator(manifoldConnectionDecorator)
+  .add('default', () => {
+    const resourceLabel = text('resource-label', 'config');
+    text('📋 test paste', '');
+    return `
+      <style>
+        button {
+          cursor: hover;
```
davidleger95: what's a hover cursor? 🤔I don't see it as a valid property. is it just a pointer?

src/components/manifold-copy-credentials/manifold-copy-credentials.tsx
```diff
@@ -0,0 +1,116 @@
...
+        const detail: ErrorDetail = { resourceLabel: this.resourceLabel, message };
+        this.error.emit(detail);
+      });
+    }
+
+    // if creds, set it and continue
+    if (data && data.resource && data.resource.credentials) {
+      const credentials: string[] = data.resource.credentials.edges.reduce(
+        (credList, { node }) => (node ? [...credList, `${node.key}=${node.value}`] : credList),
+        []
```
davidleger95: any reason this isn't a map?

```js
data.resource.credentials.edges.map(({ node }) => `${node.key}=${node.value}`)
```

src/components/manifold-copy-credentials/manifold-copy-credentials.tsx
```diff
@@ -0,0 +1,116 @@
...
+  @State() credentials?: string;
+  @Event({ eventName: 'manifold-copyCredentials-error', bubbles: true }) error: EventEmitter;
+  @Event({ eventName: 'manifold-copyCredentials-success', bubbles: true }) success: EventEmitter;
+
+  @Watch('resourceLabel') labelChange() {
+    this.refresh();
+  }
+
+  componentWillLoad() {
+    this.refresh();
```
davidleger95: didn't know components had this. is it a Stencil thing or a Web Components thing?

stories/manifold-copy-credentials.stories.js
dangodev: Good catch! Typo.

src/components/manifold-copy-credentials/manifold-copy-credentials.tsx
dangodev: Yeah it should be. I think I was turning it into a string before, but stopped because I find `reduce` hard to follow. `map` would be even easier you’re right.

src/components/manifold-copy-credentials/manifold-copy-credentials.tsx
dangodev: Neither; it’s just a generic class method I named. Oh it does need a `@Method` decorator to be accessible from the DOM like I wanted, though.

davidleger95: oh haha, didn't notice that was the method below. I thought it was built in and just remounted the component

dangodev: Oh got it. Yeah I wanted to name it something users would find intuitive. I kinda like this pattern, and think we should use `@Method` probably in more instances.

PR #470, opened 2019-09-04
Get the forestry previews working

Minivera:
This adds the required commands to get the forestry,io previews working. You can see them in action here: http://uno0idirlrdk6w.instant.forestry.io/getting-started/?theme=default

Work is related to manifoldco/engineering#9198


package.json
```diff
@@ -48,6 +49,7 @@
     "lint:js": "eslint --ext .js,.ts,.tsx src",
     "now-build": "npm run build:storybook",
     "prepare:docs": "npm run prepare:package && npm run copy-changelog",
+    "prepare:forestry": "npm install && cd docs && npm install @manifoldco/ui@latest",
```
Minivera: This will change the docs `package.json` file if run locally, maybe there's a better way to do this? Building the UI repo in the forestry preview is sadly not possible as is causes the small container they give us to run out of memory.

package.json
```diff
@@ -33,6 +33,7 @@
     "copy-changelog": "node scripts/copy-changelog",
     "dev": "concurrently 'stencil build --watch' 'wait-on dist/loader/index.cjs.js && cp .manifold-local.yml .manifold.yml && manifold run -- npm run storybook'",
     "docs": "npm run prepare:docs && cd docs && npm start",
+    "docs:forestry": "cd docs && npm run start:forestry",
```
dangodev: This is the dev command; maybe this should just replace `docs`? Don’t see why we need 2 dev commands

dangodev: Oh it’s for previews. I see. This is fine, then

package.json
dangodev: That’s fine for now. Maybe we want to version-lock docs later and we can improve this then

PR #474, opened 2019-09-04
convert manifold-data-resource-list to gql

davidleger95:
## Reason for change

- convert `manifold-data-resource-list` to GraphQL

## Testing

- In Storybook, with a valid token set, go to "Resource List 🔒 [Data]"
- Check that the resources are being loaded.
- Check that the links are set properly and the event triggered has the correct data

## Checklist


src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -50,71 +59,75 @@ export class ManifoldDataResourceList {
...
 
-  userResources(resources: RealResource[]) {
-    return resources.filter(
-      ({ body: { owner_id, user_id } }) =>
-        user_id || (owner_id && owner_id.indexOf('/user/') !== -1)
+  static userResources(resources: ResourceConnection) {
+    return resources.edges.filter(
+      ({ node }) => node && node.owner && node.owner.id.indexOf('/user/') !== -1
     );
   }
```
davidleger95: Not quite sure what this is for. Do we even need it?

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -50,51 +59,52 @@ export class ManifoldDataResourceList {
...
+      }
+
+      if (errors) {
+        errors.forEach(error => {
+          throw new Error(error.message);
+        });
+      }
+    } catch (e) {
+      console.error(e);
     }
```
davidleger95: This might be redundant

dangodev: I think this is. Also, I don’t know if we need to throw errors; let’s just display them to the user. GraphQL fetch should handle reporting.

dangodev: Also, according to this we would `throw` if there were errors and data. I don’t think there’s a situation where those would both be something, but if there was `data` should we care about errors? IMO if we have what we need, let’s display it to the user without interrupting their experience, and rely on our logging to surface any errors to the platform.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
dangodev: If this isn’t being used, then just delete it

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
davidleger95: there could be some data in a query that requires auth or isn't found for some other reason, which would return an error, and other data that is resolved correctly. I don't think that would happen in this instance, but it is possible for some queries.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
davidleger95: It was being used in for filtering results from the rest API. It was filtering out team resources, but would those even be possible to get on a platform?

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
dangodev: The auth part is handled in `graphqlFetch`. You’re right that a bad query would produce no data, but that would also be logged via `graphqlFetch` as well. IMO maybe only handle errors specific to this component? If something goes wrong, I’d expect an empty list to be returned, but that should be handled by the `if (data) {` part that TypeScript already kinda makes you do

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -12,28 +13,36 @@ interface EventDetail {
...
-  body: RealResourceBody;
-}
-
-interface RealResourceBody extends Marketplace.ResourceBody {
-  team_id?: string;
-  user_id?: string;
-}
+const query = gql`
+  query RESOURCES {
+    resources(first: 500) {
```
sslotsky: Similar to our discussion about using GraphQL for the marketplace grid, how are we going to know when it's time to change this to use pagination? I'm ok with temporary solutions as long as we have plans for a long term solution. IMO we either need to prioritize that implementation now by making a ticket that we agree to work on in the very near future, or we have some way of alerting the team that we need to take action soon.

davidleger95: I feel like this is a design discussion (which I'm just remembering now we were supposed to have at the start of this week haha). Do we want infinite scrolling? a 'Load More' button? something else?

davidleger95: but yes, I agree. we should handle this.

davidleger95: one thing to keep in mind as well: this seems like uncovered additional work beyond the graphql conversion. it involves behavior changes to the UI, which is going to slow us down in our main objective. in my opinion -- putting my product hat on here -- we better serve the customers  with getting the performance increases of GQL out the door faster with setting this hard limit, and then circle back immediately after to handle scaling the UI with pagination/lazyloading/etc. This would also give design time to think through these changes.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
dangodev: Oh. Hm. Yeah that‘s a good point. Let me see if this is working in Render; my gut says it won’t be needed, because filtering should be happening from the auth token in platforms (we were only doing it here as a stopgap)

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
sslotsky: I don't disagree with any of this, and I especially agree that talking about it with design and product will be necessary (we'll also come up with different answers for the resource list than we will for the marketplace grid, almost certainly). 

I just want to make sure we prioritize it, otherwise we risk leaving it like this longer than we should. Granted, it seems unlikely that a user would have 500 resources provisioned. But it _could_ happen so we need to be prepared for it.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
dangodev: Ah yeah this actually doesn’t work in our platforms. The `/user/` part is filtering out resources it should be able to see. Let’s remove filtering altogether in the front-end here.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
dangodev: Ah yeah this actually doesn’t work in our platforms. The /user/ part is filtering out resources it should be able to see. Let’s remove filtering altogether in the front-end here.

davidleger95: awesome :D

dangodev: Cool. This part’s working now. Maybe delete the now-unused class method? I can ✅

PR #476, opened 2019-09-04
Simplify GraphQL query

dangodev:
## Reason for change

I realized there’s a problem with the GraphQL types we’re using (which is kinda my fault). Consider the following example:

```ts
const { data } = await graphqlFetch<'products'>({ query });
if (data) {
  this.products = data.products
}
```

The `<'product'>` generic declares what we’re expecting from the GraphQL query. But it brings with it the following problems:

**Problem 1: no combined queries** If we queried for `products` and `resources` in the same query, this would throw a type error. So this type system restricts us unnecessarily in the biggest benefit of GraphQL: combined queries.

**Problem 2: bad queries** Further, say our `products` query succeeded but `resources` failed. How would we know? We need to handle the case of either—or both—being undefined.

**Problem 3: manually-managing types** But lastly, this is a manual list that we have to maintain outside of our GraphQL endpoint. If our GraphQL endpoint changes, then our tests won’t catch it because we’re using manually-managed types. This isn’t smart.

### What this PR does
This uses the auto-generated `Query` to fix all of the above problems. Sure, it results in an extra null check on our side, but gives us much more in return.

Props to @davidleger95 who did it this way originally and I suggested he change it to this. I’m a dummy.

## Testing

Tests should pass. Also inspect the following components in Storybook to make sure they still work:

- `manifold-data-product-logo`
- `manifold-data-product-name`
- `manifold-data-provision-button`
- `manifold-product`
- `manifold-resource-list`

## Checklist

### DIFF
--- a/src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
+++ b/src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
@@ -46,7 +46,7 @@ export class ManifoldDataProductLogo {
     const variables = { productLabel };
     const { data } = await this.graphqlFetch({ query, variables });
 
-    this.product = data && data.product ? data.product : undefined;
+    this.product = (data && data.product) || undefined;
   };
 
   @logger()
--- a/src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
+++ b/src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
@@ -44,12 +44,9 @@ export class ManifoldDataProductLogo {
 
     this.product = undefined;
     const variables = { productLabel };
-    const { data } = await this.graphqlFetch<'product'>({
-      query,
-      variables,
-    });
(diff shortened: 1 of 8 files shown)


src/components/manifold-data-product-logo/manifold-data-product-logo.tsx
```diff
@@ -44,12 +44,9 @@ export class ManifoldDataProductLogo {
...
     this.product = undefined;
     const variables = { productLabel };
-    const { data } = await this.graphqlFetch<'product'>({
-      query,
-      variables,
-    });
+    const { data } = await this.graphqlFetch({ query, variables });
 
-    this.product = data ? data.product : undefined;
+    this.product = data && data.product ? data.product : undefined;
```
sslotsky: I'm guessing that `data.product` is a `Product | null` rather than `Product | undefined`, right? So maybe if you change the component type to `Product | null` you could just do `data && data.product` here?

PR #477, opened 2019-09-04
Update shadowcat

dangodev:
## Reason for change

Updates UI with the latest security fixes from manifoldco/shadowcat.

## Testing

Testing has been completed in a partner platform, and everything still worked fine with all the new security patches. That’d be the recommended way to test this PR.

## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -9,9 +9,9 @@ import { Subscriber } from '../../state/connection';
 @Component({ tag: 'manifold-auth-token' })
 export class ManifoldAuthToken {
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() setAuthToken: (s: string) => void = () => {};
+  @Prop() setAuthToken?: (s: string) => void = () => {};
   /** _(hidden)_ Passed by `<manifold-connection>` */
-  @Prop() subscribe: (s: Subscriber) => () => void = () => () => {};
+  @Prop() subscribe?: (s: Subscriber) => () => void = () => () => {};
   /* Authorisation header token that can be used to authenticate the user in manifold */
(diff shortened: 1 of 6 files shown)


src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -26,15 +25,6 @@ export class ManifoldAuthToken {
...
     this.setExternalToken(this.token);
-    if (this.subscribe) {
-      this.unsubscribe = this.subscribe((oldToken?: string, newToken?: string) => {
-        if (oldToken && !newToken && this.oauthUrl) {
-          const url = new URL(this.oauthUrl, window.location.href);
-          url.searchParams.set('ts', new Date().getTime().toString());
-          this.oauthUrl = url.href;
-        }
-      });
-    }
```
dangodev: This is the bit I was unsure about to get it working in UI.

`<manifold-auth>` no longer has the `oauth-url` prop, which means we can’t set it. Which means we should remove the `oauth-url` prop from the `<manifold-auth-token>` component as well. Which means… should we be listening for changes? Is there anything else happening here we need to reserve?

dangodev: @sslotsky

sslotsky: We still need this code. Just kill the oauthUrl check. This subscription is how we refresh the oauth token.

sslotsky: Oh I see. Sorry I need to think that over

sslotsky: Maybe shadowcat can accept a `key` prop. When that value changes it can refresh the iframe.

dangodev: Updated!

PR #483, opened 2019-09-06
Request all pages from a GraphQL connection

sslotsky:
## Reason for change

For our initial GraphQL conversion, we'll handle pagination by eagerly fetching all pages of a connection. This PR introduces a function that allows this to be done generically.

## Testing

Tests should pass.

## Checklist

### DIFF
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -1,4 +1,5 @@
 import { h, Component } from '@stencil/core';
+import '../../utils/fetchAllPages';
 
 import logger from '../../utils/logger';
 
--- a/src/utils/fetchAllPages.spec.ts
+++ b/src/utils/fetchAllPages.spec.ts
@@ -251,7 +251,13 @@ describe('Fetching all pages of a GraphQL connection', () => {
     const aggregator = agg<CategoryEdge>();
     const aggSpy = jest.spyOn(aggregator, 'write');
 
-    await fetchAllPages(query, { first: 3, after: '' }, aggregator, (q: Query) => q.categories);
+    await fetchAllPages({
+      query,
+      nextPage: { first: 3, after: '' },
+      agg: aggregator,
+      getConnection: (q: Query) => q.categories,
+    });
+
     expect(fetchMock.calls).toHaveLength(aggSpy.mock.calls.length);
     expect(aggregator.entries()).toEqual(
       firstPage.categories.edges.concat(secondPage.categories.edges)
--- a/src/utils/fetchAllPages.spec.ts
+++ b/src/utils/fetchAllPages.spec.ts
@@ -1,0 +1,260 @@
+import { gql } from '@manifoldco/gql-zero';
+import fetchMock from 'fetch-mock';
+import { Query, CategoryEdge } from '../types/graphql';
+import fetchAllPages, { createAggregator as agg } from './fetchAllPages';
+
+const query = gql`
+  query CATEGORIES($first: Int!, $after: String!) {
+    categories(first: $first, after: $after) {
+      edges {
+        node {
+          label
+          products(first: 10) {
+            edges {
+              node {
+                label
+              }
+            }
+            pageInfo {
+              hasNextPage
+              endCursor
+            }
+          }
+        }
+      }
+      pageInfo {
+        hasNextPage
+        endCursor
+      }
+    }
+  }
+`;
+
+const firstPage = {
+  categories: {
+    edges: [
+      {
+        node: {
+          label: 'ai-ml',
+          products: {
+            edges: [
+              {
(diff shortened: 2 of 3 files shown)


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -1,4 +1,5 @@
 import { h, Component } from '@stencil/core';
+import '../../utils/fetchAllPages';
```
sslotsky: Uh whoops remove this import :smile:

src/utils/fetchAllPages.ts
```diff
@@ -0,0 +1,50 @@
...
+    throw new Error(`Could not fetch all pages of query: ${query}`);
+  }
+
+  if (page.data) {
+    const { edges, pageInfo } = getConnection(page.data);
+    agg.write(edges);
+
+    if (pageInfo.hasNextPage) {
+      const next = { first: nextPage.first, after: pageInfo.endCursor || '' };
+      await fetchAllPages(query, next, agg, getConnection);
```
dangodev: It’d be nice if there would be a way to parallelize this, but I guess we don’t know if there is a next page or not until fetching. Still, it would be neat to have some kind of first call where it retrieves all the pages, then everything can fetch in parallel. Which would slow down single-page calls but dramatically speed up 3+ pages.

sslotsky: Let's use a reasonably big page size at first. Then if it gets to the point where we're requesting 3+ pages, it should be pretty evident that we have legitimate need to implement lazy loading or something.

src/utils/fetchAllPages.ts
```diff
@@ -0,0 +1,50 @@
...
+interface NextPage {
+  first: number;
+  after: string;
+}
+
+export default async function fetchAllPages<Edge>(
+  query: string,
+  nextPage: NextPage = { first: 25, after: '' },
+  agg: PageAggregator<Edge>,
+  getConnection: (q: Query) => Connection<Edge>
```
dangodev: What does this parameter do?

src/utils/fetchAllPages.ts
```diff
@@ -0,0 +1,50 @@
...
+
+interface NextPage {
+  first: number;
+  after: string;
+}
+
+export default async function fetchAllPages<Edge>(
+  query: string,
+  nextPage: NextPage = { first: 25, after: '' },
+  agg: PageAggregator<Edge>,
```
dangodev: See other question about if we can just return the aggregator rather than require it, but also: should these parameters be an object? If you have to specify the third parameter every time, then the 2nd param’s default won’t do anything

src/utils/fetchAllPages.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+      .once(
+        'https://api.manifold.co/graphql',
+        {
+          status: 200,
+          body: { data: secondPage },
+        },
+        { overwriteRoutes: false }
+      );
+
+    const aggregator = agg<CategoryEdge>();
```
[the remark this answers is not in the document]
sslotsky: It probably could if it a) used looping instead of recursion or b) used an outer function with a recursive inner function. My journey simply led me to this though, and I'm not unhappy with it so I didn't elect to change it.

src/utils/fetchAllPages.ts
sslotsky: I found it difficult for `fetchAllPages` to access the `CategoryConnection` object without doing some unsafe typing. I was sort of solving this out loud in Slack yesterday so you can see what led up to this https://manifoldco.slack.com/archives/CDW445F53/p1567721666418400

src/utils/fetchAllPages.ts
sslotsky: Yeah an object would be fine.

src/utils/fetchAllPages.spec.ts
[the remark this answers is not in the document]
dangodev: Gotcha. Personally I think this approach kept the `fetchAllPages` function as clean and readable as it could be. Trying to go that route would make it messier it would seem.

But in usage I worry that having to import 2 things and set them up properly is a lot of memorization and copy/pasting. Using it would frustrate me long-term, especially since this differs so much from `graphqlFetch` being a one-liner.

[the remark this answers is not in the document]
sslotsky: In practice I don't really think this will be an issue. You don't actually have to import `createAggregator` but it is probably more convenient than writing your own. And we're talking three lines of code instead of one, for a solution that's temporary.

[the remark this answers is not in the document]
dangodev: Yeah that’s fine. I’d rather get real experience with it than imagine how I’d feel. Thanks for talking this through! 🙂

PR #486, opened 2019-09-09
Return records from fetchAllPages

sslotsky:
## Reason for change

This is an improvement on the `fetchAllPages` implementation. This function originally wrote each page to an aggregator, rather than returning all the pages itself. This change allows the function to return the data so that the aggregator isn't needed.

## Testing

Tests should pass.

## Checklist

### DIFF
--- a/src/utils/fetchAllPages.spec.ts
+++ b/src/utils/fetchAllPages.spec.ts
@@ -233,6 +233,8 @@ const secondPage = {
 };
 
 describe('Fetching all pages of a GraphQL connection', () => {
+  afterEach(fetchMock.reset);
+
   it('fetches all the pages', async () => {
     fetchMock
       .once('https://api.manifold.co/graphql', {
--- a/src/utils/fetchAllPages.spec.ts
+++ b/src/utils/fetchAllPages.spec.ts
@@ -1,7 +1,7 @@
 import { gql } from '@manifoldco/gql-zero';
 import fetchMock from 'fetch-mock';
 import { Query, CategoryEdge } from '../types/graphql';
-import fetchAllPages, { createAggregator as agg } from './fetchAllPages';
+import fetchAllPages from './fetchAllPages';
 
 const query = gql`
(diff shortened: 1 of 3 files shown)


sslotsky:
As discussed @dangodev !

src/utils/fetchAllPages.ts
```diff
@@ -36,22 +30,15 @@ export default async function fetchAllPages<Edge>({
 
   if (page.data) {
     const { edges, pageInfo } = getConnection(page.data);
-    agg.write(edges);
 
     if (pageInfo.hasNextPage) {
       const next = { first: nextPage.first, after: pageInfo.endCursor || '' };
-      await fetchAllPages({ query, nextPage: next, agg, getConnection });
+      const remaining = await fetchAllPages({ query, nextPage: next, getConnection });
+      return edges.concat(remaining);
```
davidleger95: This might just be my personal preference but to me `return [...edges, ...remaining]` is more readable. 🤷‍♂ `concat` just makes me think of strings for some reason so it took me a few seconds to realize what was happening. I might be alone in my mild confusion. Not a major thing, so if you prefer this style, you can leave it as is.

sslotsky: Ha, that's funny, some people really hate the spread operator! In terms of syntax I'm fine with either way but I think there would be a performance cost to destructure the arrays. I'll probably leave it as is :smile:

src/utils/fetchAllPages.spec.ts
```diff
@@ -248,19 +250,13 @@ describe('Fetching all pages of a GraphQL connection', () => {
         { overwriteRoutes: false }
       );
 
-    const aggregator = agg<CategoryEdge>();
-    const aggSpy = jest.spyOn(aggregator, 'write');
-
-    await fetchAllPages({
+    const entries = await fetchAllPages({
```
dangodev: 👍

PR #489, opened 2019-09-09
Convert manifold-marketplace to GraphQL

sslotsky:
## Reason for change

Because we are converting all our components to GraphQL! 

Note: this does not use the categories query yet. Instead it uses the products query to maintain a similar data shape for `marketplace-grid` to work with. I have some code comments in the PR that explain my long term vision for using the categories query to drive the UX.

## Testing

Marketplace grid should still work in Storybook, docs, and Render dash. 

## Checklist


src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -1,16 +1,81 @@
...
+      setupStepsHtml: '',
+      state: ProductState.Available,
+      supportEmail: skel.body.support_email,
+      termsUrl: '',
+      valueProps: [],
+      valuePropsHtml: '',
+      categories: skel.body.tags
+        ? skel.body.tags.map(t => ({
+            label: t,
+            products: {
```
sslotsky: This is a pain point. Our queries don't ask for the product connection here, but our TypeScript types indicate that it's required. We plan to look into [using documents in our type generation process](https://graphql-code-generator.com/docs/getting-started/documents-field#document-scanner) to produce more accurate types, but for now we'll have to provide default values like this in some cases.

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -184,16 +217,16 @@ export class ManifoldMarketplaceGrid {
...
+  private renderServiceCard = (product: ProductEdge) => (
     <manifold-service-card-view
-      description={product.body.tagline}
-      isFeatured={this.featured && this.featured.includes(product.body.label)}
-      isFree={this.freeProducts && this.freeProducts.includes(product.id)}
-      logo={product.body.logo_url}
-      name={product.body.name}
+      description={product.node.tagline}
+      isFeatured={this.featured && this.featured.includes(product.node.label)}
+      isFree={product.node.plans ? product.node.plans.edges.length > 0 : false}
```
sslotsky: This isn't great either. I know that the query that I'm using is only requesting _free_ plans, which is why I can say that `isFree` should be true if there are any plans present. But if the query were to change so that it asks for all plans instead of free ones, this logic would break!

GraphQL queries allow field aliases, so I can write the query like this:

```
        freePlans: plans(first: 1, free: true) {
          edges {
            node {
              free
            }
          }
        }
```

But TypeScript won't know that the return object has a field called `freePlans`! This is another thing that using documents for our type generation could help with.

dangodev: 🤔 yeah this seems like something we need. Didn’t know about this!

docs/docs/advanced/custom-loaders.md
```diff
@@ -74,7 +74,9 @@ be [lazy-loaded][lazy-loading] using `import()`).
 ## Currently supported components
 
 - [manifold-plan-selector][plan-selector]
+- [manifold-marketplace][marketplace]
```
dangodev: Can this be alphabetized, please? 🙏

docs/docs/advanced/custom-loaders.md
```diff
@@ -74,7 +74,9 @@ be [lazy-loaded][lazy-loading] using `import()`).
...
 
 [skeletons]: https://css-tricks.com/building-skeleton-screens-css-custom-properties/
 [plan-selector]: /components/plan-selector
+[marketplace]: /components/manifold-marketplace
```
dangodev: Same here—alphabetical 🙏

src/components/manifold-marketplace-grid/manifold-marketplace-grid-happo.ts
```diff
@@ -1,23 +0,0 @@
...
-
-export const allProducts = () => {
-  const grid = document.createElement('manifold-marketplace-grid');
-  grid.services = fromJSON(products);
-  grid.freeProducts = [];
-
-  document.body.appendChild(grid);
-
-  return grid.componentOnReady();
-};
```
dangodev: 💯 💯 💯

src/components/manifold-marketplace-grid/manifold-marketplace-grid.e2e.ts
```diff
@@ -59,42 +47,41 @@ describe('<manifold-marketplace-grid>', () => {
     const serviceCards = await page.findAll(
       'manifold-marketplace-grid >>> manifold-service-card-view'
     );
-    expect(serviceCards).toHaveLength(services.length);
+
+    const labels = await Promise.all(serviceCards.map(c => c.getProperty('productLabel')));
```
dangodev: I recently learned about `.allSettled`, which includes rejected ones, too. Should we use that instead?

davidleger95: `allSettled` has poor browser support, unless Stencil polyfills this for us or we polyfill it? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -1,15 +1,51 @@
...
+ * if it's not. The filter would be debounced, and if needed, we can cache results
+ * using the filter value as a key. In this view, I could see the category menu
+ * either being disabled or remaining enabled. If it is enabled, clicking a
+ * different category would continue to present the 'all-products' interface for
+ * the new category.
+ *
+ * Filtering within the category view would also submit to the API, but it would
+ * not submit a category label. This will present products as it does today, but
+ * will also need to fetch more results when the user scrolls if more pages remain.
+ */
```
dangodev: 👍

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
```diff
@@ -64,13 +100,10 @@ export class ManifoldMarketplaceGrid {
...
 
     // Iterate through services, only add unique categories
-    this.filteredServices.forEach(({ body: { tags } }: Catalog.Product) => {
-      if (!Array.isArray(tags)) {
-        return;
-      }
-      tags.forEach(tag => {
-        if (!categoryList.includes(tag)) {
-          categoryList.push(tag);
+    this.filteredServices.forEach(({ node }: ProductEdge) => {
```
dangodev: Why do you have to redeclare a type here?

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
davidleger95: What about defining the type as: `Product & { freePlans: PlanEdge[] }`? Not the best longterm solution but it works 🤷‍♂ (I think it works, I didn't test it :P )

src/components/manifold-marketplace-grid/manifold-marketplace-grid.e2e.ts
sslotsky: In this case I wouldn't worry about it anyway. I'm just getting properties from elements that we already know are present on the page.

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
sslotsky: That could work as a temporary solution, I'll check into it. Thanks!

src/components/manifold-marketplace-grid/manifold-marketplace-grid.tsx
sslotsky: It turns out I don't, thanks :smile:

PR #491, opened 2019-09-10
MVP for Digital Ocean auth refresh

sslotsky:
## MVP

Digital Ocean won't be using OAuth. Instead, they're going to give us an auth token to use, and either a function or a URL to call when it needs to be refreshed (I've heard both, and I'm not sure which one we're supposed to go with at this time).

This MVP is an example of how this might look. Two main things are happening here:

1. If they're not using oauth, we'll skip rendering the iframe.
1. If they provide a refresh function that returns a `Promise<AuthToken>`, then as soon as we detect that the auth token has been unset, we'll call that function to get a new auth token.

One point worth discussing is that I'm considering moving all of this into `manifold-connection` and deleting the `manifold-auth-token` component, for two reasons: 

1. I think it's fine to try and authenticate unconditionally when you're using `manifold-connection`.
1. It's easier for us if the initial auth token is provided to `manifold-connection` along with any configuration options we may have to support. Then we can easily know when all available configuration values have been provided.

Please feel free to blow holes in this proposal.

## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -13,6 +13,8 @@ export class ManifoldAuthToken {
   @Prop() subscribe?: (s: Subscriber) => () => void = connection.subscribe;
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
+  @Prop() useOauth?: boolean;
+  @Prop() refresh?: () => Promise<AuthToken>;
   @State() tick?: string;
   @Event({ eventName: 'manifold-token-receive', bubbles: true })
   manifoldOauthTokenChange: EventEmitter<{ token: string }>;
@@ -28,8 +30,12 @@ export class ManifoldAuthToken {
     if (this.subscribe) {
       this.unsubscribe = this.subscribe((oldToken?: string, newToken?: string) => {
         if (oldToken && !newToken) {
-          // changing this to any new string will cause a token refresh. getTime() does that wonderfully.
-          this.tick = new Date().getTime().toString();
+          if (this.useOauth) {
+            // changing this to any new string will cause a token refresh. getTime() does that wonderfully.
+            this.tick = new Date().getTime().toString();
+          } else if (this.refresh) {
+            this.refresh().then(this.setToken);
+          }
         }
       });
     }
@@ -47,8 +53,11 @@ export class ManifoldAuthToken {
     }
   }
 
-  setInternalToken = (e: CustomEvent) => {
-    const payload = e.detail as AuthToken;
+  setInternalToken = (e: CustomEvent<AuthToken>) => {
+    this.setToken(e.detail);
+  };
+
+  setToken = (payload: AuthToken) => {
     if (!payload.error && payload.expiry) {
       const formattedToken = `${payload.token}|${payload.expiry}`;
       if (this.setAuthToken) {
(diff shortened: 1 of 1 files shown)


src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -13,6 +13,8 @@ export class ManifoldAuthToken {
   @Prop() subscribe?: (s: Subscriber) => () => void = connection.subscribe;
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
+  @Prop() useOauth?: boolean;
+  @Prop() refresh?: () => Promise<AuthToken>;
```
dangodev: Should this be a `@Method` instead?

sslotsky: Nope. As the PR description states, they will either provide us with a function to retrieve the new token, or a URL for us to call. I'm not sure which, but this MVP assumes that they'll give us a function.

dangodev: Oh right. Hm. With a function, though, we can only attach that as a prop after the component has rendered and already started loading. I imagine this would almost always miss `componentWillLoad` because it’s not a primitive.

I’d like to use `oauth-url` as an attribute instead, and see if we can do without this (or maybe we can come up with some way to fire this if it’s added after the component loads).

dangodev:
What would their consumer code look like? Something like this?

```html
<manifold-connection token={myToken} oauth-url="https://gettoken.com">
```

src/components/manifold-auth-token/manifold-auth-token.tsx
```diff
@@ -60,6 +69,10 @@ export class ManifoldAuthToken {
 
   @logger()
   render() {
-    return <manifold-oauth tick={this.tick} onReceiveManifoldToken={this.setInternalToken} />;
+    return (
+      this.useOauth && (
```
dangodev: When we’re not rendering the iframe, we’ll need to render `null`, I think (not sure if `false` will render as text or not)

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: > With a function, though, we can only attach that as a prop after the component has rendered and already started loading

Two things:

1. I don't think that's accurate. The outer node can be, and often is, present in the DOM before the component has started loading or rendering. 
1. This function doesn't need to be present on load anyway. It only needs to be there after we receive a 401, which causes the token to be unset, which then causes this component to ask for a refresh.

src/components/manifold-auth-token/manifold-auth-token.tsx
sslotsky: I considered this and will make sure to test that.

src/components/manifold-auth-token/manifold-auth-token.tsx
dangodev: > It only needs to be there after we receive a 401
Oh right! Because this is within the subscription. I get it now. 👍

sslotsky:
> What would their consumer code look like? Something like this?
> 
> ```
> <manifold-connection token={myToken} oauth-url="https://gettoken.com">
> ```

If we do wind up using a URL then that's probably close, although it's not really OAuth so I think the attribute would be called `refresh-url` or something.

src/components/manifold-auth-token/manifold-auth-token.tsx
dangodev: > The outer node can be, and often is, present in the DOM before the component has started loading or rendering.

Yeah I guess that’s _possible_; I was mostly just thinking about React’s `useRef` and whether or not `ref.current` will always be accessible _before_ `componentWillLoad` in Stencil, and wasn’t sure if that 100% will always fire every time. But like you said, it’s within the subscription anyway so even if that were the case we’d be fine. Just wanted to talk it through.

sslotsky: That's all worth testing, but I'm pretty sure that `ref.current` being accessible really just means that the node has been created. It doesn't mean that it's been attached to the DOM or that the web component has loaded or rendered.

PR #495, opened 2019-09-11
Allow manual authentication

sslotsky:
## Reason for change

Some platforms will want to authenticate and refresh manually, completely skipping the OAuth process. Platforms will provide an initial token and then listen for the `manifold-token-clear` event, which will allow them to fetch a new token and reassign it to the web component.

Example platform usage (vanilla):

```html
<manifold-auth-token auth-type="manual" token="my-token"></manifold-auth-token>
```

```js
const component = document.querySelector('manifold-auth-token');
component.addEventListener('manifold-token-clear', async () => {
  const token = await myGetTokenFunction();
  component.token = token;
});
```

## Testing

Automated tests should pass. 

## Checklist

- [x] **CHANGELOG**: The **Unreleased** section of CHANGELOG was updated
- [ ] **Prop changes**: [docs][docs] were updated
- [ ] **Prop changes**: E2E tests were updated, testing from the highest level possible
- [ ] **Platform testing**: If this change should be [tested against a platform][platform-testing], has it been?

[docs]: https://ui.sandbox.manifold.co
[platform-testing]: https://app.gitbook.com/@manifold/s/engineering/playbooks/testing-your-code-against-a-platform

PR #496, opened 2019-09-11
MVP for manual auth

sslotsky:
## MVP

Some platforms won't be using OAuth. Instead, they're going to give us an auth token to use, and either a function or a URL to call when it needs to be refreshed (I've heard both, and I'm not sure which one we're supposed to go with at this time).

This MVP is an example of how this might look. Two main things are happening here:

1. If they're not using oauth, we'll skip rendering the iframe.
1. If they provide a refresh function that returns a `Promise<AuthToken>`, then as soon as we detect that the auth token has been unset, we'll call that function to get a new auth token.

Example usage (vanilla:)

```html
<manifold-auth-token auth-type="manual" token="my-token"></manifold-auth-token>
```

```js
const component = document.querySelector('manifold-auth-token');
component.token = myTokenValue;
component.refresh = async () => {
  return await myGetTokenFunction();
};
```

One point worth discussing is that I'm considering moving all of this into `manifold-connection` and deleting the `manifold-auth-token` component, for two reasons: 

1. I think it's fine to try and authenticate unconditionally when you're using `manifold-connection`.
1. It's easier for us if the initial auth token is provided to `manifold-connection` along with any configuration options we may have to support. Then we can easily know when all available configuration values have been provided.
## Checklist

### DIFF
--- a/src/components/manifold-auth-token/manifold-auth-token.tsx
+++ b/src/components/manifold-auth-token/manifold-auth-token.tsx
@@ -13,6 +13,8 @@ export class ManifoldAuthToken {
   @Prop() subscribe?: (s: Subscriber) => () => void = connection.subscribe;
   /* Authorisation header token that can be used to authenticate the user in manifold */
   @Prop() token?: string;
+  @Prop() useOauth?: boolean;
+  @Prop() refresh?: () => Promise<AuthToken>;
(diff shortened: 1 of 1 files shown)


sslotsky:
Preferring #495

PR #501, opened 2019-09-12
Emit time-to-render event after first render without skeleton elements

markjamieson:
## Reason for change

This change emits a `manifold-time-to-render` event after a `<manifold-*>` component's first render without skeleton elements. This represents the first "meaningful" render after data loads and the component is displaying useful information.

To start the time-to-render lifecycle, I'm relying on a mutation observer set up in `<manifold-performance>` to monitor the subtree for any added nodes that start with `manifold-`. When a new node is added, a `data-start` attribute is added to that node to represent the initialization time.

To monitor the end of the time-to-render lifecycle, I'm relying on the existing `logger` decorator that is required on all our elements. I've altered the decorator to examine the rendered output for skeleton elements, and if none exist, I add a `data-end` attribute to the element and emit the event. This part relies on the component being measured to have `@Element() el: HTMLElement;` in the component definition. I'll be seeking advice on making that a required part of our components by lint rule.

TODO in followup PRs:
1. Observe subtree additions within shadowRoots. Currently when a component has a shadowRoot the mutation observer can't observe within the shadow DOM. A follow up is needed to attach those shadowRoot components to the observer (or something like that, if I knew exactly what to do I'd just do it).
1. The lint rule mentioned above

## Testing

The component rendered in the Performance story in Storybook should now log the `manifold-time-to-render` event to DataDog

## Checklist


src/components/manifold-performance/manifold-performance.tsx
```diff
@@ -54,13 +57,31 @@ export class ManifoldPerformance {
     }
   };
 
+  mutationCallback = (mutationsList: [MutationRecord]) => {
+    mutationsList.forEach(mutation => {
+      mutation.addedNodes.forEach((node: HTMLElement) => {
+        if (node.localName && node.localName.startsWith('manifold-')) {
+          /* eslint-disable no-param-reassign */
+          node.dataset.start = String(performance.now());
```
dangodev: How do we know this won’t fire multiple times for each component, each time it Renders?

src/utils/logger.tsx
```diff
@@ -24,7 +39,24 @@ export default function logger<T>() {
...
+          this.el.dataset.start &&
+          !this.el.dataset.end &&
+          !hasSkeletonElements(rendered)
+        ) {
+          this.el.dataset.end = performance.now();
+          const evt = new CustomEvent('manifold-time-to-render', {
+            bubbles: true,
+            detail: {
+              component: target.constructor.name,
+              duration: this.el.dataset.end - this.el.dataset.start,
```
dangodev: I think `dataset.start` will always be a string?

dangodev: It seems like TypeScript doesn’t know that `this.el` is an `HTMLElement`

src/components/manifold-performance/manifold-performance.tsx
dangodev: Hm. Actually if you test it on `<manifold-plan-selector>`, it seems to get blown away by Stencil if multiple renders happen. I think everything else about this approach is great, but specifically we may need to use something other than a `data-*` attribute to hold this number.

sslotsky: This seems like a difficult problem. I'm going to think out loud here so don't take any of this as an official suggestion.

Ideally, we'd be hooking into `componentOnReady()` to determine the timestamp for the first render. If we're using skeletons though, then that won't tell us when we've rendered with all of our data. We _do_ still know when the data has come back in this case, because we're reporting it from our fetch calls. But we don't know what the delay is between retrieving the data and having the component render with that data, and if we're just hooking into `render()` then I share Drew's concern that it fires more often than it should.

sslotsky: I'll add this: I've been scheming on a new way to leverage data components. I haven't come up with a working prototype yet, but my ideal solution would have this library being built off of data components that always do an async `componentWillLoad()`, so that they aren't considered ready until they've "rendered" with data. I put "rendered" in quotes because these components ideally wouldn't even display anything, which is the more interesting part of this.

Traditionally, we've had components that fetch the data rendering presentational components that display that data. What if we flipped this on its head? What if our display components could _consume_ the data components, and have the option to display skeletons while they wait? I feel like the idea is promising but in practice I'm having a hard time with it, because in Stencil, parent components actually won't display anything until every component that they render has fired `componentOnReady()`. So the course I've tried to take may be a dead end, but I just wanted to share where my head is right now, because I think solving the problem this way could really help with the problem that this PR attempts to solve.

src/utils/logger.tsx
markjamieson: You're right, but for some ridiculous reason JS allows you to do math on strings.

src/components/manifold-performance/manifold-performance.tsx
markjamieson: This approach was because I did not want to add a `componentOnReady` decorator to every single component. It's a fine approach with the logger decorator but I don't love the idea of requiring yet another decorator for a method that we don't even define on most components. Ideally we could use a class decorator, but Stencil doesn't allow that.

markjamieson: Also the MutationObserver should only observe the component when it's added to the DOM, not on every render. With the data attr getting blown away though, I'll have to work around that.

dangodev: Ah that makes sense! So this probably won’t fire more than once.

But I think what you said made me think of another scenario: what if the consumer is using pure HTML, and the component is in the DOM tree at the same time as `<manifold-performance>`? I think we may have to hook into `componentWillLoad` regardless. But we’re back at that classic problem of how to centralize that somehow.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Added a time to first meaningful render event to be consumed by `<manifold-performance>` component
```
dangodev: This should be under a section (maybe “Added”)

src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -42,6 +43,9 @@ export class ManifoldActivePlan {
     return this.plans.find(({ id }) => id === this.selectedPlanId);
   }
 
+  @loadMark()
```
dangodev: Are you able to add this as an option to the ESLint plugin, so that new components would require this? Or would we need a PR on our lint plugin to be able to do that? cc @sslotsky

sslotsky: We'd need to add a new rule to the plugin.

markjamieson: Needs to be a new rule. We could potentially modify the existing rule to take a list of pairs <methodName, requiredDecoratorName>

But we also need a new rule to require `componentWillLoad` to be defined in the first place

dangodev:
Screen Shot 2019-09-18 at 09 29 20

This is so cool! 🎉

markjamieson:
@sslotsky Is it fair to say that any component that does load data does it in `componentWillLoad`? If so, that's a simplification I'm happy with. We never load our data in `connectedCallback` for example?

Also, do we have any components where render time could be negatively affected by anything other than async data?

PR #508, opened 2019-09-13
Make a separate query for free plans until backend query can be fixed

sslotsky:
## Reason for change

The logic to determine whether a product had free plans was broken, because the backend doesn't know if the plan is free until after it's retrieved it from the database. We therefore need - at least for now - to make separate queries for all the plans in order to determine if any of them are free.

## Testing

In the storybook site, the products that show up with a FREE badge should be the ones that actually have free plans.

## Checklist

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -25,12 +25,14 @@ import {
   Marketplace,
 } from './types/marketplace';
 import {
-  Product,
   ProductEdge,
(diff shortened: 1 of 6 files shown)


sslotsky:
We decided to hardcode free plans for now, until the API can be fixed, which has been prioritized.

PR #515, opened 2019-09-17
Allow configurations through manifold-connection

sslotsky:
## Reason for change

The `manifold-connection` was supposed to allow us to configure certain things, like which environment to point to, or how long to wait before timing out. This became broken along the way as we wrestled with state tunnels and race conditions.

This change restores configurability. To accomplish this, the `manifold-connection` calls an `initialize` function on the connection when it loads. The `restFetch` and `graphqlFetch` functions will wait for the connection state to be initialized before proceeding.

The consequence of this is that no component can make network calls unless the `manifold-connection` is present on the page. This is how we originally intended it to work, so this is fine.

In the future, the `initialize` function will also accept the auth token, which will similarly be passed along by `manifold-connection`, allowing us to eliminate the `manifold-auth-token` component.

## Testing

Tests should still pass, and now you should be able to set `env="stage"` on `manifold-connection` and have network calls go to staging.

## Checklist

### DIFF
--- a/src/state/connection.ts
+++ b/src/state/connection.ts
@@ -6,15 +6,46 @@ const baseWait = 15000;
 
 export type Subscriber = (oldToken?: string, newToken?: string) => void;
 
+const INITIALIZED = 'manifold-connection-initialize';
+
+interface Initialization {
+  env?: 'local' | 'stage' | 'prod';
+  waitTime?: number;
+  authToken?: string;
+}
+
 export class ConnectionState {
-  /** _(optional)_ Specify `env="stage"` for staging */
+  authToken?: string;
   env: 'local' | 'stage' | 'prod' = 'prod';
-  /** _(optional)_ Wait time for the fetch calls before it times out */
   waitTime: number = baseWait;
 
-  authToken?: string;
-
   private subscribers: Subscriber[] = [];
+
+  private initialized: boolean = false;
+
+  initialize = ({ env = 'prod', waitTime = baseWait }: Initialization) => {
+    this.env = env;
+    this.waitTime = waitTime;
+    this.initialized = true;
+    const event = new CustomEvent(INITIALIZED);
+    document.dispatchEvent(event);
+  };
+
+  onReady = () => {
+    return new Promise(resolve => {
+      if (this.initialized) {
+        resolve();
+      }
+
+      const done = () => {
+        this.initialized = true;
+        document.removeEventListener(INITIALIZED, done);
+        resolve();
+      };
+
+      document.addEventListener(INITIALIZED, done);
+    });
+  };
 
   subscribe = (s: Subscriber) => {
     this.subscribers.push(s);
@@ -22,14 +53,6 @@ export class ConnectionState {
     return () => {
       this.subscribers.splice(this.subscribers.indexOf(s), 1);
     };
-  };
-
-  setEnv = (newValue: 'local' | 'stage' | 'prod') => {
-    this.env = newValue;
-  };
-
-  setWaitTime = (newValue: number) => {
-    this.waitTime = newValue;
   };
 
   setAuthToken = (token: string) => {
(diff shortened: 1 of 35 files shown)


stories/scenarios.stories.js
```diff
@@ -22,7 +22,7 @@ storiesOf('Scenarios', module)
   .addDecorator(manifoldConnectionDecorator)
   .add('OAuth', () => {
     return `
-      <manifold-connection>
+      <manifold-connection env="stage">
```
sslotsky: I was just doing this to test, probably want to change it back.

stories/data/connectionDecorator.js
```diff
@@ -6,7 +6,7 @@ export const manifoldConnectionDecorator = storyFn => {
   const token = text('manifold_api_token', localStorage.getItem('manifold_api_token') || '');
   localStorage.setItem('manifold_api_token', token); // update localStorage to persist
   return `
-  <manifold-connection>
+  <manifold-connection env="stage">
```
sslotsky: This was just for testing, should probably be changed back.

src/state/connection.ts
```diff
@@ -49,18 +72,20 @@ export class ConnectionState {
...
     retries: 1,
   });
 
   graphqlFetch = createGraphqlFetch({
     getAuthToken: this.getAuthToken,
     setAuthToken: this.setAuthToken,
-    endpoint: connections[this.env].graphql,
-    wait: this.waitTime,
+    onReady: this.onReady,
+    endpoint: () => connections[this.env].graphql,
```
dangodev: A function makes a lot more sense! Dunno why we didn’t do it this way before.

src/utils/connections.ts
```diff
@@ -20,6 +21,7 @@ export const connections: Connections = {
     provisioning: 'http://api.provisioning.arigato.tools/v1',
     connector: 'http://api.connector.arigato.tools/v1',
     graphql: 'http://graphql.arigato.tools/graphql',
+    oauth: 'https://login.arigato.tools/signin/oauth/web', // does this work?
```
dangodev: @tainacleal is this correct?

stories/data/connectionDecorator.js
dangodev: I mean, if we’re developing on `stage` in dev, maybe we should keep it? Assuming we get better stage data (that’s still in the backlog, but I still see value in it)

dangodev:
Suggestion: would it be possible to add a `select` Storybook Knob to `<manifold-marketplace>` to select the env? I’d like to be able to at least switch between listing stage and prod products to make sure this is working.

stories/data/connectionDecorator.js
dangodev: Actually on second thought, it is really good to be able to look specifically at the plan selector’s specific product displays, since we don’t have good staging data yet that shows similar products.

We should change this back, unless all the **catalog** stories are updated with the correct product & plan labels everywhere:

Screen Shot 2019-09-17 at 10 10 13

But I still think adding a dropdown to the marketplace component for env would be helpful.

stories/scenarios.stories.js
domenicrosati: ya i'd say

stories/data/connectionDecorator.js
domenicrosati: my 2c's: prod is good because it'd be what platforms will dev against?

src/utils/connections.ts
domenicrosati: it sure is: https://github.com/manifoldco/marketplace/blob/c8ffaa2e93b80f7651358d823b0611bdfec4027d/docker-compose.yml#L716

domenicrosati: and it works!

sslotsky: :tada:

PR #520, opened 2019-09-18
Add owner-id helper

dangodev:
## Reason for change

Screen Shot 2019-09-18 at 11 21 32

This is a small timesaver for the resource **provision** story.

Before, in order to test **provision**, you had to go look up your own user ID, which was a pain. This adds a call to Manifold’s identity service to grab your user ID for you automatically from your token, and save it in localStorage.

## Testing

1. Add a valid token to **manifold_api_token** from the Manifold Dashboard in Storybook. Refresh.
1. Go to the **Resource > provision** story in Storybook, and see that the `owner-id` is set for you (may take a second refresh if you start out on the provision story).

## Checklist


stories/data/connectionDecorator.js
```diff
@@ -7,6 +7,18 @@ export const manifoldConnectionDecorator = storyFn => {
   const options = { Production: 'prod', Staging: 'stage' };
   const env = radios('env', options, 'prod');
   localStorage.setItem('manifold_api_token', token); // update localStorage to persist
+
+  // grab user ID from Manifold (we need this in other stories)
+  if (token && !localStorage.getItem('manifold_user_id')) {
+    fetch('https://api.identity.manifold.co/v1/self', {
```
domenicrosati: identity only really works for manifold users... something for consideration is, for platform users, to use:
```
  query ProfileId {
    profile {
      id
    }
  }
```
instead. thought i am not certain that works for manifold users!

dangodev: Ohh shoot. Lemme try that actually

src/types/graphql.ts
```diff
@@ -1036,6 +1189,40 @@ export type SubLineItemEdge = {
...
+  ownerId?: Maybe<Scalars['ID']>,
+};
+
+/** The payload for a resource whose plan changed. */
+export type UpdateResourcePlanPayload = {
+   __typename?: 'UpdateResourcePlanPayload',
+  /** The updated resource body */
+  data: Resource,
+};
+
```
dangodev: Updated GraphQL schema just because it’s been a while 🤷‍♂

dangodev: Update: reverted this; it broke a test because the mock structure needed changing

PR #521, opened 2019-09-18
Automatically update GraphQL schema for tests

dangodev:
## Reason for change
I noticed the GraphQL schema was a bit out of date, which could cause problems. But how we solve this automatically brings new challenges:

- We could update the schema **before** `stencil build` each time. This has the advantage of early warning if anything changes, but it would slow down _every_ build and basic tasks. Do we want that?
- We could alternately update the schema **after** every `stencil build`. Builds would happen faster, but we wouldn’t catch errors in the build phase, and it’d still slow down whatever task happened right afterward.
- Rather than either of the two above, this PR takes the option of only fetching a new schema during our CI tests. So it won’t slow down local dev, but it may be a bit of a pain if it’s unexpected (still, I suppose it’s better than errors in production).

**What do we want to do?**

## Testing

This should pass tests.

## Checklist


src/spec/mock/graphql.ts
```diff
@@ -124,6 +125,7 @@ export const plan: Plan = {
           __typename: 'PlanFixedFeature',
           displayName: feature.name,
           displayValue: feature.value_string as string,
+          label: feature.label,
```
dangodev: This was now required in the specs

src/spec/mock/graphql.ts
```diff
@@ -206,11 +208,13 @@ export const product: Product = {
   })),
   valuePropsHtml: '',
   setupStepsHtml: '',
+  /*
   integration: {
     __typename: 'ProductIntegration',
     baseUrl: prodMock.body.integration.base_url,
     ssoUrl: prodMock.body.integration.sso_url as string,
   },
+  */
```
dangodev: This is now… gone??

sslotsky:
What steps do we take if CI fails due to a schema update? I'd presume that the developer who made the PR would then have to update the schema locally and push?

dangodev:
Yes that’s the tradeoff for detecting the live schema changed. But updating the cached schema can happen in a separate PR, too

package.json
```diff
@@ -41,8 +41,8 @@
     "generate:specs:marketplace": "npx @manifoldco/swagger-to-ts src/spec/marketplace/v1.yaml -w 'export namespace Marketplace' -o src/types/marketplace.ts",
     "generate:specs:provisioning": "npx @manifoldco/swagger-to-ts src/spec/provisioning/v1.yaml -w 'export namespace Provisioning' -o src/types/provisioning.ts",
     "generate:specs:connector": "npx @manifoldco/swagger-to-ts src/spec/connector/v1.yaml -w 'export namespace Connector' -o src/types/connector.ts",
-    "happo": "npm run build && manifold run -t manifold -p manifold-ui-ci -- happo run",
-    "happo-ci-circleci": "npm run build && happo-ci-circleci",
+    "happo": "npm run build &&  npm run generate:gql && manifold run -t manifold -p manifold-ui-ci -- happo run",
+    "happo-ci-circleci": "npm run build && npm run generate:gql && happo-ci-circleci",
```
dangodev: The obvious possible downside to this is that the code generation fails during our test suite. Buuut this is also our prod endpoint it’s generating it from, so if this fails, then there are probably bigger problems than a failed test.

PR #522, opened 2019-09-19
Increase page size to fetch all products in one go

dangodev:
## Reason for change
Marketplace is fetching in 2 calls rather than one, which is causing it to be slower than the original REST query and double the calls it needs to. Let’s grab all the products in one request.

This PR also adds a patch for the `freePlans` query until that is fixed (https://github.com/manifoldco/engineering/issues/9381).

Lastly, this PR also adds a header to the GraphQL request which dramatically speeds up performance in Chrome.

## Testing
1. Test the `<manifold-marketplace>` component in Storybook. Confirm that:
    - [ ] It only fires one GraphQL request now
    - [ ] the request it does fire is performant in Chrome AND Firefox (`100ms` or less)

## Checklist

### DIFF
--- a/src/components/manifold-marketplace/manifold-marketplace.tsx
+++ b/src/components/manifold-marketplace/manifold-marketplace.tsx
@@ -65,28 +65,33 @@ const productQuery = gql`
   }
 `;
 
-const freePlans = gql`
-  query FREE_PLANS($first: Int!, $after: String!) {
-    products(first: $first, after: $after) {
-      edges {
-        node {
-          label
-          plans(first: $first) {
-            edges {
-              node {
-                free
-              }
-            }
-          }
-        }
-      }
-      pageInfo {
-        hasNextPage
-        endCursor
-      }
-    }
-  }
-`;
+// TODO: fix the freePlans query in GraphQL
+const TEMPORARY_FREE_PRODUCTS = [
+  'blitline',
+  'bonsai-elasticsearch',
+  'cloudamqp',
+  'cloudcube',
+  'custom-recognition',
+  'degraffdb-generators-stage',
+  'elegant-cms',
+  'generic-tagging',
+  'hypdf',
+  'informant',
+  'logdna',
+  'mailgun',
+  'memcachier-cache',
+  'pdfshift',
+  'piio',
+  'posthook',
+  'prefab',
+  'scaledrone',
+  'scoutapp',
+  'till',
+  'timber-logging',
+  'timber-logging-staging',
+  'valence',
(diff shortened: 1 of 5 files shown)


sslotsky:
I'm not sure if we need this because it's included in https://github.com/manifoldco/ui/pull/508/files already, and it seems like we have to go ahead with some variation of that no matter what.

src/components/manifold-marketplace/manifold-marketplace.tsx
```diff
@@ -72,6 +65,34 @@ const productQuery = gql`
...
+  'prefab',
+  'scaledrone',
+  'scoutapp',
+  'till',
+  'timber-logging',
+  'timber-logging-staging',
+  'valence',
+  'websolr',
+];
+
```
dangodev: @sslotsky how dumb is this? Be honest

src/utils/graphqlFetch.ts
```diff
@@ -50,7 +50,11 @@ export function createGraphqlFetch({
 
     const response = await fetch(endpoint(), {
       method: 'POST',
-      headers: { 'content-type': 'application/json', ...auth },
+      headers: {
+        Connection: 'keep-alive',
```
dangodev: So this is a last-minute find in this PR, but I added a test. This dramatically speeds up Chrome, because it doesn’t send this header by default (Firefox does). James said this should be sent.

src/components/manifold-marketplace/manifold-marketplace.tsx
domenicrosati: IMO it depends on how quick were gonna fix this hehehe

src/utils/graphqlFetch.ts
domenicrosati: +1

domenicrosati:
woo 100ms now for me!

src/utils/graphqlFetch.ts
dangodev: Reliability will investigate why that is; it might be a misconfiguration in the load balancer. But for now, it’s a quick win (and because Firefox sends this by default, I don’t consider this “hacky;” we’re just standardizing headers across browsers which is always a good thing).

src/components/manifold-marketplace/manifold-marketplace.tsx
dangodev: Guillaume has already scoped out the work for this. Basically this only needs to be fixed when the new provider dashboard launches and providers can change this on-the-fly.

PR #524, opened 2019-09-19
Use GET for GraphQL

dangodev:
## Reason for change

Screen Shot 2019-09-19 at 16 39 49

Outlined here: https://github.com/manifoldco/engineering/issues/9392

TL;DR our load balancers & CDN will do better if requests are cacheable via URL.

Most of this PR is actually test changes, because our tests were trying to _exactly_ match URL rather than just the beginning domain. They were updated to match the beginning segment of a URL now for GraphQL.

## Testing

⚠️**This affects all the GraphQL queries** so test each component!!!

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- GraphQL now uses GET instead of POST requests (#524)
+
 ## [v0.5.11]
 
 ### Added
--- a/src/utils/gql-min.ts
+++ b/src/utils/gql-min.ts
@@ -1,3 +1,6 @@
+/**
+ *  Minifies a GraphQL string query to its smallest shape
+ */
 export default function gqlMin(query: string) {
   return query
     .trim()
@@ -5,3 +8,19 @@ export default function gqlMin(query: string) {
     .replace(/([^A-z\d])\s+/g, '$1') // remove whitespace trailing special chars (brackets, commas, colons, etc.)
     .replace(/\s+([^A-z\d])/g, '$1'); // remove whitespace leading special chars
 }
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/**
+ * Converts a GraphQL response body into a GET request with URL Params
+ */
+export function gqlSearch(requestBody: any) {
+  const minifiedRequest = Object.entries(requestBody).reduce(
+    (req, [key, value]) => ({
+      ...req,
+      [key]: key === 'query' ? gqlMin(value as string) : JSON.stringify(value),
+    }),
+    {}
+  );
+  return new URLSearchParams(minifiedRequest).toString();
+}
+/* eslint-enable @typescript-eslint/no-explicit-any */
--- a/src/utils/gql-min.ts
+++ b/src/utils/gql-min.ts
@@ -1,0 +1,7 @@
+export default function gqlMin(query: string) {
+  return query
+    .trim()
+    .replace(/[\n\s]+/g, ' ') // remove newlines and indentation
+    .replace(/([^A-z\d])\s+/g, '$1') // remove whitespace trailing special chars (brackets, commas, colons, etc.)
+    .replace(/\s+([^A-z\d])/g, '$1'); // remove whitespace leading special chars
+}
--- a/src/utils/graphqlFetch.spec.ts
+++ b/src/utils/graphqlFetch.spec.ts
@@ -21,12 +21,12 @@ describe('graphqlFetch', () => {
         wait: () => 0,
         /* no endpoint */
       });
-      fetchMock.mock('https://api.manifold.co/graphql', {
+      fetchMock.mock('begin:https://api.manifold.co/graphql', {
         status: 200,
         body: { data: {} },
       });
       await fetcher({ query: '' });
-      expect(fetchMock.called('https://api.manifold.co/graphql')).toBe(true);
+      expect(fetchMock.called('begin:https://api.manifold.co/graphql')).toBe(true);
     });
 
     it('returns data from server', async () => {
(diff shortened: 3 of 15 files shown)


src/utils/gql-min.ts
```diff
@@ -0,0 +1,7 @@
+export default function gqlMin(query: string) {
+  return query
+    .trim()
+    .replace(/[\n\s]+/g, ' ') // remove newlines and indentation
+    .replace(/([^A-z\d])\s+/g, '$1') // remove whitespace trailing special chars (brackets, commas, colons, etc.)
+    .replace(/\s+([^A-z\d])/g, '$1'); // remove whitespace leading special chars
+}
```
dangodev: I looked for a GQL minification library, but I kept finding very heavy ones. This seems to do an OK job.

src/utils/graphqlFetch.spec.ts
```diff
@@ -21,12 +21,12 @@ describe('graphqlFetch', () => {
         wait: () => 0,
         /* no endpoint */
       });
-      fetchMock.mock('https://api.manifold.co/graphql', {
+      fetchMock.mock('begin:https://api.manifold.co/graphql', {
```
dangodev: Fortunately fetchMock had an easy solution for the fact that the URL is now longer 🎉

src/utils/gql-min.ts
dangodev: Screen Shot 2019-09-19 at 16 41 57

src/utils/graphqlFetch.spec.ts
jbowes: strange that it's not just regex. not that i particularly like regex :)

src/utils/gql-min.ts
jbowes: This is going to miss a lot of edge cases, like special chars inside strings, multiline strings, etc. parsing languages is not easy, and so probably why those other libs are so big.

src/utils/graphqlFetch.ts
```diff
@@ -48,14 +49,23 @@ export function createGraphqlFetch({
...
+      let minifiedVal = JSON.stringify(value);
+      if (key === 'query' || key === 'mutation') {
+        minifiedVal = gqlMin(value);
+      }
+      return { ...req, [key]: minifiedVal };
+    }, {});
+    const search = new URLSearchParams(minifiedRequest);
+
+    const response = await fetch(`${endpoint()}?${search.toString()}`, {
+      method: 'GET',
```
jbowes: We still need to POST mutations.

jbowes: And post urls that are too long

src/utils/gql-min.ts
dangodev: `URLSearchParams.toString()` handles the encoding, so we should be fine there. The query is actually encoded; it’s just decoded in the Chrome Network tab for readability

src/utils/graphqlFetch.spec.ts
dangodev: Yeah you can use regex for fetchMock, but I found `begin:` to be fairly readable when doing ``fetchMock.called(`begin:${graphqlEndpoint}`)`` (as opposed to `new RegExp()`)

src/utils/gql-min.ts
dangodev: Also I’m thinking that won’t be an issue with GQL queries? Because we’re only requesting the shape of data, and we are also using our `@manifoldco/gql-zero` library that helps with formatting to keep it in a predictable shape. We should also be using variables as much as possible, further reducing the chance of “weird” strings.

We may run into the issue you described for mutations if, say, we were mutating valuePropsHTML, but that’ll be POST’d, to your other feedback.

Maybe I’m missing something? I don’t disagree there _might_ be edge cases but I’d rather solve them as they arise than imagine them now.

src/utils/graphqlFetch.ts
```diff
@@ -37,28 +46,48 @@ export function createGraphqlFetch({
...
+    };
+
+    // if valid token, add to request
+    const token = getAuthToken();
+    if (token && token !== 'undefined') {
+      options.headers = {
+        ...options.headers, // thank TypeScript for this dumb spread
+        authorization: `Bearer ${token}`,
+      };
+    }
```
dangodev: I found the really long ternary to be confusing when it came to auth, so I made it a bit more readable.

src/utils/graphqlFetch.ts
```diff
@@ -37,28 +46,48 @@ export function createGraphqlFetch({
...
+
+    // if request is GQL query, and it’s short enough, send as GET
+    if ((request as GraphqlQuery).query) {
+      const MAX_LENGTH = 2048; // should this be higher?
+      const queryUrl = `${url}?${gqlSearch(request)}`;
+      if (queryUrl.length < MAX_LENGTH) {
+        options.method = 'GET';
+        delete options.body;
```
dangodev: IDK why; I always have fun when I get to use `delete`. I don’t get to often!

src/utils/graphqlFetch.ts
```diff
@@ -37,28 +46,48 @@ export function createGraphqlFetch({
...
+    if (token && token !== 'undefined') {
+      options.headers = {
+        ...options.headers, // thank TypeScript for this dumb spread
+        authorization: `Bearer ${token}`,
+      };
+    }
+
+    // if request is GQL query, and it’s short enough, send as GET
+    if ((request as GraphqlQuery).query) {
+      const MAX_LENGTH = 2048; // should this be higher?
```
dangodev: From what I could find, all modern browsers are in the tens of thousands. But I’m not sure if a higher # has any impact on our backend. WDYT @jbowes @domenicrosati?

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- GraphQL now uses GET instead of POST requests (#524)
```
jbowes:
```suggestion
- GraphQL now uses GET instead of POST requests  where possible (#524)
```

dangodev:
With James’ feedback, moving to `wip`. Will move the minification logic into @manifoldco/gql-zero and handle more edge cases like comments and “weird“ strings (and of course, use a lightweight library if possible)

dangodev:
Just gonna close this for now. Will take a new utility, and it’ll be easier to redo this branch from scratch.

PR #528, opened 2019-09-20
Fix strategy for checking auth expiration in GraphQL responses

sslotsky:
## Reason for change

We were checking for 401 responses from the GraphQL API to determine if auth had expired. But the GraphQL API doesn't ever send 401 responses. Instead, they send a regular errors array, where at least one of those errors should have an `extensions.type` with a value of `AuthFailed`. This PR updates `graphqlFetch` to check for the right conditions when testing auth expiration.

## Testing

All tests should still pass. For exploratory testing, you could set an expired token in a Storybook story that _only_ calls the GraphQL API (if the story also calls the REST API, you might get a false positive, because the 401 recovery could happen before `graphqlFetch` gets a chance to execute the new logic).

## Checklist

### DIFF
--- a/src/utils/graphqlFetch.spec.ts
+++ b/src/utils/graphqlFetch.spec.ts
@@ -128,20 +128,25 @@ describe('graphqlFetch', () => {
       const body = {
         data: null,
         errors: [
-          { message: 'User does not have permission to access the resource', path: ['resources'] },
+          {
+            message: 'User does not have permission to access the resource',
+            path: ['resources'],
+            extensions: { type: 'AuthFailed' },
+          },
         ],
       };
+
       const fetcher = createGraphqlFetch({
         wait: () => 0,
         endpoint: () => graphqlEndpoint,
         getAuthToken: () => '1234',
       });
 
-      const response = { status: 401, body };
+      const response = { status: 200, body };
       fetchMock.mock(graphqlEndpoint, response);
 
       return fetcher({ query: '' }).catch(() => {
-        expect(errorReporter).toHaveBeenCalledWith(body);
+        expect(errorReporter).toHaveBeenCalledWith(body.errors);
       });
     });
   });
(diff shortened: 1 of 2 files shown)


src/utils/graphqlFetch.ts
```diff
@@ -88,9 +79,21 @@ export function createGraphqlFetch({
       };
     }
 
-    // report errors if any, but continue
     if (body.errors) {
       report(body.errors);
+
+      const authExpired = body.errors.some(e => {
+        return e.extensions && e.extensions.type === 'AuthFailed';
+      });
```
sslotsky: cc @seriousben 

I believe this is the correct check as we discussed.

src/utils/graphqlFetch.spec.ts
```diff
@@ -244,11 +258,15 @@ describe('graphqlFetch', () => {
...
-    it('401: unauthorized (token expired)', async () => {
+    it('Unauthorized (token expired)', async () => {
       const body = {
         data: null,
         errors: [
-          { message: 'User does not have permission to access the resource', path: ['resources'] },
+          {
+            message: 'User does not have permission to access the resource',
+            path: ['resources'],
+            extensions: { type: 'AuthFailed' },
```
dangodev: Are we sure these are the error types? If so, should we abstract these somewhere central in the codebase?

If they’re shipped in the GraphQL schema, can we type this object to make sure this is the correct type?

sslotsky: 1. Yes, they are the correct error types and values. You can check api.manifold.co for the shape, and Ben left a comment on the issue with the value to look for.
1. The shape of the errors array does not appear anywhere in the schema.
1. A string constant for `AuthFailed` is reasonable. Maybe not obviously necessary right now. We don't use constants for the HTTP status codes that we check against. This is not really any different.

PR #533, opened 2019-09-24
converted service card to graphql

davidleger95:
## Testing

- go to service card in storybook
  - make sure cards load properly.
  - make sure free tag is being properly applied to products with free plans.
  - make sure error messages are displayed when an error occurs.

## Checklist

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -600,16 +600,15 @@ export namespace Components {
     'required'?: boolean;
   }
   interface ManifoldServiceCard {
-    'isFeatured'?: boolean;
-    'preserveEvent'?: boolean;
-    'product'?: Catalog.Product;
-    'productId'?: string;
-    'productLabel'?: string;
-    'productLinkFormat'?: string;
     /**
     * _(hidden)_
     */
-    'restFetch'?: RestFetch;
+    'graphqlFetch'?: GraphqlFetch;
+    'isFeatured'?: boolean;
+    'preserveEvent'?: boolean;
+    'product'?: Product;
+    'productLabel'?: string;
+    'productLinkFormat'?: string;
   }
   interface ManifoldServiceCardView {
     'description'?: string;
@@ -1697,16 +1696,15 @@ declare namespace LocalJSX {
     'required'?: boolean;
   }
   interface ManifoldServiceCard extends JSXBase.HTMLAttributes<HTMLManifoldServiceCardElement> {
-    'isFeatured'?: boolean;
-    'preserveEvent'?: boolean;
-    'product'?: Catalog.Product;
-    'productId'?: string;
-    'productLabel'?: string;
-    'productLinkFormat'?: string;
     /**
     * _(hidden)_
     */
-    'restFetch'?: RestFetch;
+    'graphqlFetch'?: GraphqlFetch;
+    'isFeatured'?: boolean;
+    'preserveEvent'?: boolean;
+    'product'?: Product;
+    'productLabel'?: string;
+    'productLinkFormat'?: string;
   }
   interface ManifoldServiceCardView extends JSXBase.HTMLAttributes<HTMLManifoldServiceCardViewElement> {
     'description'?: string;
--- a/src/components/manifold-service-card/manifold-service-card.tsx
+++ b/src/components/manifold-service-card/manifold-service-card.tsx
@@ -2,7 +2,7 @@ import { h, Component, Element, State, Prop, Watch } from '@stencil/core';
 import { gql } from '@manifoldco/gql-zero';
 
 import connection from '../../state/connection';
-import { GraphqlFetch } from '../../utils/graphqlFetch';
+import { GraphqlFetch, GraphqlError } from '../../utils/graphqlFetch';
 import { Product } from '../../types/graphql';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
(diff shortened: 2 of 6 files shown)


src/components.d.ts
```diff
@@ -1697,16 +1696,15 @@ declare namespace LocalJSX {
...
   }
   interface ManifoldServiceCard extends JSXBase.HTMLAttributes<HTMLManifoldServiceCardElement> {
+    /**
+    * _(hidden)_
+    */
+    'graphqlFetch'?: GraphqlFetch;
     'isFeatured'?: boolean;
     'preserveEvent'?: boolean;
-    'product'?: Catalog.Product;
-    'productId'?: string;
```
davidleger95: Can't remember if this was a decision we made, but I know we wanted to simplify the component APIs a bit. Did we actually decide if we should remove `id` fields?

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,131 +1,98 @@
...
 
+const query = gql`
+  query PRODUCT_CARD($productLabel: String!) {
+    product(label: $productLabel) {
+      id
+      displayName
+      tagline
+      label
+      logoUrl
+      # must get all plans until 'free' filter works
```
davidleger95: @dangodev will this work with your gql minifier?

src/components.d.ts
```diff
@@ -600,16 +600,15 @@ export namespace Components {
     'required'?: boolean;
   }
   interface ManifoldServiceCard {
+    /**
+    * _(hidden)_
+    */
+    'graphqlFetch'?: GraphqlFetch;
```
domenicrosati: what does the hidden thing do?

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,131 +1,98 @@
...
-import { Catalog } from '../../types/catalog';
 import connection from '../../state/connection';
-import { RestFetch } from '../../utils/restFetch';
+import { GraphqlFetch } from '../../utils/graphqlFetch';
+import { Product } from '../../types/graphql';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
 
+const query = gql`
+  query PRODUCT_CARD($productLabel: String!) {
```
domenicrosati: why do we make reusable queries here?

domenicrosati: is it that they are more reusable? if so would the benefit from their own files/namespace 🤷‍♂

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,131 +1,98 @@
...
+          productLabel={this.product.label}
+          productLinkFormat={this.productLinkFormat}
+        >
+          <manifold-forward-slot slot="cta">
+            <slot name="cta" />
+          </manifold-forward-slot>
+        </manifold-service-card-view>
+      );
+    }
+    // ☠
```
domenicrosati: hahah what re: `// ☠`

domenicrosati: oh skeleton

src/components/manifold-service-card/manifold-service-card.tsx
domenicrosati: ++
is there a front end equivlenat to a `todo`?
in the backend we do `XXX(dom)` or `TODO(dom)` so that folks know who to ask about tech debt... not sure if its useful

src/components.d.ts
davidleger95: it adds this comment to the generated docs

src/components/manifold-service-card/manifold-service-card.tsx
davidleger95: I just have it here so that it's defined once when the file loads and then is reused for each instance/call. It's not much more efficient, but just a little bit, so probably doesn't actually matter much to put this here but my instinct is to define strings that don't change at the top-level of files. 🤷‍♂

src/components/manifold-service-card/manifold-service-card.tsx
davidleger95: ☠️

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,131 +1,98 @@
...
-    if (Array.isArray(response) && response.find(plan => plan.body.free === true)) {
-      this.free = true;
-    } else {
-      this.free = false;
+    const { data } = await this.graphqlFetch({ query, variables: { productLabel } });
+    if (data && data.product && data.product.plans) {
+      this.product = data.product;
+      // Product has at least one free plan.
+      this.free = !!data.product.plans.edges.find(({ node }) => node.free);
     }
```
davidleger95: going to add some error handling here too

src/components/manifold-service-card/manifold-service-card.tsx
dangodev: We’re not using that (yet), so the answer is: it _will_ 🙂

src/components.d.ts
dangodev: If it can be removed, it should be. IIRC IDs are mainly used with the REST resource endpoints where label wasn’t available. I don’t think there are any scenarios where we allow ID for the user’s sake (also, those may change over time so labels are safer).

src/components/manifold-service-card/manifold-service-card.tsx
davidleger95: hmm, might be good to do that. we don't usually but I can see it being useful

src/components/manifold-service-card/manifold-service-card.tsx
davidleger95: Errors added. Do we actually want errors here? I only see a few components display them like this

src/components/manifold-service-card/manifold-service-card.tsx
domenicrosati: ah ok makes sense!

PR #537, opened 2019-09-25
convert resource card to graphql

davidleger95:
## Reason for change

- Convert resource card to graphql

## Testing

Check that all states for resource card and resource list work the same in storybook

## Checklist


src/types/graphql.ts
```diff
@@ -856,26 +856,26 @@ export type Query = {
   /** 
  * Look up a profile by `id`, by `subject`, or based on the currently authenticated profile.
    * 
-   * Note: This query can only be performed when authenticated.
+   * Note: Subject owner is currently only supported with use of a Platform API token.
```
domenicrosati: excellent

src/components/manifold-resource-card/manifold-resource-card.tsx
```diff
@@ -1,79 +1,89 @@
...
+      const { label, displayName, plan, id, status } = this.resource;
+      const { logoUrl } = (plan && plan.product) || {};
+
+      return (
+        <manifold-resource-card-view
+          label={label}
+          name={displayName}
+          logo={logoUrl}
+          resourceId={id}
+          resourceStatus={status.label}
```
domenicrosati: awesome that this seems to work out so well

src/components/Error/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+  errors?: GraphqlError[];
+}
+
+export const Error: FunctionalComponent<ErrorProps> = ({ errors }) => {
+  if (errors) {
+    errors.map(({ message }) => <manifold-toast alertType="error">{message}</manifold-toast>);
+  }
+
+  return [];
+};
```
davidleger95: Rather than repeating this everywhere I just made a simple error component

src/components/manifold-resource-card/manifold-resource-card.tsx
davidleger95: Thinking more about it, I think I'll need to make some changes here. The status doesn't use the same values as the old one

src/components/Error/index.tsx
domenicrosati: nice. This mutates errors eh?

davidleger95:
> Not sure what the frontend convention is for commits but you have 23 commits here.

yeah, this one kinda lagged behind for a while (because of other priorities) and then there were conflicts and then there were changes (not using `displayName` anymore) and then there were more conflict and then more conflict 🤷‍♂ We do typically squash on merge though, if that changes your opinion

PR #549, opened 2019-09-27
Reduce sidebar width on product modal at tablet sizes

markjamieson:
## Reason for change

Fixes https://app.zenhub.com/workspaces/engineering-56990653b8516b1b4867e8b2/issues/manifoldco/engineering/9469

## Testing

Open a product modal on render, adjust width through the various breakpoints

## Checklist

### DIFF
--- a/src/components/manifold-product-page/manifold-product-page.css
+++ b/src/components/manifold-product-page/manifold-product-page.css
@@ -92,7 +92,7 @@
 .grid {
   display: grid;
   grid-column-gap: 3rem;
-  grid-template-columns: repeat(auto-fit, minmax(12rem, max-content));
+  grid-template-columns: 1fr;
 
   @media (min-width: 600px) {
     grid-template-columns: max-content auto;
@@ -104,6 +104,10 @@
   margin-right: auto;
   margin-left: auto;
   padding-top: var(--overlap);
+
+  @media (min-width: 1106px) {
+    max-width: 20rem;
+  }
 }
 
 .sidebar-cta {
--- a/src/components/manifold-product-page/manifold-product-page.css
+++ b/src/components/manifold-product-page/manifold-product-page.css
@@ -100,10 +100,14 @@
 }
 
 .sidebar {
-  max-width: 20rem;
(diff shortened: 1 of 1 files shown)


dangodev:
I’m scared this won’t fix the problem well because it relies on window width. As in the case with our platforms, sometimes this is in a wrapper that isn’t full-window-width. For that reason, perhaps an approach using CSS Grid would be better—that actually responds based on the size of the columns rather than the browser width. Example: https://labs.jensimmons.com/2017/03-009.html

markjamieson:
This seems to fix the problem for me. @davidleger95 mentioned JS media queries in the original issue, but I'm not sure why those are needed. I might be oversimplifying!

The original problem comes from the screenshot menu having an absolute width (4rem * number of items), and just not having enough room to display in the grid at tablet width. Reducing the sidebar seemed to work, but we could also look at having an overflow scroll on just that menu.

dangodev:
@markjamieson what about this, instead of all the media queries?

```css
.grid {
  grid-template-columns: repeat(auto-fit, minmax(12rem, max-content))
}
```

Something like this just auto-wraps when either column goes smaller than an arbitrary width (here, `12rem`). I’m fine with this fix, but something like this which is a function of content size rather than browser size may last us a little longer?

dangodev:
Ah sorry; wasn’t trying to strongarm this; just was scared that once this component was touched again in the future we’d reintroduce the problem for platforms.

If we go with this, then be sure to remove all the `grid-template-columns` in the other media queries as well.

markjamieson:
I don't understand your solution well enough to figure out how to fix the issues I'm having with it, and I have to put it aside for a bit. I've done barely anything though, so if you want to take over please feel free. I was just hoping to knock this one out quick.

dangodev:
Let’s go with your tested fix then if it works for now. Sorry for blocking this. I can explain more reasoning later but my solution was half-baked here.

markjamieson:
My fix only works if the max number of screenshots is 3. I can't remember if that's the case. And obviously that's not very robust CSS.

PR #556, opened 2019-09-29
Add manifold-resource-loaded event to manifold-resource-container

markjamieson:
## Reason for change

I _think_ I need this event to improve loading experience. Putting up this PR to better demonstrate my approach. Do not merge.

## Checklist

### DIFF
--- a/src/components/manifold-resource-container/manifold-resource-container.tsx
+++ b/src/components/manifold-resource-container/manifold-resource-container.tsx
@@ -1,4 +1,4 @@
-import { h, Component, Prop, State, Watch } from '@stencil/core';
+import { h, Component, Prop, State, Watch, Event, EventEmitter } from '@stencil/core';
 import { gql } from '@manifoldco/gql-zero';
 
 import { Gateway } from '../../types/gateway';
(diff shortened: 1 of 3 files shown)


src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -39,6 +39,7 @@ export class ManifoldResourceContainer {
   @State() gqlData?: Resource;
   @State() loading: boolean = true;
   @State() timeout?: number;
+  @Event({ eventName: 'manifold-resource-loaded' }) resourceLoaded: EventEmitter;
```
dangodev: Should we change this to `-load` to be more consistent with the other `-load` events? (I think we decided to use present-tense verbs)

markjamieson:
@dangodev Is there anything special I need to do in the docs since I've added an event?

dangodev:
> Is there anything special I need to do in the docs since I've added an event?

Ideally yes we’d document this event and add tests like the other events have to ensure they fire and we don’t break this functionality. But as this is somewhat of a patch, I’m fine with having this be an “internal” event that’s not documented.

PR #557, opened 2019-09-29
increase toast-line-height

domenicrosati:
## Reason for change
Fixes: https://github.com/manifoldco/engineering/issues/9495

Meg noticed the toast text is a little squished. Lets bump it.

## Testing

Have a look at storybook at this componenet and docs component. Does it look alright to you?

## Checklist

### DIFF
--- a/src/components/manifold-toast/manifold-toast.css
+++ b/src/components/manifold-toast/manifold-toast.css
@@ -27,7 +27,7 @@
   margin-top: 0;
   margin-bottom: 0;
   color: var(--manifold-color-primary);
-  line-height: 1;
+  line-height: 1.3;
   will-change: height;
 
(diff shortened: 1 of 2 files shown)


src/components/manifold-toast/manifold-toast.css
```diff
@@ -27,7 +27,7 @@
   margin-top: 0;
   margin-bottom: 0;
   color: var(--manifold-color-primary);
-  line-height: 1;
+  line-height: 1.3;
```
dangodev: Nice. Always keep `line-height` unitless like this, because `px` is bad whenever you change the font-size! Unitless automatically scales no matter the font-size so it’s always more versatile.

domenicrosati:
that would be preferrable i am having a hard time selecting it - maybe if I target the span directly

PR #559, opened 2019-09-30
Use label instead of name for resource display

dangodev:
## Reason for change
fixes manifoldco/engineering#9503

**Before** (resource was renamed to conflict)

**After** (resource was named to conflict, but resolved to `-2`)
Screen Shot 2019-09-30 at 07 43 22

## Checklist


dangodev:
Just a note: I had never seen the “Resource” dropdown missing from a Happo screen before, but it was just now. But I verified in Storybook it’s still there.

Either way, when we resume GraphQL conversion I have plans to remove that standalone resource dropdown component as that had to fetch its own resource data and we no longer have to do for GraphQL so it can be just a “dumb“ dropdown again.

PR #560, opened 2019-09-30
Let manifold-product emit an event when the request is complete

sslotsky:
## Reason for change

This is needed to support a full page view for product details in Render. Mainly, we need to be able to detect if a product was not found as a result of navigating to a URL with a bad product name, e.g. `addon/details/foo`. Additionally, when the product is found, emitting the product data will allow Render to add the product display name in their button, e.g. "Choose LogDNA".

## Testing

All tests should pass.

## Checklist

### DIFF
--- a/src/components/manifold-product/manifold-product.tsx
+++ b/src/components/manifold-product/manifold-product.tsx
@@ -1,4 +1,4 @@
-import { h, Component, Prop, State, Element, Watch } from '@stencil/core';
+import { h, Component, Prop, State, Element, Watch, Event, EventEmitter } from '@stencil/core';
 import { gql } from '@manifoldco/gql-zero';
 
 import { Product } from '../../types/graphql';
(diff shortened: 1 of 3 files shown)


src/components/manifold-product/manifold-product.tsx
```diff
@@ -63,6 +65,7 @@ export class ManifoldProduct {
     const { data } = await this.graphqlFetch({ query, variables });
 
     this.product = (data && data.product) || undefined;
+    this.loaded.emit(data && data.product);
```
dangodev: Do we want this event to fire if there’s no data? (up to you; just asking)

sslotsky: Yeah that's a fair question. I think we want to do more here to handle errors other than `NotFound`, but I'm not sure if now is the time to consider that.

PR #565, opened 2019-10-01
Align the alert icon to the top of the toast text

sslotsky:
## Reason for change

The vertical alignment of the alert icon is currently centered, and when the toast text spans multiple lines, this becomes a bit awkward, as seen in the screenshot below. Per @megthesmith, we should align this icon towards the top. 

image

## Testing

You might try modifying or creating a toast within Storybook where the text spans multiple lines. Or you can have a look at the screenshot below, which is using a build that includes these changes.

image

## Checklist

### DIFF
--- a/src/components/manifold-toast/manifold-toast.tsx
+++ b/src/components/manifold-toast/manifold-toast.tsx
@@ -101,7 +101,7 @@ export class ManifoldToast {
                 <manifold-icon icon={x} />
               </button>
             )}
-            <manifold-icon class="alert" icon={this.icon || this.alertTypeIcon} />
+            <manifold-icon icon={this.icon || this.alertTypeIcon} />
             <slot />
           </div>
         </p>
--- a/src/components/manifold-toast/manifold-toast.tsx
+++ b/src/components/manifold-toast/manifold-toast.tsx
@@ -102,7 +102,9 @@ export class ManifoldToast {
               </button>
             )}
             <manifold-icon icon={this.icon || this.alertTypeIcon} />
-            <slot />
+            <div class="content">
+              <slot />
+            </div>
           </div>
         </p>
       </Host>
--- a/src/components/manifold-toast/manifold-toast.tsx
+++ b/src/components/manifold-toast/manifold-toast.tsx
@@ -101,7 +101,7 @@ export class ManifoldToast {
                 <manifold-icon icon={x} />
               </button>
             )}
-            <manifold-icon icon={this.icon || this.alertTypeIcon} />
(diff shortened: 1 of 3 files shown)


sslotsky:
@dangodev or @davidleger95 I have no doubt that I didn't go about this in the best way, please chime in if I can do the following things better than I'm currently doing them:

1. Using the right selector
1. Pushing the top of the icon down to be even with the top of the text

sslotsky:
Hmm, this also looks pretty bad in the Happo diff when we only have one line of text. Advice is welcome here.

sslotsky:
FWIW my goal with these changes was to emulate the [screenshot](https://camo.githubusercontent.com/f41f9bb92980a6558808d059dabe40f34f7e37e3/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3561636236626465346235383036626332626365343638302f30666434326230652d346462352d343132362d386165332d323135653738383036303731) that @megthesmith posted in https://github.com/manifoldco/engineering/issues/9468

sslotsky:
If I remove that padding, it looks like this, which feels a bit off to me.

image

src/components/manifold-toast/manifold-toast.tsx
```diff
@@ -101,7 +101,7 @@ export class ManifoldToast {
                 <manifold-icon icon={x} />
               </button>
             )}
-            <manifold-icon icon={this.icon || this.alertTypeIcon} />
+            <manifold-icon class="alert" icon={this.icon || this.alertTypeIcon} />
```
dangodev: While you’re in this file, can you add a `<div>` around the slot? I’ve noticed that if a slot is > 1 element (e.g. text + `<a>` is a common one) the flexbox applies to it too and it looks weird

dangodev:
Hm. So the icon is perfectly-aligned at 16px because that’s how it’s laid-out. In the Render Dashboard is it possible to make the entire `<manifold-toast>` a smaller font size? IMO that should keep the icon aligned.

`0.25rem` is a very specific fix for one font size. I’m not against it in times like these, but I’m worried it’ll throw off alignment for the other ones as well.

sslotsky:
> `0.25rem` is a very specific fix for one font size.

I was concerned about that too but I wasn't sure how to get it right via `calc` and `--padding`. I'll play with font sizes to see if that helps me avoid the use of padding, and maybe that will help with the Happo diffs too. Thanks!

dangodev:
I think you may also need to remove `line-height: 1;` from `.toast`

PR #566, opened 2019-10-01
Convert the `manifold-resource-product` component to GraphQL

Minivera:

## Reason for change

The container still partially uses REST, I am converting it in smaller steps to not create a giant untestable PR.

## Testing

Test using storybook, the component should work like it used to.

## Checklist

### DIFF
--- a/src/components/manifold-resource-product/manifold-resource-product-happo.ts
+++ b/src/components/manifold-resource-product/manifold-resource-product-happo.ts
@@ -1,3 +1,5 @@
+import { Resource } from '../../types/graphql';
+
 export const skeleton = () => {
   const resourceProduct = document.createElement('manifold-resource-product');
   document.body.appendChild(resourceProduct);
@@ -21,9 +23,7 @@ export const loadedResource = () => {
 
   const resourceProduct = document.createElement('manifold-resource-product');
   resourceProduct.loading = false;
-  // TODO: Make good mocks or rethink the types, this ignore stops the compiler from wanting a complete resource
-  // @ts-ignore
-  resourceProduct.gqlData = GraphqlResource;
+  resourceProduct.gqlData = GraphqlResource as Resource;
 
   document.body.appendChild(resourceProduct);
   return resourceProduct.componentOnReady();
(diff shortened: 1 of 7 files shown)


src/components/manifold-resource-product/manifold-resource-product-happo.ts
```diff
@@ -3,3 +3,28 @@ export const skeleton = () => {
...
+        label: 'product',
+        logoUrl: 'https://fillmurray.com/200/200',
+      },
+    },
+  };
+
+  const resourceProduct = document.createElement('manifold-resource-product');
+  resourceProduct.loading = false;
+  // TODO: Make good mocks or rethink the types, this ignore stops the compiler from wanting a complete resource
+  // @ts-ignore
```
dangodev: Why not `GraphqlResource as thing` instead?

src/components/manifold-resource-product/manifold-resource-product.tsx
```diff
@@ -1,39 +1,50 @@
...
 import ResourceTunnel from '../../data/resource';
-import { Gateway } from '../../types/gateway';
+import { Product, Resource } from '../../types/graphql';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
 
 @Component({ tag: 'manifold-resource-product' })
 export class ManifoldResourceProduct {
-  @Prop() data?: Gateway.Resource;
+  @Prop() gqlData?: Resource;
```
dangodev: Why rename this? (just curious)

src/components/manifold-resource-product/manifold-resource-product-happo.ts
Minivera: Very, very good question 😅

src/components/manifold-resource-product/manifold-resource-product.tsx
Minivera: The resource container injects two pieces of data into its state Tunnel, `data` which is the REST resource and `gqlData`, which is its GraphQL equivalent. I'm using `gqlData` only as a migration path and will rename everything when all the consumers are converted 😄

PR #569, opened 2019-10-02
Function to convert graphql plan data to rest plan data

davidleger95:
## Reason for change

We need a function to convert between these data structures in order to ease the transition from REST to GraphQL. Here is the first pass at the function. There are still a few tasks left to do though.

TODO

- [x] Handle metered features
- [x] Handle Configurable features
- [x] Add tests

## Checklist

### DIFF
--- a/src/utils/plan.spec.ts
+++ b/src/utils/plan.spec.ts
@@ -190,34 +190,26 @@ describe('plan sort function', () => {
 });
 
 const graphqlPlan = {
-  id: '235exy25wvzpxj52p87bh87gbnj4y',
-  displayName: 'Custom',
-  label: 'custom',
+  id: '2352cvkk4f3rp17pbgbrmffe1jcw8',
+  displayName: 'Standard',
+  label: 'standard',
   product: {
-    id: '234w1jyaum5j0aqe3g3bmbqjgf20p',
+    id: '234htwpkzvg1vuyez6uybfhv8rjb2',
     provider: {
-      id: '2346mdxcuca9ez2n93f72nb2fpjgu',
+      id: '2343d7p36xwrydjy7120jxqxc7t22',
     },
   },
-  cost: 3300,
+  cost: 700,
   free: false,
   state: 'AVAILABLE',
   regions: {
     edges: [
       {
         node: {
-          id: '235m2c51y0625vvtk6ptf55bhpkty',
-          displayName: 'AWS - EU West 1 (Ireland)',
-          platform: 'aws',
-          dataCenter: 'eu-west-1',
-        },
-      },
-      {
-        node: {
-          id: '235mhkk15ky7ha9qpu4gazrqjt2gr',
-          displayName: 'AWS - US East 1 (N. Virginia)',
-          platform: 'aws',
-          dataCenter: 'us-east-1',
+          id: '235n4f9pxf8eyraj3y159x89z6jer',
+          displayName: 'All Regions',
+          platform: 'all',
+          dataCenter: 'global',
         },
       },
     ],
@@ -226,42 +218,62 @@ const graphqlPlan = {
     edges: [
       {
         node: {
-          id: 'cduq6x3fdnfq6x31ehmp6bbkd5q6e',
+          id: 'd1gpuvb5e9m6arb4bxtq8rbmd5hju',
           label: 'static-single-tenant',
           displayName: 'Single Tenant',
           displayValue: 'true',
         },
       },
+      {
+        node: {
+          id: 'edu62vk4c5t68qv2c5q68xv9chu6g',
+          label: 'bandwidth',
+          displayName: 'Bandwidth',
+          displayValue: 'Unlimited',
+        },
+      },
+      {
+        node: {
+          id: 'edu62vk4c5t68qv6e9jpabb8dxuq4',
+          label: 'free-hours',
+          displayName: 'Free Hours',
+          displayValue: '2',
+        },
+      },
+      {
+        node: {
+          id: 'edu62vk4c5t68qv6e9jpabb9dngpe',
+          label: 'free-imagga-images',
+          displayName: 'Free Imagga Images',
+          displayValue: '100',
+        },
+      },
     ],
   },
   meteredFeatures: {
-    edges: [],
-  },
-  configurableFeatures: {
     edges: [
       {
         node: {
-          id: 'cduq6x3fdnfp4rb3dduq0wr000000',
-          label: 'backups',
-          displayName: 'Backups',
-          type: 'NUMBER',
-          options: [
-            {
-              id: '00000000000000000000000000000',
-              label: '',
-              displayName: 'Backups',
-              displayValue: 'backups',
-            },
-          ],
+          id: 'edu62vk4c5t68qv9dngpetv15nmpu',
+          label: 'imagga-images',
+          displayName: 'Imagga Images',
           numericDetails: {
-            increment: 1,
-            min: 0,
-            max: 35,
-            unit: 'Days',
+            unit: 'image',
+            costTiers: [],
+          },
+        },
+      },
+      {
+        node: {
+          id: 'edu62vk4c5t68qvge9qp6tbkedmpw',
+          label: 'processing-duration',
+          displayName: 'Processing Duration',
+          numericDetails: {
+            unit: 'hour',
             costTiers: [
               {
-                limit: 35,
-                cost: 0,
+                limit: -1,
+                cost: 790000000,
               },
             ],
           },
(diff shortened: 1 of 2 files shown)


src/utils/plan.spec.ts
```diff
@@ -186,3 +188,369 @@ describe('plan sort function', () => {
...
+              },
+            ],
+          },
+        },
+      },
+    ],
+  },
+};
+
+const restPlan = [
```
dangodev: Let’s set this type with the Catalog Spec

src/utils/plan.spec.ts
```diff
@@ -186,3 +188,369 @@ describe('plan sort function', () => {
     expect(planSort(rightOrder, { free: true })).toEqual([freePlan, fauxFreePlan]);
   });
 });
+
+const graphqlPlan = {
```
dangodev: Same here: let’s set the type with our GraphQL type

src/utils/plan.spec.ts
dangodev: Actually on second thought, this is likely handled by the typing on the method itself, so it’s probably fine to leave untyped

src/utils/plan.spec.ts
```diff
@@ -186,3 +188,286 @@ describe('plan sort function', () => {
...
+      { feature: 'processing-duration', value: 'processing-hours' },
+      { feature: 'video-encoding', value: 'video-encoding' },
+    ],
+    label: 'standard',
+    name: 'Standard',
+    product_id: '234htwpkzvg1vuyez6uybfhv8rjb2',
+    provider_id: '2343d7p36xwrydjy7120jxqxc7t22',
+    regions: ['235n4f9pxf8eyraj3y159x89z6jer'],
+    // Not in GraphQL API
+    // resizable_to: null,
```
davidleger95: just commenting stuff out here to make it clear that this data is still in the REST structure, but not able to be derived from the GraphQL data. I think that most of it isn't used. If anything that's commented out is used, we'll have to add it to the API

src/utils/plan.spec.ts
davidleger95: I'll add this. 
Just for context though, the reason why I don't have it typed currently is that using the GraphQL types requires pagination data and a few other things that are not relevant to the test (this also means the types are unreliable because we could end up trying to access something that's undefined).

src/utils/plan.ts
```diff
@@ -325,3 +333,119 @@ export function planCost(restFetch: RestFetch, { planID, features, init }: PlanC
...
+    );
+  }
+  if (plan.meteredFeatures) {
+    features.push(
+      ...plan.meteredFeatures.edges.map(({ node }) => ({
+        feature: node.label,
+        value: node.label,
+      }))
+    );
+  }
```
davidleger95: It appears that the data to get a feature map isn't full available in graphql. If we do actually need the `plan.body.features` we'll have to make changes to the API

src/utils/plan.ts
```diff
@@ -325,3 +333,119 @@ export function planCost(restFetch: RestFetch, { planID, features, init }: PlanC
...
+
+  const newFeatures: Catalog.ExpandedFeature[] = features.edges.map(({ node: feature }) => {
+    return {
+      label: feature.label,
+      name: feature.displayName,
+      measurable: true,
+      type: 'number',
+      value: {
+        label: feature.label,
+        name: feature.displayName,
```
davidleger95: `name` and `label` for number features are unavailable in graphql

src/utils/plan.ts
```diff
@@ -325,3 +333,119 @@ export function planCost(restFetch: RestFetch, { planID, features, init }: PlanC
...
+      features: convertPlanFeatures(plan),
+      label: plan.label,
+      name: plan.displayName,
+      product_id: (plan.product && plan.product.id) || '',
+      provider_id: (plan.product && plan.product.provider && plan.product.provider.id) || '',
+      regions: convertRegionData(plan.regions),
+      state: plan.state && plan.state.toLocaleLowerCase(),
+      expanded_features: [
+        ...convertFixedFeatureData(plan.fixedFeatures),
+        ...convertMeteredFeatureData(plan.meteredFeatures),
```
davidleger95: we don't have configurable features for any products on platforms yet. Also, there are fields missing from graphql which means there will be a fair bit of missing data when we convert it. I won't add configurable features to this function unless we really need it because it's quite complex and not worth the effort if it won't be used.

src/utils/plan.ts
davidleger95: [Here's the diff from the test](https://circleci.com/gh/manifoldco/ui/4937) if you want more context.

src/utils/plan.ts
davidleger95: [Here's the diff from the test](https://circleci.com/gh/manifoldco/ui/4937) if you want more context.

src/utils/plan.ts
dangodev: @davidleger95 which fields are missing from GraphQL? For catalog, if we have all the configurable features we should add them here (as long as they’re not too much effort). I was thinking configurable features were only missing from resources

src/utils/plan.ts
dangodev: I’m a bit confused by the diff from the test—it just seems like the plans are out-of-order?

dangodev: Also nice multiline comment! New GH features are fun 😎

src/utils/plan.ts
dangodev: Are they unavailable from GraphQL? Or do you mean `value` and `displayValue`—i.e. the `label` and `displayName` is not what we expect them to be?

src/utils/plan.ts
davidleger95: The per-feature pricing is missing (already a ticket for this). Don't think I came across anything else that was uniquely missing from configurable features.

src/utils/plan.ts
davidleger95: I think they're in order 🤔 If there's something I'm missing, feel free to reorder them. What I'm seeing though is the value for `value` in each `feature` is not always present in the GraphQL data.

dangodev: Oh I was seeing things like this:

```diff
    -           "label": "no-overage",
    -           "name": "No Overage",
    +           "label": "imagga-images",
    +           "name": "Imagga Images",
```

Wasn’t sure if that was the wrong `label`, or what

dangodev: Oh yeah I see what you mean. Those should be `value` / `displayValue`

src/utils/plan.ts
davidleger95: that is correct. `feature.label` and `feature.displayName` are not what we want here. there are different name/label values for the feature and feature value.

src/utils/plan.spec.ts
dangodev: IMO `resizable_to` is fine to omit for now

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,327 @@ describe('plan sort function', () => {
...
+    state: 'available',
+    // Not in GraphQL API
+    // trial_days: 0,
```
dangodev: `trial_days` is also fine to omit

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,327 @@ describe('plan sort function', () => {
...
+    // defaultCost: 700,
```
dangodev: `defaultCost` is OK to omit

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,327 @@ describe('plan sort function', () => {
...
+    expanded_features: [
+      {
+        label: 'static-single-tenant',
+        name: 'Single Tenant',
+        type: 'boolean',
+        value: { label: 'true', name: 'true' },
+        value_string: 'true',
+        // Not in GraphQL API for fixed/metered features
+        // values: [{ label: 'true', name: 'true' }, { label: 'false', name: 'false' }],
```
dangodev: I’m fine omitting this

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,327 @@ describe('plan sort function', () => {
...
+    state: 'available',
+    // Not in GraphQL API
+    // trial_days: 0,
+    // defaultCost: 700,
+    expanded_features: [
+      {
+        label: 'static-single-tenant',
+        name: 'Single Tenant',
+        type: 'boolean',
+        value: { label: 'true', name: 'true' },
```
dangodev: I think this is omitted too? Which I’m fine with for fixed features

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,327 @@ describe('plan sort function', () => {
...
+        value: { label: 'two', name: '2' },
+        value_string: '2',
+      },
+      {
+        label: 'free-imagga-images',
+        name: 'Free Imagga Images',
+        type: 'string',
+        // Not in GraphQL API for fixed/metered features
+        // values: [{ label: 'one-hundred', name: '100' }],
+        value: { label: 'one-hundred', name: '100' },
```
dangodev: These are the only parts that seem to be throwing the test suite. For these, let’s just add comments like you have been doing so well, and just fake these. The comment acknowledges that these don’t match REST, but it’s the best we can do for now.

I think Goal 1 is moving to GraphQL.
Goal 2 is filling in the GraphQL gaps (ideally this happens alongside #1)
Goal 3 is re-implementing configurable features as a concerted effort.

davidleger95: I'm in favor of pushing this forward but one thing I want to be sure of is that the right values are in the right place. Especially for submitting features for provisioning -- or do we only submit configurable features when provisioning?

dangodev:
Gary asked me to figure out what’s still undone for GraphQL in preparation for the squad change next week, and I used your PR to see what’s undone for GraphQL for configurable features. Anyway to get a feel for the work myself, I wanted to see how much work it’d be to convert configurable features.

I think I have a quick add—may be imperfect—but it should add configurable features to your PR by pasting in a couple functions and some test data: 

https://gist.github.com/dangodev/797d841dc34045eda2fa3f9218399ae5

Again, once **you fix up the current tests** (whatever that means), let’s merge this and keep going! 🚀

src/utils/plan.spec.ts
dangodev: Only configurable. Otherwise it’s not needed

davidleger95: awesome :)

davidleger95:
Not sure what's going on here with VRT 🤔 I reran the job a couple times and it's still hanging

src/utils/plan.spec.ts
```diff
@@ -186,3 +189,442 @@ describe('plan sort function', () => {
...
+    //   { feature: 'free-imagga-images', value: 'one-hundred' },
+    //   { feature: 'imagga-images', value: 'no-overage' },
+    //   { feature: 'processing-duration', value: 'processing-hours' },
+    //   { feature: 'video-encoding', value: 'video-encoding' },
+    // ],
+    features: [
+      { feature: 'instance_class', value: 'db.t2.micro' },
+      { feature: 'storage', value: '5 GB' },
+      { feature: 'redundancy', value: 'false' },
+    ],
```
davidleger95: Only including confugurable features in this array because we don't care about the values for fixed/metered features

src/utils/plan.ts
```diff
@@ -325,3 +333,177 @@ export function planCost(restFetch: RestFetch, { planID, features, init }: PlanC
...
+      name: feature.displayName,
+      type,
+      value,
+      values: feature.options
+        ? feature.options.map(option => ({
+            label: option.label,
+            name: option.displayValue,
+          }))
+        : [],
+      value_string: firstOption ? firstOption.displayValue : feature.displayName,
```
davidleger95: small tweak here to use the first option value if it's there

dangodev: Makes sense

src/utils/plan.spec.ts
dangodev: Yeah that was one of the confusing parts of the previous API—the `value` wasn’t always needed. I’ll save you from the full explanation but it basically was an artifact of grafting product data onto plan data that we are dealing with now in a much friendlier way.

src/utils/plan.spec.ts
dangodev: (we had plans for this but never ended up using it; we could drop this field altogether from the database)

src/utils/plan.spec.ts
dangodev: This was a temporary add because the meaning of `cost` changed when we added metered & configurable features. Ideally there should only be one `cost` for plans, and we can improve that going forward. We just needed `cost` and `defaultCost` because… reasons

PR #575, opened 2019-10-03
have connecition decorator use graphql when available to get identity

domenicrosati:
## Reason for change

We were waiting until we had graphql ready to get owner id for things. Now that its ready we can do that. We still fall back to identity incase folks are using manifold api token.
Also the string undefined was not being checked for; add that check.

## Testing

Use story book with both a old manifold api token and a manifold platform token.

## Checklist

### DIFF
--- a/stories/data/connectionDecorator.js
+++ b/stories/data/connectionDecorator.js
@@ -1,8 +1,7 @@
 import { text, radios } from '@storybook/addon-knobs';
 import { gql } from '@manifoldco/gql-zero';
 
-// Creates a fake token provider and manifold connection that adds the api token from localstorage
-// with an expiriy date a week from when it is created. 6.04e8 is a week in milliseconds.
+// Creates a fake token provider and manifold connection that adds the api token from localstorage.
 export const manifoldConnectionDecorator = storyFn => {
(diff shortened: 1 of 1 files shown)


stories/data/connectionDecorator.js
```diff
@@ -11,15 +10,40 @@ export const manifoldConnectionDecorator = storyFn => {
...
-    fetch('https://api.identity.manifold.co/v1/self', {
-      headers: { authorization: `Bearer ${token}`, 'Content-type': 'application/json' },
-    })
-      .then(data => data.json())
-      .then(({ id }) => {
-        localStorage.setItem('manifold_user_id', id);
+  if (
+    token &&
+    (!localStorage.getItem('manifold_user_id') ||
+      localStorage.getItem('manifold_user_id') === 'undefined')
```
sslotsky: I find it odd that the value would ever evaluate to the string `'undefined'`, how does that happen?

domenicrosati: this does it:
```
const { id } = {	noidhere: "poo" }
localStorage.setItem('silly_id', id);
```
so prolly when that old identity call failed and we did: localStorage.setItem('manifold_user_id', id);

sslotsky: Weird.

PR #576, opened 2019-10-03
Ability to retrieve auth token from UI project

sslotsky:
## Reason for change

In the Render dashboard, we are using a thing called `manifoldClient` in order to make GraphQL calls directly to our API, but it uses an unreliable mechanism for retrieving the auth token.

This PR exposes a `ensureAuthToken` function that can be imported from `@manifoldco/ui`. This function is built on our `waitForAuthToken` logic, which means that it uses the same logic as our API calls do for making sure the auth token is present.

### Bonus

This solution came from experimenting with ways to solve the LogDNA check for Render more reliably. While I was trying different solutions, I realized that our `fetchAllPages` function expects callers to supply a query that asks for the `pageInfo`, but it doesn't have any way to enforce it.

This PR now includes a check for `pageInfo`, and if it doesn't exist, it will throw an error with a helpful message explaining how to assemble a query that includes the `pageInfo`. 

## Testing

You might try this out in Render dashboard as follows:

```js
...

import { ensureAuthToken } from "@manifoldco/ui";
...

const authLink = setContext(async (_, { headers }) => {
  const token = await ensureAuthToken();

  return {
    headers: {
      ...headers,
      ...(token ? { authorization: `Bearer ${token}` } : {}),
      "content-type": "application/json",
      "access-control-allow-origin": "*",
      Connection: "keep-alive"
    }
  };
});
```

If you make this change, then making calls using the `manifoldClient` should always result in the auth token being retrieved correctly.

## Downside

None! Just kidding.

If we use the technique shown above in Render dashboard, it will mean that public API calls will require authentication. Currently, any page in Render dashboard that calls our API will authenticate the user, so this isn't a problem today. But if they ever wanted to show public stuff without authenticating the user first, the `manifoldClient` would barf :nauseated_face: 

We can get around this, though. If Render ever needs to make a call to the catalog, we can define two clients: one that relies on the auth token, and another that doesn't.

## Checklist

### DIFF
--- a/docs/docs/advanced/authentication.md
+++ b/docs/docs/advanced/authentication.md
@@ -196,6 +196,36 @@ minor performance boost (“minor” because typically our services detect an in
 within milliseconds, which does take time but in many instances your users may not notice a
 difference of milliseconds).
 
+## Using our GraphQL API
+
+If you are calling our GraphQL API directly from your client code, and the call that you're making
+requires authentication, you can ask us for the auth token by calling the `ensureAuthToken`
+function:
+
+```js
+import { ensureAuthToken } from '@manifoldco/ui';
+
+const authLink = setContext(async (_, { headers }) => {
+  const token = await ensureAuthToken();
+
+  return {
+    headers: {
+      ...headers,
+      ...(token ? { authorization: `Bearer ${token}` } : {}),
+      'content-type': 'application/json',
+      'access-control-allow-origin': '*',
+      Connection: 'keep-alive',
+    },
+  };
+});
+```
+
+This function will return the auth token as soon as authentication has completed. If the
+authentication process does not occur, the function will eventually timeout and throw an exception.
+The wait time defaults to 15 seconds, but is configurable through [the
+manifold-connection][connection] component. You should only use this function if the call you're
+making requires authentication, since public calls to our catalog don't require an auth token.
+
 [authentication]: https://docs.manifold.co/docs/platforms-auth-AzsO1HvPT1Hnojsrsb10L
 [connection]: /connection
 [oauth2]: https://www.oauth.com/oauth2-servers/access-tokens/authorization-code-request
--- a/docs/docs/advanced/authentication.md
+++ b/docs/docs/advanced/authentication.md
@@ -200,10 +200,19 @@ difference of milliseconds).
 
 If you are calling our GraphQL API directly from your client code, and the call that you're making
 requires authentication, you can ask us for the auth token by calling the `ensureAuthToken`
-function:
+function. Here's how you might use this with ApolloClient to add our token to your request headers:
 
 ```js
 import { ensureAuthToken } from '@manifoldco/ui';
+import ApolloClient from 'apollo-client';
+import { createHttpLink } from 'apollo-link-http';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { setContext } from 'apollo-link-context';
+
+const cache = new InMemoryCache();
+const link = new createHttpLink({
+  uri: 'https://api.manifold.co/graphql',
+});
 
 const authLink = setContext(async (_, { headers }) => {
   const token = await ensureAuthToken();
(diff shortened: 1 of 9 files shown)


dangodev:
This looks good to me! Just would like to add documentation on how to use this. But it looks like Stencil is exporting this function properly from the  `@manifoldco/ui` package.

src/utils/fetchAllPages.ts
```diff
@@ -39,6 +71,10 @@ export default async function fetchAllPages<Edge>({
     },
   };
 
+  if (!pageInfo) {
+    throw new MissingPageInfo(query);
+  }
```
dangodev: Will this throw for queries like:
```gql
product(label: "jawsdb-mysql") {
}
```
?

dangodev: Oh I guess in that instance you wouldn’t use `fetchAllPages`, I guess

sslotsky:
Had some time to mull this over and I don't think it's quite enough. It improves Render's ability to reliably get an auth token, but it won't help with things like retrying the call if that token turns out to be expired. I'm going to research Apollo think on this a bit, and see if we can do better.

sslotsky:
In fact, [this](https://www.apollographql.com/docs/react/data/error-handling/#usage) seems like a better path forward.

sslotsky:
Oh, [dangit](https://www.apollographql.com/docs/link/links/retry/)

> It does not (currently) handle retries for GraphQL errors in the response, only for network errors.

sslotsky:
Interesting! If we changed our minds about returning 401 errors from unauthenticated GraphQL queries, [this example](https://www.apollographql.com/docs/link/links/context/#caching-lookups) would fit the bill quite nicely.

src/utils/fetchAllPages.ts
sslotsky: Right. This is strictly to handle cases where `fetchAllPages` is called with a query that doesn't have `pageInfo`.

sslotsky:
Ok, I had a chance to think about this and I think this approach is fine if we rename things a bit.

1. The `manifoldClient` being used here could be called `manifoldAuthenticatedClient` or something like that, to indicate that it should only be used for authenticated requests. If a platform wants to make a public request, they would use a client that doesn't rely on `getAuthToken()`.
1. We should rename `getAuthToken()` to something like `ensureAuthToken()`, or something similar that indicates that it will fail if it's unable to retrieve an auth token.

Open to naming suggestions. Thoughts @dangodev ?

docs/docs/connection.md
```diff
@@ -31,3 +31,16 @@ The `<manifold-connection>` component is only needed once, and can be placed any
...
+
+## Authentication timeout
+
+Many API calls will require authentication. If no auth token is present, we will wait for the auth
+token to be received and then try the call again. This process will eventually timeout and throw an
+exception. The default wait time is 15 seconds, but this is can be configured by setting the
+`waitTime` property on this component.
+
+```html
+<manifold-connection waitTime="20">
```
dangodev: Should be `wait-time`

dangodev:
> Thoughts @dangodev ?

I think your naming suggestions are fine! Makes sense to me.

I see in the example in the docs you use `setContext` and `authLink`. Are those written for a particular GraphQL client (Apollo)? Maybe we should indicate as much how to pass those headers in.

docs/docs/connection.md
sslotsky: Oh right, thanks!

dangodev:
> Maybe less is more so we don't have to be specific to an implementation, or maybe more is more

I think more is more here 😄. Apollo is kind of the lingua franca it seems. I don’t think we have to add _much_ more, but having a drop-in for Apollo would be great here (and FWIW I think it’s fine to link to their docs for adding headers; I don’t think we have to write much more, but for me it’d be nice if this was paste-able as possible).

sslotsky:
@dangodev updated docs as suggested!

docs/docs/advanced/authentication.md
```diff
@@ -196,6 +196,52 @@ minor performance boost (“minor” because typically our services detect an in
...
+  };
+});
+
+const manifoldAuthenticatedClient = new ApolloClient({
+  link: authLink.concat(link),
+  cache,
+});
+
+export default manifoldAuthenticatedClient;
+```
```
dangodev: 💯

PR #578, opened 2019-10-04
Convert the manifold-resource-plan component to GraphQL

Minivera:

## Reason for change

This converts the manifold-resource-plan component to GraphQL in an effort to convert the manifold-resource-container component. It does not actually uses the graphQL data, but instead uses the plan data conversion function created by dave.js.

## Testing

Test with storybook, I suggest using the `manifold-mock-resource` component in the resource stories for easy testing.

## Checklist


src/components/manifold-resource-plan/manifold-resource-plan.tsx
```diff
@@ -1,42 +1,47 @@
...
   componentWillLoad() {}
 
   @logger()
   render() {
+    let product: Product | null | undefined;
+    let plan: Plan | null = null;
+    if (!this.loading && this.gqlData) {
+      if (this.gqlData.plan) {
+        plan = this.gqlData.plan;
+        product = this.gqlData.plan.product;
```
davidleger95: if `data` is no longer used here, we can rename `gqlData` to `data`

dangodev:
Screen Shot 2019-10-04 at 10 43 47

This doesn’t seem to be working

src/components/manifold-resource-plan/manifold-resource-plan.tsx
Minivera: It is still used, I'm slowly moving it to the new gql data type

dangodev:
> Where are you seeing this issue? the docs?

In Storybook, pulling from a resource in production with my auth token. On the *resource detail* story

Minivera:
Will try that, it's working fine with logdna for me

Minivera:
@dangodev It should be good now, I've added the query that fetches the plan's data

src/components/manifold-resource-plan/manifold-resource-plan-happo.ts
```diff
@@ -0,0 +1,18 @@
...
+  return resourcePlan.componentOnReady();
+};
+
+export const loadedResource = () => {
+  const resourcePlan = document.createElement('manifold-resource-plan');
+  resourcePlan.gqlData = resource as Resource;
+
+  document.body.appendChild(resourcePlan);
+  return resourcePlan.componentOnReady();
+};
```
dangodev: 💯

src/components/manifold-resource-plan/manifold-resource-plan.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  it('Renders a skeleton if loading', async () => {
+    element.loading = true;
+    const root = page.root as HTMLElement;
+    root.appendChild(element);
+
+    await page.waitForChanges();
+
+    expect(element).toEqualHtml(`
+      <manifold-resource-plan>
+        <manifold-plan-details data-test-loading="true"></manifold-plan-details>
```
dangodev: I’m not crazy about this test relying on a data attribute that doesn’t have meaning to the element. What if instead you added `components: [ManifoldResourcePlan, ManifoldPlanDetails]`, and you queried the HTML to see if `<manifold-skeleton-text>` was in the DOM? It’d be only another step, and would let us more easily refactor this component if we needed to without understanding why `data-test-loading` is or isn’t used.

dangodev: It’d shift the meaning of the test from whether or not an arbitrary attribute exists, to what’s being actually rendered to the user, which is more meaningful IMO

src/components/manifold-resource-plan/manifold-resource-plan.spec.ts
```diff
@@ -24,7 +25,40 @@ describe('<manifold-resource-product>', () => {
...
+               </manifold-skeleton-text>
+               <footer class="footer">
+                 <manifold-skeleton-text>
+                   Free
+                 </manifold-skeleton-text>
+               </footer>
+             </div>
+           </section>
+         </mock:shadow-root>
+       </manifold-plan-details>
```
dangodev: Nit: this is still a little too restrictive, because we want to be able to _slightly_ modify this HTML. What if, instead, you simply queried to make sure that `<manifold-skeleton-text>` was a child descendant? You should be able to select it with `page.root.querySelector('manifold-plan-details').shadowRoot.querySelector('manifold-skeleton-text')` (but broken up for TS, of course)

PR #581, opened 2019-10-07
Use documents to generate types specific to queries

sslotsky:
## Reason for change

The type definitions that we currently use for our GraphQL queries contain every possible field, which doesn't account for the fact that GraphQL queries can omit fields.

This PR introduces [the documents field](https://graphql-code-generator.com/docs/getting-started/documents-field) in our codegen configuration, which will generate types for queries and mutations found in `.graphql` files. Some additional changes were also needed in order to import strings from `.graphql` files in TypeScript. These include some tsconfig changes and a custom rollup plugin.

One thing you may notice is that I was unable to get the newly generated types to fit the `fetchAllPages` scenario in `manifold-marketplace.tsx`. This is unfortunate, but ultimately okay, because `fetchAllPages` is a temporary solution which will only be around until we implement actual pagination.

## Checklist


stencil.config.ts
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  const filter = createFilter(opts.include, opts.exclude);
+
+  return {
+    name: 'gql',
+    transform(code, id) {
+      if (filter(id)) {
+        return {
+          code: `export default ${JSON.stringify(code)}`,
```
sslotsky: I actually stole this custom plugin from [a similar example that imports html files as strings](https://github.com/bdadam/rollup-plugin-html/blob/master/index.js#L18). @dangodev you'll notice that they are minifying the string at build time. We could maybe do the same!

tsconfig.json
```diff
@@ -16,8 +16,10 @@
     "jsx": "react",
     "jsxFactory": "h",
     "strictNullChecks": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "typeRoots": ["node_modules/@types", "src/@types"]
   },
+  "files": ["src/@types/graphql.d.ts"],
```
dangodev: Do we need this and `typeRoots`? We’re using `observe-rect.d.ts` and it didn’t seem to need this.

stencil.config.ts
dangodev: Oh nice! Yeah this will enable us to minify at build-time which will be much faster/lighter-weight. I’ll see if that plugin minifies the way I’d like it to (removing GraphQL comments but keeping strings, e.g.)

dangodev: Ah no, this doesn’t minify for GraphQL properly when strings are added. And it leaves whitespaces that can be taken out. I think we can port our [minify function](https://github.com/manifoldco/gql-zero/blob/dangodev/minify/src/minify/index.ts) into this to run at build-time though

tsconfig.json
sslotsky: I don't know! I was just following a blog post. I can try removing it and see if it still works.

stencil.config.ts
sslotsky: Sorry, that's what I meant :smile: Their example uses an HTML minifier which I did not expect to work for GraphQL syntax :grin:

tsconfig.json
dangodev: Yeah I think because our `src/types` directory is included with `include: ["src"]`, TypeScript knows about this already. There are definitely ways to use `typeRoots` that… well I’m sure there’s a use for it I just haven’t run across it yet 😅

stencil.config.ts
```diff
@@ -1,17 +1,43 @@
...
+        };
+      }
+    },
+  };
+}
```
dangodev: I don’t mind this function at all, but is there a premade Rollup plugin that’ll do this? Just curious, and trying to avoid debugging this myself

sslotsky: The only one I found assumed that it was a GraphQL document rather than a string. You should never need to debug this, it just exports a string from .graphql files.

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -119,7 +121,7 @@ describe('<manifold-data-resize-button>', () => {
       );
     });
 
-    it('success', async () => {
+    it.only('success', async () => {
```
sslotsky: woops better remove that!

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -40,13 +40,15 @@ describe('<manifold-data-deprovision-button>', () => {
       const body = (req.body && req.body.toString()) || '';
       const { variables } = JSON.parse(body);
       const errors: GraphqlError[] = [{ message: 'something went wrong' }];
-      if (body.includes('mutation DELETE_RESOURCE')) {
+      if (body.includes('mutation DeleteResource')) {
```
dangodev: Oh queries _are_ normally PascalCased—you’re right 👍. Don’t know where I got that snake case from

sslotsky: I think previously it didn't really matter which one we chose. But now that we're generating types from them, using screaming snake case would generate types like `Delete_Resource_Mutation`, so I figured Pascal would be better for consistency of type definition casing.

jest-transform-graphql.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  process: src => {
+    return `module.exports=${JSON.stringify(src)};module.exports.default=module.exports;`;
+  },
+};
```
dangodev: 👍

src/components/manifold-data-resize-button/manifold-data-resize-button.tsx
```diff
@@ -5,6 +5,8 @@ import connection from '../../state/connection';
 import { GraphqlFetch } from '../../utils/graphqlFetch';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
+import updatePlan from './updatePlan.graphql';
+import { ResourceChangePlanMutation } from '../../types/graphql';
```
dangodev: Love this! Good stuff

PR #583, opened 2019-10-07
Format docs w/ Prettier

dangodev:
## Reason for change
This just runs `prettier` on all the Markdown files to make them consistent. No changes.

## Checklist


docs/.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     'class-methods-use-this': 'off', // component lifecycle methods sometimes don't use `this`
     'import/prefer-default-export': 'off', // named exports are perfectly fine
     'lines-between-class-members': 'off', // class members don’t need that space!
+    'prettier/prettier': 'error',
```
dangodev: Prettier will now throw an error in our JS, but won’t in our Markdown. IMO I don’t want to mess with Prettier markdown plugins unless someone feels strongly we should

PR #603, opened 2019-10-08
Convert the manifold resource sso component to GraphQL

Minivera:

## Reason for change

Converted the container wrapper for the resource rename component to GraphQL.

## Testing

Test with Storybook.

## Checklist

### DIFF
--- a/src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
+++ b/src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
@@ -1,0 +1,51 @@
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+
+import { Resource } from '../../types/graphql';
+import { resource } from '../../spec/mock/graphql';
+import { ManifoldResourceSso } from './manifold-resource-sso';
+import { ManifoldDataSsoButton } from '../manifold-data-sso-button/manifold-data-sso-button';
+
+describe('<manifold-resource-sso>', () => {
+  let page: SpecPage;
+  let element: HTMLManifoldResourceSsoElement;
+  beforeEach(async () => {
+    page = await newSpecPage({
+      components: [ManifoldResourceSso, ManifoldDataSsoButton],
+      html: `<div></div>`,
+    });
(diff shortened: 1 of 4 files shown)


src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  let element: HTMLManifoldResourceSsoElement;
+  beforeEach(async () => {
+    page = await newSpecPage({
+      components: [ManifoldResourceSso, ManifoldDataSsoButton],
+      html: `<div></div>`,
+    });
+    element = page.doc.createElement('manifold-resource-sso');
+  });
+
+  it('Renders a skeleton if loading', async () => {
```
dangodev: Why is this test testing for a skeleton when loading then querying for a button?

src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    const root = page.root as HTMLElement;
+    root.appendChild(element);
+
+    await page.waitForChanges();
+
+    const button = element.querySelector('manifold-data-sso-button');
+    expect(button).toBeDefined();
+
+    if (button) {
+      expect(button.loading).toBeFalsy();
```
dangodev: Testing the `resourceId` and `resourceLabel` here seems OK, but is there something better (semantic HTML) we can query for, rather than a `loading` prop?

src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
Minivera: I want to test that the button receives the loading prop depending on the state of the data. I'll rename.

src/components/manifold-resource-sso/manifold-resource-sso.spec.ts
Minivera: I'd like to not go too deep into the button's implementation. It receives the loading prop if this wrapper component is loading, what it does with it is its business.

PR #604, opened 2019-10-08
Convert deprovision button to GraphQL

dangodev:
## Reason for change

Converts `<manifold-data-deprovision-button>` to GraphQL.

## Testing

In order to test this, you need a **PLATFORM TOKEN** (grab one from `localStorage.manifold_api_token` from Render or TacoCloud). Then just use that for `manifold_api_token` in Storybook.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,19 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Converted `manifold-resource-plan` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#578)
+- Converted `manifold-resource-credentials` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#584)
+- Converted `manifold-resource-deprovision` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#595)
 - Converted `<manifold-data-provision-button>` to use GraphQL data (#600)
 - Converted `manifold-resource-rename` to use the GraphQL data fetched from the
   `manifold-resource-container`. (#602)
 - Converted `<manifold-data-rename-button>` to use GraphQL (#596)
 - Converted `manifold-resource-sso` to use the GraphQL data fetched from the
   `manifold-resource-container`. (#603)
+- Converted `<manifold-data-deprovision-button>` to use GraphQL (#604)
 
 ### Breaking Changes:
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -13,10 +13,17 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Converted `manifold-resource-plan` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#578)
+- Converted `manifold-resource-credentials` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#584)
+- Converted `manifold-resource-deprovision` to use the GraphQL data fetched from the
+  `manifold-resource-container`. (#595)
```
dangodev: These weren’t released in the previous version. CHANGELOG updated.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -108,7 +158,7 @@ export class ManifoldDataDeprovisionButton {
       <button
         type="submit"
         onClick={() => this.deprovision()}
-        disabled={!this.resourceId && !this.loading}
+        disabled={!this.resourceId || !this.ownerId}
```
dangodev: @Minivera I’m worried about the logic here. Before, it **won’t** disable this if `resourceId` is missing and it’s **not** loading. Isn’t `loading` just a check to make it wait for `resourceId`?

We do something similar in the rename button as well. I’d rather just make this available as soon as it has the data it needs.

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -54,52 +88,68 @@ export class ManifoldDataDeprovisionButton {
...
-        options: { method: 'DELETE' },
-      });
-      const success: SuccessMessage = {
-        message: `${this.resourceLabel} successfully deprovisioned`,
-        resourceLabel: this.resourceLabel || '',
+    const { data, errors } = await this.graphqlFetch({
+      query: deleteMutation,
+      variables: {
         resourceId: this.resourceId,
+        ownerId: this.ownerId,
```
dangodev: `ownerId` doesn’t seem to be required, but we’re passing it anyway (we can also overwrite this via a prop if we need to)

CHANGELOG.md
domenicrosati: thanks!
it'd be cool to do some automation here.
A linter that fails if the a componenet is changed but not in the change log.

domenicrosati: github uses: https://hubot.github.com/ quite a bit

PR #607, opened 2019-10-09
Include error messages on performance logs for REST and GraphQL calls

markjamieson:
## Reason for change

Adds any error messages returned from the API to performance logs on REST and GraphQL calls

## Testing

In the storybook performance story, change the `product-label` attr to an invalid label, view logs in Datadog

## Checklist

### DIFF
--- a/src/utils/graphqlFetch.ts
+++ b/src/utils/graphqlFetch.ts
@@ -25,6 +25,13 @@ export interface GraphqlError {
   };
 }
 
+interface GraphqlFetchEventDetail {
+  type: 'manifold-graphql-fetch-duration';
+  request: GraphqlArgs;
+  duration: number;
+  errors?: GraphqlError[];
+}
+
 // TODO remove this in favor of GraphQL documents
 export interface GraphqlData extends Query {
   data?: Resource; // CreateResourcePayload, etc.
@@ -85,7 +92,7 @@ export function createGraphqlFetch({
     }
 
     const fetchDuration = performance.now() - rttStart;
-    const detail = {
+    const detail: GraphqlFetchEventDetail = {
       type: 'manifold-graphql-fetch-duration',
       request,
       duration: fetchDuration,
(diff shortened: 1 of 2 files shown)


src/utils/graphqlFetch.ts
```diff
@@ -84,6 +84,25 @@ export function createGraphqlFetch({
       };
     }
 
+    const fetchDuration = performance.now() - rttStart;
+    const detail = {
+      type: 'manifold-graphql-fetch-duration',
+      request,
+      duration: fetchDuration,
+      errors: body.errors,
+    };
```
dangodev: Is there a type we can use for this? Maybe the `logger` should emit an interface to make sure this is typed correctly

src/utils/restFetch.ts
```diff
@@ -66,28 +66,31 @@ export function createRestFetch({
     }
 
     const body = await response.json();
+    const fetchDuration = performance.now() - rttStart;
+    const message = Array.isArray(body) ? body[0].message : body.message;
```
dangodev: This will be `undefined` in all successful requests that don’t throw an error, I think. Maybe there should be an `if (message)` check?

src/utils/restFetch.ts
```diff
@@ -66,28 +66,31 @@ export function createRestFetch({
     }
 
     const body = await response.json();
+    const fetchDuration = performance.now() - rttStart;
```
dangodev: nit: can there be an extra line here, and maybe a comment, to show this is a logical break from the response parsing?

src/utils/restFetch.ts
markjamieson: True. I didn't think `errorMessage: undefined` was a problem, but I can switch it up

src/utils/graphqlFetch.ts
markjamieson: I've added an interface to both fetch utils

PR #609, opened 2019-10-09
Rename manage button to resize button, convert to GraphQL

dangodev:
## Reason for change

Converts `<manifold-data-manage-button>` to GraphQL. Also renames to `resize` because currently GraphQL only supports plan changes, not feature changes (see this Epic for that scope of work: manifoldco/engineering#9424)

Just trust me this response in the **Network** tab means it’s working:

## Checklist


src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.spec.ts
```diff
@@ -73,7 +69,7 @@ describe('<manifold-data-deprovision-button>', () => {
     });
   });
 
-  describe('events', () => {
```
dangodev: Minor cleanup from the previous PR

src/components/manifold-data-provision-button/manifold-data-provision-button.spec.ts
```diff
@@ -88,7 +84,7 @@ describe('<manifold-data-provision-button>', () => {
       expect(fetchMock.called(`begin:${graphqlEndpoint}`)).toBe(false);
     });
 
-    describe('events', () => {
```
dangodev: Same here: cleanup, and making the 3 similar buttons (provision, resize, deprovision) have more test consistency

stories/resource.stories.js
```diff
@@ -93,7 +89,6 @@ storiesOf('Resource', module)
       <manifold-plan-selector product-label="${productLabel}"></manifold-plan-selector>
       <menu>
         <manifold-data-provision-button
-          owner-id="${ownerId}"
```
dangodev: Cleanup from previous owner PR

src/utils/restFetch.ts
```diff
@@ -83,7 +83,12 @@ export function createRestFetch({
       status: response.status,
     };
 
-    const message = Array.isArray(body) ? body[0].message : body.message;
+    let message: string | undefined;
+    if (Array.isArray(body) && body.length > 0) {
+      message = body[0].message;
+    } else if (typeof body === 'object') {
+      message = body.message;
+    }
```
dangodev: cc @markjamieson was noticing that the previous code threw an error, especially if a REST query returned an empty array (`[]`). This is a little more verbose but this won’t throw any errors.

Fortunately we’re almost entirely off of REST so it only affected a couple components 🎉

stories/data/connectionDecorator.js
```diff
@@ -1,51 +1,19 @@
...
-  const env = radios('env', options, 'prod');
-  const authType = radios('authType', { Manual: 'manual', Oauth: 'oauth' }, 'manual');
+  const env = radios('env', options, localStorage.getItem('manifold_auth_env') || 'prod');
+  const authType = radios(
+    'authType',
+    { Manual: 'manual', OAuth: 'oauth' },
+    localStorage.getItem('manifold_auth_type') || 'manual'
+  );
+  localStorage.setItem('manifold_auth_env', env);
+  localStorage.setItem('manifold_auth_type', authType);
```
dangodev: A little friendly Storybook cleanup to save your **env** and **authType** values across refreshes 🎉

stories/data/connectionDecorator.js
```diff
@@ -1,51 +1,19 @@
...
-        const identityPayload = await fetch('https://api.identity.manifold.co/v1/self', {
-          headers: { authorization: `Bearer ${token}`, 'Content-type': 'application/json' },
-        });
-
-        const { id } = identityPayload.json();
-        localStorage.setItem('manifold_user_id', id);
-      }
-    })();
-  }
-
```
dangodev: Currently didn’t need this anymore with GraphQL

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  afterEach(() => fetchMock.restore());
+
+  describe('v0 props', () => {
+    it('[resource-label]: fetches ID', async () => {
+      await setup({ resourceLabel: 'my-resource' });
+      expect(fetchMock.called(`begin:${graphqlEndpoint}`)).toBe(true);
+    });
+  });
+
+  describe('v0 events', () => {
```
dangodev: ✨ New events tests to make sure these events are actually emitting properly

docs/docs/data/manifold-data-resize-button.md
```diff
@@ -0,0 +1,46 @@
...
+  Change plan
+</manifold-data-resize-button>
+```
+
+## Events
+
+For loading, error and success messages, it will emit custom events.
+
+```js
+document.addEventListener('manifold-resizeButton-click', ({ detail: { resourceLabel } }) =>
```
sslotsky: I've been curious about this naming convention. Why `resizeButton` instead of `resize-button`, given that the latter actually matches the tag name?

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+const graphqlEndpoint = 'http://test.com/graphql';
+
+async function setup(props: Props) {
+  const page = await newSpecPage({
+    components: [ManifoldDataResizeButton],
+    html: '<div></div>',
+  });
+
+  const component = page.doc.createElement('manifold-data-resize-button');
+  component.graphqlFetch = createGraphqlFetch({
```
sslotsky: FWIW you probably only really need to specify the `endpoint` property in these initializers.

docs/docs/data/manifold-data-resize-button.md
dangodev: I’ve been following the paradigm `manifold-[component name]-[present-tense verb]` for events to stay consistent. Makes the event names more predictable, but the naming itself I’m not crazy about. Might make a good lint candidate

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+        throw new Error('button not found in document');
+      }
+
+      const mockClick = jest.fn();
+      await new Promise(resolve => {
+        // listen for event and fire
+        mockClick.mockImplementation(() => resolve());
+        page.doc.addEventListener('manifold-resizeButton-error', mockClick);
+        button.click();
+      });
```
sslotsky: Ah, here we go. I think this would be simpler like this:

```ts
      const mockClick = jest.fn();
      page.doc.addEventListener('manifold-resizeButton-error', mockClick);
      button.click();
      await page.waitForChanges();
```

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      const resourceId = 'success-resource-id';
+
+      const { page } = await setup({ planId, resourceId, resourceLabel });
+      const button = page.root && page.root.querySelector('button');
+      if (!button) {
+        throw new Error('button not found in document');
+      }
+
+      const mockClick = jest.fn();
+      await new Promise(resolve => {
```
sslotsky: Similar thing here. I've never needed to use this trick at all. Just `await page.waitForChanges()` after the `click()` and you should be GTG.

src/components/manifold-data-resize-button/manifold-data-resize-button.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      const button = page.root && page.root.querySelector('button');
+      if (!button) {
+        throw new Error('button not found in document');
+      }
+
+      // listen for event and fire
+      page.doc.addEventListener('manifold-resizeButton-click', mockClick);
+      button.click();
+
+      expect(mockClick).toBeCalledWith(
```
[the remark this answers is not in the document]
dangodev: Well actually I probably should, because this will leave an unfinished promise hanging around. But yeah I was surprised too at first—apparently an async function gets sync priority until the first `await`

stories/resource.stories.js
```diff
@@ -77,11 +80,11 @@ storiesOf('Resource', module)
   })
   .add('create', () => {
     const newResource = text('resource-label', 'pdfshift');
-    const productLabel = select('product-label', productLabels, 'pdfshift');
-    const ownerId = text('owner-id', localStorage.getItem('manifold_user_id') || '');
-    if (ownerId) {
-      localStorage.setItem('manifold_user_id', ownerId);
-    }
+    const productLabel = select(
```
sslotsky: I'm a little confused by all the `productLabel` vs `resourceLabel` stuff going on here, can you explain?

docs/docs/data/manifold-data-resize-button.md
sslotsky: Yeah it's just that `[component name]` as it appears here isn't something that users see anywhere else, right? They only ever see the tag name, i.e. `manifold-data-resize`. So I think `[tag name]-[present tense verb]` makes more sense. Agree though that lint might be a good call.

dangodev: I don’t mind changing it to your suggestion to just use tag names, but Dave and I decided on this way earlier. If it’s alright I’d rather keep this pattern for now because it’s consistent. What about v1 switching to tag events like you suggest?

dangodev: But ideally the React wrappers just come out soon so everyone uses those and doesn’t have to futz with event listeners anymore 😅

sslotsky: Yeah, it's not a change request, just a point of discussion. Something we should definitely take action on later.

stories/resource.stories.js
dangodev: In storybook for the resize, you have to specify which resource label you’d like to resize, and you have to specify the product to pass to the plan selector. Then you resize the plan with both the plan selector and the resize button.

PR #611, opened 2019-10-10
convert resource status to gql

domenicrosati:
## Reason for change

We want to convert this componenet too!

ToDo:
- [ ] did i do this right
- [ ] how can i test this / add tests for this! please teach me

## Checklist


dangodev:
Very nice! Yeah let’s pair tomorrow and write some tests for this.

src/components.d.ts
```diff
@@ -1794,68 +1794,7 @@ export { LocalJSX as JSX };
 
 declare module "@stencil/core" {
   export namespace JSX {
-    interface IntrinsicElements {
```
dangodev: This is a big diff in this file. Try running `npm i` to install the latest Stencil version (should be `1.6.1`)

domenicrosati: Weird!!

src/components/manifold-resource-status/manifold-resource-status.tsx
```diff
@@ -16,14 +16,15 @@ export class ManifoldResourceStatus {
 
   @logger()
   render() {
+    console.log(this.gqlData)
     return (
       <manifold-resource-status-view
         size={this.size}
-        resourceState={this.data && (this.data.state as string)}
+        resourceState={this.gqlData && (this.gqlData.status.label as string)}
```
dangodev: Not sure when this got there but you shouldn’t have to type `as string` for simple things like this; this basically undoes the type safety of TypeScript. You might have to try `this.gqlData && this.gqlData.status && this.gqlData.status.label` but hopefully it’ll just work without `as string`

src/components/manifold-resource-status/manifold-resource-status.tsx
```diff
@@ -16,14 +16,15 @@ export class ManifoldResourceStatus {
 
   @logger()
   render() {
+    console.log(this.gqlData)
```
dangodev: delete this

src/components.d.ts
dangodev: Once you’ve installed the latest Stencil version (maybe rebase from latest `master`), run `npm run build` to regenerate this file. You should only have a couple lines difference in this file, not over a hundred

domenicrosati:
> I’d like to see in manifold-resource-status-view you use the ResourceStatusLabel type from our generated GraphQL types. Replace that object, and only display text straight from GraphQL. Currently, if our GraphQL added labels, we’d have to add them manually to our status component. That is dumb and we should automatically display new statuses from our backend.

@dangodev - manifold-resource-status-view is still being used in manifold-resource-list component (still rest) 

How do you feel if we merge this code (after tests are added), convert manifold-resource-list and then come back to manifold-resource-status-view with the proper types?
Alternatively, we could also mark this code as blocked on manifold-resource-list converstion.

dangodev:
Oh good call. Well how about this: we still remove that manual object and just let `resourceState` display whatever’s passed in? My main concern is that that object blocks some of the new status labels.

You could also transform text case with CSS
```
text-transform: lowercase;

&::first-letter {
  text-transform: uppercase;
}
```

src/components/manifold-resource-status-view/manifold-resource-status-view.tsx
```diff
@@ -46,34 +15,18 @@ export class ManifoldResourceStatusView {
   @Prop() resourceState?: string = OFFLINE;
```
domenicrosati: TBD in resource-list component PR -> change this to ResourceStatusLabel enum rather than string

PR #614, opened 2019-10-10
Convert plan selector to GraphQL

dangodev:
## Reason for change

We’re hiding custom plans until resources support features, but here’s an example of the custom plan interactions working to at least show the work’s done:

### Why dis so big?

I added **many new mocks** that make up the bulk of the diff. On the whole this PR is **mostly deletions.** The working-ness of this PR is evident in 3 ways:

1. All the tests pass
2. We have more plan VRT tests now
3. Play around in Storybook and make sure everything works

After this PR is merged I’m going to do a cleanup PR for some of the leftover mock data that will **result in a lot more deletions.** I just didn’t want to put it in this PR.

### Why did you add more mock data?

For our GraphQL data, I didn’t like that our mocks were transforming REST data. I just pulled GraphQL data straight from our production endpoint, and I’m going to delete all the old mock data we’re not using after this PR merges in a followup. That would require touching tests outside the plan components.

## Testing

**Testing this is actually easy!** Despite the diff, all you must do is test the plan selector in Storybook **WITH ALL THE KNOBS.** If that works, then ✅ this PR.

VRT is doing the bulk of the work for us to ensure it’s working as well; manual testing is just more assurance.

## Checklist


src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -54,19 +52,17 @@ export class ManifoldActivePlan {
...
-        scrollLocked={true}
         isExistingResource={this.isExistingResource}
         plan={this.selectedPlan}
         product={this.product}
         regions={this.regions}
-        resourceFeatures={this.selectedResource && this.selectedResource.features}
         resourceRegion={resourceRegion}
+        scrollLocked={true}
       >
         <manifold-forward-slot slot="cta">
```
dangodev: Changes is this file were pretty minimal and straightforward

src/components/manifold-performance/manifold-performance.e2e.ts
```diff
@@ -3,9 +3,12 @@ import { newE2EPage } from '@stencil/core/testing';
 describe('<manifold-performance>', () => {
   it('should receive manifold-time-to-render event', async () => {
     const page = await newE2EPage();
-    await page.setContent(
-      `<manifold-connection><manifold-performance><manifold-plan product-label="elegant-cms" plan-label="manifold-developer"></manifold-plan></manifold-performance></manifold-connection>`
-    );
+    await page.setContent(`
+      <manifold-connection>
+        <manifold-performance>
+          <manifold-plan plan-id="2351kz26adxddkbewu644wx4b186y"></manifold-plan>
```
dangodev: Now we query plans by ID. This test was updated to match.

src/components/manifold-plan-cost/manifold-plan-cost.tsx
```diff
@@ -1,133 +1,65 @@
...
-                limit
-              }
-            }
-          }
-        }
-      }
-    }
-  }
-`;
-
```
dangodev: The temporary code from the previous plan cost PR was removed here.

src/components/manifold-plan-details/components/BooleanFeature.tsx
```diff
@@ -1,49 +0,0 @@
...
-  // customizable feature
-  return (
-    <manifold-toggle
-      aria-label={feature.name}
-      defaultValue={feature.value && booleanFeatureDefaultValue(feature.value)}
-      name={feature.label}
-      onUpdateValue={(e: CustomEvent) => onChange && onChange(e)}
-    />
-  );
-};
```
dangodev: This component was deleted, and merged into `<ConfigurableFeature />`

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
```diff
@@ -0,0 +1,105 @@
...
+          />
+        </dd>,
+      ];
+
+    default:
+      return [];
+  }
+};
+
+export default ConfigurableFeature;
```
dangodev: This encompasses most of `StringFeature`, `NumberFeature`, and `BooleanFeature`.

src/components/manifold-plan-details/manifold-plan-details.spec.ts
```diff
@@ -1,108 +1,80 @@
...
+            planId: freePlan.id,
+            planLabel: freePlan.label,
+            planName: freePlan.displayName,
+            productId: product.id,
+            productLabel: product.label,
+            regionId: region && region.id,
+            regionName: region && region.displayName,
+          },
+        })
+      );
```
dangodev: This test was improved to listen for event callbacks on the document.

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -261,25 +212,57 @@ export class ManifoldPlanDetails {
...
+                />
+              </div>
+            )}
             <footer class="footer">
-              <manifold-plan-cost
-                defaultCost={defaultCost || cost}
-                planId={this.plan.id}
-                selectedFeatures={this.features}
-              />
+              <manifold-plan-cost plan={this.plan} selectedFeatures={this.features} />
```
dangodev: Obviously the most diffs here are from the different display logic used. This handles a lot of debt suggested at way back in manifoldco/engineering#8104. With these changes, that can be closed.

src/components/manifold-plan-menu/manifold-plan-menu.tsx
```diff
@@ -1,40 +1,84 @@
...
-import { PlanConnection } from '../../types/graphql';
+import { PlanEdge } from '../../types/graphql';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
-import { Catalog } from '../../types/catalog';
 
 @Component({
   tag: 'manifold-plan-menu',
-  styleUrl: 'plan-menu.css',
+  styleUrl: 'manifold-plan-menu.css',
```
dangodev: Since I was touching these components anyway, I made their CSS filenames consistent.

src/components/manifold-plan-menu/manifold-plan-menu.tsx
```diff
@@ -1,40 +1,84 @@
...
   componentWillLoad() {}
 
   @logger()
   render() {
     if (this.plans) {
+      // TODO: re-add configurable plans when GraphQL supports resource features
+      const TEMP_HIDE_CONFIGURABLE_PLANS = this.plans.filter(
+        ({ node: { configurableFeatures } }) =>
+          !configurableFeatures || configurableFeatures.edges.length === 0
+      );
```
dangodev: Temporary code that hides configurable plans

src/data/plan-details-query.ts
```diff
@@ -0,0 +1,65 @@
...
+    edges {
+      node {
+        id
+        displayName
+        platform
+        dataCenter
+      }
+    }
+  }
+`;
```
dangodev: It seemed brittle that the _exact same_ fields had to be requested in `manifold-plan-selector` and `manifold-resource-plan` and `manifold-plan`. So it was abstracted into a single place

src/spec/mock/graphql/product.ts
```diff
@@ -0,0 +1,1066 @@
...
+    },
+  ],
+  valuePropsHtml:
+    '<h3>Aggregate Logs &amp; Analyze Events</h3><p>Centralize your logs in one place instead of tailing one file at a time. Aggregate, search &amp; filter from all your hosts and apps. See related events across your infrastructure on one screen.</p><h3>Powerful Search &amp; Alerts</h3><p>With our blazing fast speeds, search like a pro and set up alerts with just a few clicks. Easily send important event notifications through email, Slack, or even a custom webhook.</p><h3>Easy Setup</h3><p>Get up and running in mere minutes! Spend that saved time building your awesome product instead.</p><h3>Automatic Field Parsing</h3><p>Search for specific fields and values automatically picked up from common log formats, such as JSON, Redis, HTTP access logs, and a whole host of others.</p><h3>Debug &amp; Troubleshoot Faster</h3><p>Diagnose issues, chase down server errors and look up customer activity, fast! Use live streaming tail to instantly surface hard-to-find bugs.</p><h3>Modern User Interface</h3><p>Filter, search, live tail, jump back in time, create views and alerts, all from one interface without leaving the page.</p>',
+};
+
+export default product;
```
dangodev: This (and the other mock files) are where most of the diff comes from. This is raw GraphQL data straight from our API, no mapping. I’d like to delete the REST mock data in a followup PR after this

package.json
```diff
@@ -68,7 +68,7 @@
     },
     {
       "path": "./dist/esm/manifold-!(mock|number-input)*.js",
-      "maxSize": "32 kB",
+      "maxSize": "16 kB",
```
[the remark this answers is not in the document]
dangodev: 🎉 🎉 🎉

src/components/manifold-region-selector/manifold-region-selector.tsx
```diff
@@ -1,113 +0,0 @@
...
-      <manifold-select
-        aria-label={this.ariaLabel}
-        name={this.name}
-        onUpdateValue={this.handleChange}
-        defaultValue={this.value}
-        options={this.regionOptions(regions)}
-      />
-    );
-  }
-}
```
dangodev: The region selector we can just delete now that we’re no longer on REST. We’re just using a dropdown straight in `<manifold-plan-details>`

src/components/manifold-resource-plan/manifold-resource-plan.spec.ts
```diff
@@ -1,124 +1,53 @@
...
-    `);
+describe('<manifold-resource-plan>', () => {
+  it('Renders a skeleton if loading', async () => {
+    const { page } = await setup({ loading: true });
+    const planDetails = page.root && page.root.querySelector('manifold-plan-details');
+    const skeleton =
+      planDetails &&
+      planDetails.shadowRoot &&
+      planDetails.shadowRoot.querySelector('manifold-skeleton-text');
+    expect(skeleton).not.toBeNull();
```
dangodev: This test broke because it prevented any markup changes from happening.

I changed these tests to be a little more forgiving while keeping their original intent.

package.json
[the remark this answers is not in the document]
sslotsky: :clap: :clap: :clap:

src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -1,29 +1,27 @@
...
     if (this.selectedResource && this.selectedResource.plan && this.selectedResource.plan.id) {
       this.selectPlan(this.selectedResource.plan.id);
-    } else {
-      this.selectPlan(newPlans[0].id);
+    } else if (newPlans && newPlans.length > 0) {
+      this.selectPlan(newPlans[0].node.id);
     }
   }
-  @Watch('selectedResource') resourceChange(newResource: Gateway.Resource) {
+  @Watch('selectedResource') resourceChange(newResource: Resource) {
```
sslotsky: This is not an immediate request, but do you think that this component is a good candidate for a functional component? My memory (unreliable of course) says that this is one that we separated out to make Happo happy, before we understood how to test smart components. If we don't actually need to expose this to customers, can we make a ticket for changing this into a functional component later?

src/components/manifold-cost-display/manifold-cost-display.tsx
```diff
@@ -49,19 +49,9 @@ const MeteredCost: FunctionalComponent<MeteredCostProps> = ({ meteredFeatures })
...
-    return [cost, <small>&nbsp;per&nbsp;{suffix}</small>];
-  }
-
-  if (displayString.indexOf('/') > 0) {
-    const [cost, suffix] = displayString.split(' / ');
-    return [cost, <small>&nbsp;/&nbsp;{suffix}</small>];
-  }
-
-  return displayString;
+  return [displayValue.cost, displayValue.per ? <small>&nbsp;{displayValue.per}</small> : ''];
```
sslotsky: :clap: :clap: :clap:

src/components/manifold-active-plan/manifold-active-plan.tsx
dangodev: Yes! Let’s do that as a followup. Definitely on-board with more functional components in general. I don’t like creating web components unless we want to expose them (but I know we must in some cases).

I removed a lot of functional components in general in this PR because I was trying to pay down the complexity and number of files the plan selector needed. But in general I like the idea of fewer web components, more functional components.

src/components/manifold-number-input/manifold-number-input-happo.ts
```diff
@@ -13,7 +13,7 @@ export const atMin = () => {
   const input = document.createElement('manifold-number-input');
   input.max = 42;
   input.value = 0;
-  input.suffix = 'Marklars';
+  input.unit = 'Marklars';
```
sslotsky: marklar

src/components/manifold-service-card/manifold-service-card.e2e.ts
```diff
@@ -1,91 +0,0 @@
...
-        { product: Product }
-      );
-      await page.waitForChanges();
-
-      const card = await page.find('manifold-service-card-view');
-      const preserveEvent = await card.getProperty('preserveEvent');
-      expect(preserveEvent).toBe(true);
-    });
-  });
-});
```
dangodev: I don’t know why, but at the very end this test decided to fail (seemed to be some change to Stencil’s test page that caused an error—our actual tests weren’t even running, not technically failing). Replaced it with a Happo test and touched up the spec test. I never planned on touching this, but 🤷‍♂

src/components/manifold-performance/manifold-performance.e2e.ts
sslotsky: Thank you for formatting!

src/components/manifold-service-card/manifold-service-card.e2e.ts
dangodev: Update: this will be handled in https://github.com/manifoldco/engineering/issues/9484

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
```diff
@@ -0,0 +1,105 @@
...
+    // boolean
+    case PlanFeatureType.Boolean:
+      return [
+        <dt class="feature-name">{displayName}</dt>,
+        <dd class="feature-value">
+          <manifold-toggle
+            aria-label={displayName}
+            defaultValue={!!value}
+            name={label}
+            onUpdateValue={(e: CustomEvent) => onChange && onChange(e)}
```
sslotsky: It looks like `onChange` is a required prop, so you probably wouldn't need to check for its existence right? And actually I think that passing in `undefined` for this works anyway, so even if it were an optional prop, you could still do:

```
onUpdateValue={onChange}
```

Just a suggestion, I know you don't want to change any code in this PR though. Maybe a follow-up.

src/components/manifold-plan-menu/manifold-plan-menu.tsx
```diff
@@ -1,22 +1,19 @@
...
+  styleUrl: 'manifold-plan-menu.css',
   shadow: true,
 })
 export class ManifoldPlanMenu {
-  @Prop() plans?: PlanConnection;
-  @Prop() oldPlans: Catalog.ExpandedPlan[] = [];
+  @Prop() plans?: PlanEdge[];
   @Prop() selectedPlanId?: string;
-  @Prop() selectPlan: Function = () => {};
+  @Prop() selectPlan: (planId: string) => void = () => null;
```
sslotsky: Huh. I'm surprised that this passes type checking. `null` should not be the same as `void` :thinking:

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -1,106 +1,61 @@
...
-const query = gql`
+const plansQuery = gql`
   query PLAN_LIST($productLabel: String!) {
     product(label: $productLabel) {
       id
       displayName
       label
       logoUrl
-      plans(first: 500, orderBy: { field: COST, direction: ASC }) {
+      plans(first: 25, orderBy: { field: COST, direction: ASC }) {
```
sslotsky: Hmm. I see we're not accounting for pagination here, and we really should be. Can you do this in a follow-up PR? I'm pretty sure that `fetchAllPages` will support this query but I'm happy to pair on it.

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -1,106 +1,61 @@
...
-    edges: product.plans ? product.plans.edges.filter(plan => plan.node.free) : [],
-  };
+const freePlansQuery = gql`
+  query FREE_PLAN_LIST($productLabel: String!) {
+    product(label: $productLabel) {
+      id
+      displayName
+      label
+      logoUrl
+      plans(first: 25, free: true, orderBy: { field: COST, direction: ASC }) {
```
sslotsky: I'm not sure I understand why we need to do two separate queries to get free plans. Can't we get them all together?

src/data/plan-details-query.ts
```diff
@@ -0,0 +1,65 @@
+import { gql } from '@manifoldco/gql-zero';
+
+export default gql`
+  id
+  displayName
+  label
+  free
+  cost
+  fixedFeatures(first: 25) {
```
sslotsky: Again, this is a bit of a problem. We're not supposed to be querying connection data without accounting for pagination. This was made very clear by both James and Jelmer.

At the same time, this might be difficult to solve with the current `fetchAllPages` implementation, because it seems that there are multiple nested connections that we need to account for. `fetchAllPages` assumes that you're collecting all the pages of a single connection.

So, we might have to iterate on this, but it needs to be a priority.

dangodev: Does `fetchAllPages` work for child queries within the same request?

Also in these instances, I omitted it because all of these fields are (for now) within our default page size (`25`)

dangodev:
> I got fatigued before getting through even half of these changes. I thought the plan for the first iteration was to get all the GraphQL data, convert it back to REST format, and make nothing more than minimal changes to child components (e.g. remove fetching where it's not needed, but not much beyond that). Was it not possible to do it this way? I really hoped to avoid PRs that can't be thoroughly reviewed.

Yeah, that was something that I tried to avoid in this PR. But the extreme diffs came from the reality that all of these components—manifold-plan-selector, manifold-active-plan, manifold-plan-details, and manifold-plan-menu—are all one “component.” They were all written and updated around the same time, and had structural issues that reflected the REST structure (such as features all being together, rather than separated into fixed/metered/customizable). As always, restructuring the feature data (which is most of this component) was a heavy task.

I think we did much better by only refactoring the plan-* family in this PR, but I think all these changes were necessary when you include the test changes that come with it. Perhaps this could have been broken up into 2 PRs, but I think one of them would have taken up most of the difference anyway. Again, at least this only touches the plan components, and not _everything_ like the previous PR.

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
dangodev: Oh yeah I just kept that as-is. You’re right—that seems fine to change in a followup.

src/data/plan-details-query.ts
sslotsky: I think if the connection is a child query, your strategy would look like this:

1. Get the initial payload, like you're already doing, but make sure to use the `after` variable and include the `endCursor` for the connection that you're targeting.
1. Then, once the initial payload comes back, use `fetchAllPages` to get the remaining pages of the target connection. You'd use the same query, and you'd leverage the `getConnection` argument to the `fetchAllPages` function in order to tell it how to collect the edges of the connection that you're targeting.
1. Once you have the remaining pages from your target connection, you can append all those edges to your initial payload.

What's less clear to me is whether or not we'd be able to retrieve nested connections. It might depend on what the API supports.

> Also in these instances, I omitted it because all of these fields are (for now) within our default page size (25)

As for that, the same could be said about fetching our catalog. We knew we had 42 products, so we could have set the limit to 1000. But it was made clear to me that we don't want to do stuff like that. If data is paginated, we absolutely have to account for it, 100% of the time.

dangodev: Got it. Yeah let’s do this, but in a followup if possible. There may be some revs on this.

src/components/manifold-plan-selector/manifold-plan-selector.tsx
dangodev: Oh looks like we can with aliases! Happy to do that in a followup

dangodev:
Followup items: https://github.com/manifoldco/engineering/9607

PR #621, opened 2019-10-12
update storybook to support local env

domenicrosati:
## Reason for change

Because we want to be able to exploritory test our web components against local environments.

## Checklist


stories/data/retrieve-data.js
```diff
@@ -0,0 +1,29 @@
+// TODO(dom): we can get this data from graphql now, so why not do it?
```
dangodev: Storybook can’t render async so it needs to be cached. We can add a generate step tho

stories/catalog.stories.js
```diff
@@ -26,7 +24,9 @@ storiesOf('Catalog', module)
     ></manifold-marketplace>`;
   })
   .add('product page', () => {
-    const productLabel = select('product-label', productLabels, 'jawsdb-mysql');
+    const env = localStorage.getItem('manifold_auth_env');
```
dangodev: We can set this at the top of the file rather than duplicating it

stories/data/retrieve-data.js
domenicrosati: ah k will add this to the todo

stories/catalog.stories.js
domenicrosati: No because when we switch to each page the variable doesn't change.
We add it on page change so that we can keep the values of the selector up to date.

domenicrosati:
Happo is failing on manifold-resource-plan but the diff its showing is not correct :(

Screen Shot 2019-10-12 at 7 13 21 PM

dangodev:
@domenicrosati Ugh yeah that Happo diff seems to be from a Stencil update.

You should approve the diff in Happo, though, and not merge it if it doesn’t pass. Click **details** next to it and select Approve.

PR #630, opened 2019-10-15
Add optional ownerID to manifold-data-provision-button

markjamieson:
## Reason for change

Closes https://github.com/manifoldco/engineering/issues/9651
Adds `ownerId` prop to `manifold-data-provision-button` and includes ID on createResource gql call

## Testing

1. Refer to https://app.gitbook.com/@manifold/s/engineering/testing-guides/exploratory-testing-with-webcomponents-using-storybook to find the ownerID attached to your token
1. In the Resource -> create story I have added a new ownerID field to the knobs panel. Try creating resources with:
    - No ownerID
    - A valid ownerID
    - An invalid owner ID
1. The first two should return good data, the final one should error

## Checklist

### DIFF
--- a/docs/docs/data/manifold-data-provision-button.md
+++ b/docs/docs/data/manifold-data-provision-button.md
@@ -106,3 +106,13 @@ plain ol’ CSS.
 [shadow-dom]: https://developers.google.com/web/fundamentals/web-components/shadowdom
 [slot]: https://stenciljs.com/docs/templating-jsx/
 [plan-selector]: /components/plan-selector
+
+## Owner ID
+
+You can pass a platform specific owner ID to be attached to the created resource like this:
+
+```html
+<manifold-data-provision-button owner-id="user-1234">
+  Provision Resource
+</manifold-data-provision-button>
+```
--- a/docs/docs/data/manifold-data-provision-button.md
+++ b/docs/docs/data/manifold-data-provision-button.md
@@ -107,12 +107,13 @@ plain ol’ CSS.
 [slot]: https://stenciljs.com/docs/templating-jsx/
 [plan-selector]: /components/plan-selector
 
-## Owner ID
+## Context (team, org, etc.)
 
-You can pass a platform specific owner ID to be attached to the created resource like this:
+By default, Manifold assumes the user creating the resource will own it. But 
+when creating a resource owned by another context (e.g. a team, or 
+organization, or whatever your platform calls it), you’ll need to specify an 
+owner-id different from their own:
 
 ```html
-<manifold-data-provision-button owner-id="user-1234">
-  Provision Resource
-</manifold-data-provision-button>
+<manifold-data-provision-button owner-id="team-123">
 ```
--- a/src/components/manifold-data-provision-button/create.graphql
+++ b/src/components/manifold-data-provision-button/create.graphql
@@ -6,9 +6,7 @@ mutation CreateResource($planId: ID!, $productId: ID!, $regionId: ID!, $resource
       planId: $planId
       productId: $productId
       regionId: $regionId
-      owner: {
-        id: $ownerId
-      }
+      ownerId: $ownerId
     }
   ) {
     data {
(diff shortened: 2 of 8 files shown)


src/components/manifold-data-provision-button/create.graphql
```diff
@@ -6,6 +6,9 @@ mutation CreateResource($planId: ID!, $productId: ID!, $regionId: ID!, $resource
       planId: $planId
       productId: $productId
       regionId: $regionId
+      owner: {
```
seriousben: I think it is just `ownerId` in the input. So just `ownerId: $ownerId` would work I believe.

markjamieson: Oh I see, I was looking at the data payload in the graphql explorer, and not the input. I am still a gql query novice.

markjamieson:
@domenicrosati This _looks_ good but I'm having trouble getting that profile query to work. In the schema it looks like id is a required argument. Not sure how to get it to work with only an authorization header.

domenicrosati:
Uh oh that's not good at all ill have a look tomorrow. It should not require that

src/components/manifold-data-provision-button/create.graphql
```diff
@@ -6,6 +6,7 @@ mutation CreateResource($planId: ID!, $productId: ID!, $regionId: ID!, $resource
       planId: $planId
       productId: $productId
       regionId: $regionId
+      ownerId: $ownerId
```
dangodev: This will have to conditionally be passed or not based on whether the prop exists. We might need to queries for this I think.

markjamieson: Really? All of the variables used in the query are optional component props. Are the rest of them guaranteed to exist when this query is called?

markjamieson: Well I can't argue with results, I definitely get errors when I don't use multiple queries. Fixing now.

dangodev: Oh I more meant that if you try and pass a parameter to a mutate function that’s undefined, our GraphQL endpoint will throw an error. And this should work if `owner-id` is missing. It’s not really a TypeScript thing.

So you may need 2 queries, one with and one without `owner-id`. But really whatever you can get away with with our GraphQL server is fine

markjamieson:
@dangodev Could use a re-review on this. Didn't add a unit test because the only unit testable thing is the conditional query, and that seemed unnecessary to test.

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -66,6 +67,8 @@ export class ManifoldDataProvisionButton {
   @Prop() graphqlFetch?: GraphqlFetch = connection.graphqlFetch;
   /** Product ID */
   @Prop({ mutable: true }) productId?: string = '';
+  /** Owner ID */
+  @Prop() ownerId?: string;
```
dangodev: Maybe add a section to the provider button docs showing this is an option?

docs/docs/data/manifold-data-provision-button.md
```diff
@@ -106,3 +106,13 @@ plain ol’ CSS.
 [shadow-dom]: https://developers.google.com/web/fundamentals/web-components/shadowdom
 [slot]: https://stenciljs.com/docs/templating-jsx/
 [plan-selector]: /components/plan-selector
+
+## Owner ID
+
+You can pass a platform specific owner ID to be attached to the created resource like this:
```
dangodev: Maybe change the following:

>## Context (team, org, etc.)
>
>By default, Manifold assumes the user creating the resource will own it. But when creating a resource owned by another context (e.g. a team, or organization, or whatever your platform calls it), you’ll need to specify an `owner-id` different from their own:
>
>
>```html
><manifold-data-provision-button owner-id="team-123">
>```

markjamieson: Perfect, thank you!

PR #628, opened 2019-10-15
Add TestCafe and a test for filtering in marketplace

sslotsky:
## Reason for change

Because we want automated, full-stack, end-to-end testing!

This introduced TestCafe along with a test against our StoryBook site that does some filtering on the marketplace component.

This is a work in progress. The `testcafe` task will currently only work if you have StoryBook running locally at `localhost:6006`. The next step would be to try and integrate this into CI, so that it would run against our deployed storybook instance immediately after we deploy it.

## Checklist

- [ ] **CHANGELOG**: The **Unreleased** section of CHANGELOG was updated
- [ ] **Prop changes**: [docs][docs] were updated
- [ ] **Prop changes**: E2E tests were updated, testing from the highest level possible
- [ ] **Platform testing**: If this change should be [tested against a platform][platform-testing], has it been?

[docs]: https://ui.sandbox.manifold.co
[platform-testing]: https://app.gitbook.com/@manifold/s/engineering/playbooks/testing-your-code-against-a-platform

PR #637, opened 2019-10-16
TestCafe - single test with GitHub action

sslotsky:
Resubmission of #628 
## Reason for change

To enable us to write e2e tests against our Storybook environment, and have those tests run against our preview URL as part of our CI process.

## Checklist


e2e/catalog-test.ts
```diff
@@ -0,0 +1,22 @@
...
+  );
+  const title = Selector(() =>
+    document
+      .querySelector('manifold-marketplace-grid')
+      .shadowRoot.querySelector('manifold-service-card-view')
+      .shadowRoot.querySelector('h3.name')
+  );
+
+  await t
+    .switchToIframe(iframe)
```
dangodev: Is this required because of Storybook? I think you can just visit the iframe URL directly (`https://ui.manifold.now.sh/iframe.html?id=catalog--marketplace`)

sslotsky: Oh, that would be handy! Thanks, I'll try it out!

sslotsky:
@domenicrosati @dangodev this is ready to go.

[The GH action is passing](https://github.com/manifoldco/ui/commit/0c774b494a6824ba2915fc8fd1b1e90f0139b536/checks?check_suite_id=294895494) but it doesn't currently show up on the list of checks. I think this will fix itself when the branch is merged, and then we can switch it to required.

There are some follow up items that I'll want to do as well, but I think this is good for now.

dangodev:
Awesome! Is the goal for this to be a PR blocker after merge? Is there any user action required on our part if it fails, and should this be in the docs somewhere?

PR #638, opened 2019-10-16
convert resource container to gql

domenicrosati:
## Reason for change

We want to only use graphQL for this component.

## Testing

Check docs and storybook for the things that use this component.

## Checklist

- [X] **CHANGELOG**: The **Unreleased** section of CHANGELOG was updated
- [N/A ] **Prop changes**: [docs][docs] were updated
- [N/A] **Prop changes**: E2E tests were updated, testing from the highest level possible
- [N/A] **Platform testing**: If this change should be [tested against a platform][platform-testing], has it been?


src/components/manifold-resource-container/manifold-resource-container.spec.ts
```diff
@@ -105,20 +99,18 @@ describe('<manifold-resource-credentials>', () => {
...
 
   it('will refetch the resource after load if it received an error', async () => {
     const resourceLabel = 'test-resource';
     // @ts-ignore
-    page.win.setTimeout = jest.fn(call => call());
-    fetchMock
-      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, 404)
-      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, GatewayResource);
+    page.win.setTimeout = jest.fn(call => {
+      GraphqlResource.data.resource.status.label = ResourceStatusLabel.Available;
```
sslotsky: I'm really confused by all these assignments, can you explain what's going on?

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -161,36 +158,31 @@ export class ManifoldResourceContainer {
...
-      }
-    } catch (error) {
-      // In case we actually want to keep fetching on an error
+    if (errors) {
       if (this.refetchUntilValid) {
         this.timeout = window.setTimeout(() => this.fetchResource(this.resourceLabel), 3000);
       }
-
-      console.error(error);
+      console.error(errors);
```
dangodev: I don’t think this is needed anymore. We handle it in the `graphqlFetch` component.

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -161,36 +158,31 @@ export class ManifoldResourceContainer {
...
+    if (!this.graphqlFetch || !this.resourceLabel) {
       return;
     }
 
-    try {
-      const response = await this.restFetch<Gateway.Resource & { product?: Product }>({
-        service: 'gateway',
-        endpoint: `/resources/me/${resourceLabel}`,
-      });
+    const { data, errors } = await this.graphqlFetch({ query, variables: { resourceLabel } });
```
dangodev: Is it possible to use Sam’s generated types here?

domenicrosati: yes!

domenicrosati: Mocking is inconvient with them because you need to write a mock per query but that isn't the worst

src/components/manifold-resource-container/manifold-resource-container.spec.ts
```diff
@@ -65,38 +69,32 @@ describe('<manifold-resource-credentials>', () => {
...
   it('will refetch the resource after load if given an invalid resource', async () => {
     const resourceLabel = 'test-resource';
+    // In order to test refetching resources we need to set the resource status label
+    // to available when the first timeout is called.
     // @ts-ignore
-    page.win.setTimeout = jest.fn(call => call());
-    fetchMock
-      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, {})
-      .once(`${connections.prod.gateway}/resources/me/${resourceLabel}`, GatewayResource);
+    page.win.setTimeout = jest.fn(call => {
```
sslotsky: All right, I understand. What I'd be tempted to try here is something more like this:

1. Reassign `setTimeout` in a `beforeEach`, and instead of using a mock, just make it call the callback without no wait time.
2. In your test, set up a second `fetchMock` call so that it returns modified data the second time.

To me that seems to more accurately simulate what's going on, and allows you to avoid both `ts-ignore` and reassignments. What do you think?

dangodev:
Screen Shot 2019-10-16 at 13 02 28

When I test this in Storybook, everything loads but my resource shows up as “offline”?

domenicrosati:
That will be fixed when resource-list is in

src/components/manifold-resource-container/manifold-resource-container.tsx
domenicrosati: @dangodev I tried to do it and it wouldn't work :( I wonder if its because I am trying to assign to this.graphQL ? I am not sure but I couldn't make the types match.

dangodev:
When I don’t authenticate, I don’t see anything for the resource container—it just hides everything. This seems to contrast with the resource-list that still shows the skeleton loader when authentication fails—other components show _something_ even when the call fails.

Can you try to make sure the skeleton loader shows when authentication fails? At least then it’s consistent with the other components, and it’s apparent something is on the page. We’ll have design audit the error states later to make sure this is what we want, but IMO if there’s an error all the components shouldn’t leave the page.

domenicrosati:
@dangodev I had a look into it - and it looks like that behaviour was not introduced in this PR so I am going to merge and raise a ticket for that.

PR #639, opened 2019-10-16
Remove core-js

dangodev:
## Reason for change

This got added to fix… something. I can’t remember what (it might have been an old Stencil version). But now it’s no longer needed for us to manage `core-js` ourselves. Let’s reduce the dependabot noise if we can.

## Checklist


dangodev:
> Fine with me as long as tests pass!

Yeah a few things still require `core-js`. I just remember adding it to `package.json` to fix something at some point but don’t remember what now. This effectively shouldn’t change anything; just will save us from some dependabot pings

PR #642, opened 2019-10-16
Manifold-credentials Not Working On Docs

TerrySmithDC:
## Reason for change

The `manifold-credentials` component was not not working in the example field on the docs site. Also updating the graphql call to the more type safe style.

## Testing

I tested this local running `npm run docs` on the Resource Credentials page. Should look like this.

Screen Shot 2019-10-17 at 4 14 11 PM

## Checklist


docs/src/utils/mockGraphql.ts
```diff
@@ -8,11 +8,26 @@ if (typeof window !== 'undefined') {
 
 const waitForDuration = (time: number) => new Promise(resolve => setTimeout(resolve, time));
 
+const credentialResourceMock = {
+  data: {
+    resource: {
+      credentials: {
+        edges: [{ node: { key: 'API_KEY', value: 'xxxxxxxxxxxxxxxxxxxxxxxxx' } }],
```
domenicrosati: its nit picky but want to add two values? just to folks can see what its like to have two creds?

src/components.d.ts
```diff
@@ -612,355 +612,355 @@ declare global {
...
   interface HTMLManifoldToggleElement extends Components.ManifoldToggle, HTMLStencilElement {}
-  const HTMLManifoldToggleElement: {
+  var HTMLManifoldToggleElement: {
     prototype: HTMLManifoldToggleElement;
     new (): HTMLManifoldToggleElement;
   };
 
   interface HTMLManifoldTooltipElement extends Components.ManifoldTooltip, HTMLStencilElement {}
-  const HTMLManifoldTooltipElement: {
+  var HTMLManifoldTooltipElement: {
```
domenicrosati: @dangodev why did these go from const -> var ?

dangodev: @TerrySmithDC needs to run `npm i` to update Stencil to the latest version.

src/components/manifold-credentials/resourceCredentials.graphql
```diff
@@ -0,0 +1,12 @@
...
+        credentials(first: 25) {
+            edges {
+                node {
+                    key
+                    value
+                }
+            }
+        }
+    }
+}
```
dangodev: The indentation for this seems wrong. I’d recommend installing the **Prettier** VS Code Extension and turning on **Auto-format on save** in VS Code. But that’s on me for not linting `.graphql` files.

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -33,32 +33,20 @@ export class ManifoldCredentials {
...
-                  key
-                  value
-                }
-              }
-            }
-          }
-        }
-      `,
+    const { data, errors } = await this.graphqlFetch<Resource_CredentialsQuery>({
+      query: resourceCredentialsQuery,
```
dangodev: Nice!

src/components/manifold-credentials/resourceCredentials.graphql
TerrySmithDC: Hmm I do have prettier, not sure why it's not picking up for `.graphql` 🤔  Looking into it

docs/src/utils/mockGraphql.ts
TerrySmithDC: Sure

codegen.yml
```diff
@@ -1,6 +1,9 @@
 overwrite: true
 schema: 'https://api.manifold.co/graphql'
 documents: src/**/*.graphql
+config:
+  avoidOptionals: false # generate optionals (?)
+  maybeValue: T # use optionals to determine nullability, not Maybe
```
dangodev: cc @domenicrosati @TerrySmithDC this is required to be able to use `ResourceCredentialsQuery['resource']['credentials']['edges']`.

You may think: won’t this mean TypeScript may not catch a null reference error? Fortunately, this will still return the top level as possibly undefined. So you can’t use `data.resource` without an error.

But if you check for `if (data.resource) { …`, then you can use `data.resource.credentials…` without problem. IMO this is actually better.

src/components/manifold-credentials-view/manifold-credentials-view.tsx
```diff
@@ -1,7 +1,7 @@
 import { h, Component, Prop, Event, EventEmitter, Element, State } from '@stencil/core';
 import { eye, lock, loader } from '@manifoldco/icons';
 
-import { CredentialEdge } from '../../types/graphql';
+import { ResourceCredentialsQuery } from '../../types/graphql';
```
dangodev: This is actually a good example of our generated types working! TypeScript threw an error because **it knew we weren’t querying anything.** It was trying to prevent us from calling something like `data.resource.credentials.cursor` when it didn’t exist 💯

But in true TypeScript fashion, the error message was garbage.

Anyway, moving forward we should only be importing generated query types like this, and we should move away from using the full types like we were doing.

src/components/manifold-credentials-view/manifold-credentials-view.tsx
```diff
@@ -12,7 +12,7 @@ import loadMark from '../../utils/loadMark';
 })
 export class ManifoldCredentialsView {
   @Element() private el: HTMLElement;
-  @Prop() credentials?: CredentialEdge[];
+  @Prop() credentials?: ResourceCredentialsQuery['resource']['credentials']['edges'];
```
dangodev: This is the only weird part—having to drill into the query. I couldn’t find another way to do this—this seems to be a limitation of graphql-codegen. This works, but if I come up with a better way, I’ll let you know.

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -44,9 +44,8 @@ export class ManifoldCredentials {
       this.errors = errors;
     }
 
-    if (data && data.resource && data.resource.credentials && data.resource.credentials.edges) {
-      const fetchCredentials = data.resource.credentials.edges;
-      this.credentials = fetchCredentials as CredentialEdge[];
+    if (data) {
+      this.credentials = data.resource.credentials.edges;
```
dangodev: No more annoying checks once you get past `data` 🎉

src/components/manifold-credentials/manifold-credentials.tsx
```diff
@@ -13,7 +13,7 @@ export class ManifoldCredentials {
   /** _(hidden)_ */
   @Prop() graphqlFetch?: GraphqlFetch = connection.graphqlFetch;
   @Prop() resourceLabel?: string = '';
-  @State() credentials?: CredentialEdge[];
+  @State() credentials?: ResourceCredentialsQuery['resource']['credentials']['edges'];
```
dangodev: This is the parent component’s type. Notice that **we had to change the child’s types as well.**

I’m actually super happy about this, because now TypeScript can alert us if we’re using something we didn’t query for. And that’s the goal.

src/types/graphql.ts
```diff
@@ -236,14 +236,25 @@ export type InvoiceEdge = {
   node?: Maybe<Invoice>,
 };
 
-/** InvoiceOrderBy defines how an invoice connection is to be ordered. */
+/** InvoicePreviewOrderBy defines how an invoice connection is to be ordered. */
```
dangodev: Ha glad this was in your PR! I’m fixing the GraphQL schemas now and this seems like a typo. I’ll fix.

docs/src/utils/mockGraphql.ts
```diff
@@ -13,6 +13,20 @@ const resourcesMock = {
   },
 };
 
+const credentialResourceMock = {
+  data: {
+    resource: {
+      credentials: {
+        edges: [
+          { node: { key: 'API_KEY', value: 'xxxxxxxxxxxxxxxxxxxxxxxxx' } },
+          { node: { key: 'BEST_FRUIT', value: 'Kiwi' } },
```
dangodev: Truth

docs/src/utils/mockGraphql.ts
```diff
@@ -13,6 +13,20 @@ const resourcesMock = {
...
+          { node: { key: 'BEST_VEGTABLE', value: 'Cabbage' } },
```
dangodev: Debatable

src/components/manifold-credentials-view/manifold-credentials-view.tsx
domenicrosati: Ideally! There would be sub types.
So ResourceCredentialsQuery could be it's own name space with resource, credentials, and credentials edges as their own generated types embedded in the main type for that namespace, resoucecredentialsquery. Can we do that?

dangodev: Yeah that’s specifically what I was referring to. I couldn’t figure out a way to generate that with `documents` in graphql-codegen.

domenicrosati: Ah well :shurg: I am fine with this then

src/components/manifold-credentials-view/manifold-credentials-view.tsx
TerrySmithDC: > But in true TypeScript fashion, the error message was garbage.

This is the realest thing I've read all morning.

docs/src/utils/mockGraphql.ts
dangodev: Oh also typo: `VEGETABLE` (didn’t catch that yesterday)

dangodev:
Looks like the Happo test is failing, because Happo runs strict type checks on everything. Because we changed that component’s data structure, Happo is picking up on that.

Not sure what the quickest way to handle that would be. IMO you have a few options:

1. Use the generated `ResourceWithCredentialsQuery` type in `manifold-credentials-view-happo.ts` and use your own mock data in that file, set to that generated type (may be simplest because nothing else will use that, so why clutter things up?)
1. Create some mock folder where for every generated query, there’s mock data to go with it (this seems like a pain to me)
1. Just force the existing mock data into the generated `ResourceWithCredentialsQuery` type. IMO I don’t like this because this could cause other issues.

PR #644, opened 2019-10-18
Region test

dangodev:
## Reason for change

We had a `regions` prop in `<manifold-plan-selector>` that was in our docs, but there was no test for it. This adds a test.

## Testing

Test should pass.

## Checklist


stories/catalog.stories.js
```diff
@@ -39,10 +39,13 @@ storiesOf('Catalog', module)
...
+    const regions = text('regions', '');
     return `
-    <manifold-plan-selector product-label="${productLabel}" ${
-      freePlans ? 'free-plans' : ''
-    }></manifold-plan-selector>`;
+    <manifold-plan-selector
+      ${freePlans ? 'free-plans' : ''}
+      product-label="${productLabel}"
+      regions="${regions}"
+    ></manifold-plan-selector>`;
```
dangodev: Adds Storybook test

src/data/plan-details-query.ts
```diff
@@ -52,7 +52,7 @@ export default gql`
       }
     }
   }
-  regions(first: 25) {
+  regions(first: 25, orderBy: { field: DISPLAY_NAME, direction: ASC }) {
```
dangodev: We weren’t sorting regions before. Now we are!

src/components/manifold-select/manifold-select.tsx
```diff
@@ -5,8 +5,8 @@ import loadMark from '../../utils/loadMark';
 
 @Component({
   tag: 'manifold-select',
-  styleUrl: 'mf-select.css',
-  scoped: true,
+  styleUrl: 'manifold-select.css',
+  shadow: true,
```
dangodev: I think I remember this—we were testing `scoped` wayy back. `shadow` is more standardized, and it makes testing easier. There’s no downside to this.

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -162,17 +156,16 @@ export class ManifoldPlanSelector {
...
 
-  parseRegions(regions: string) {
-    return regions.split(',').map(region => region.trim().toLowerCase());
-  }
-
   @logger()
   render() {
+    const regions =
+      (this.regions && this.regions.split(',').map(region => region.trim())) || undefined;
+
```
dangodev: Getting rid of `Watch` simplifies this component and reduces the possibility for error. The test will catch if this logic here isn’t working.

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -59,7 +59,6 @@ const resourceQuery = gql`
 
 @Component({ tag: 'manifold-plan-selector' })
 export class ManifoldPlanSelector {
-  @Element() el: HTMLElement;
```
dangodev: These were added way back in the state tunnel days. If it’s not apparent you need `@Element`, you don’t.

src/components/manifold-plan-details/manifold-plan-details.spec.ts
```diff
@@ -50,7 +55,7 @@ describe('<manifold-plan-details>', () => {
     });
 
     it('change', async () => {
-      const { page, component } = await setup();
+      const { page, component } = await setup({});
```
dangodev: I initially added the regions test here, but then moved it higher up to test from the top level. These changes are just noise now, but they do make it easier to add more tests to this file, so might just keep it.

codecov.yml
```diff
@@ -0,0 +1,8 @@
+ignore:
+  - docs/
+  - scripts/
+  - src/spec
+  - stories/
+  - *-happo.ts
+  - *.spec.ts
+  - *.e2e.ts
```
dangodev: So I just noticed that coverage went _down_ after adding a mock and I have a sneaking suspicion we’re not ignoring enough in codecov

src/data/plan-details-query.ts
domenicrosati: hooray

src/components/manifold-select/manifold-select.tsx
domenicrosati: :shadow:

src/components/manifold-plan-selector/manifold-plan-selector.tsx
domenicrosati: lgtm

PR #645, opened 2019-10-18
Use ownerId in manifold-data-resource-list and manifold-resource-list

markjamieson:
## Reason for change

In both `<manifold-data-resource-list>` and `<manifold-resource-list>`:
- add ownerId prop
- include owner argument to `resources` graphql query

## Testing

Per @domenicrosati, we don't have a way to create resources with anything other than our user's ownerId, so unfortunately we can't fully demonstrate filtered lists that look any different than a user's full list of resources. Until we can do that, the acceptance criteria is:

1. Make sure both components work without an ownerId
2. Make sure they work the same with a valid ownerId
3. Make sure they fail on an invalid ownerId

I've added an ownerId knob to both resource -> list and resource -> list (unstyled) stories to test this.

## Checklist


src/components.d.ts
```diff
@@ -609,355 +616,355 @@ declare global {
...
   interface HTMLManifoldToggleElement extends Components.ManifoldToggle, HTMLStencilElement {}
-  const HTMLManifoldToggleElement: {
+  var HTMLManifoldToggleElement: {
     prototype: HTMLManifoldToggleElement;
     new (): HTMLManifoldToggleElement;
   };
 
   interface HTMLManifoldTooltipElement extends Components.ManifoldTooltip, HTMLStencilElement {}
-  const HTMLManifoldTooltipElement: {
+  var HTMLManifoldTooltipElement: {
```
domenicrosati: @markjamieson you will have to update stencil with `npm i` so these don't change

src/utils/fetchAllPages.ts
```diff
@@ -15,6 +15,7 @@ interface NextPage {
 interface Args<Edge, T> {
   query: string;
   nextPage: NextPage;
+  variables?: object;
```
domenicrosati: +1

src/components.d.ts
```diff
@@ -609,355 +616,355 @@ declare global {
...
   interface HTMLManifoldButtonElement extends Components.ManifoldButton, HTMLStencilElement {}
-  const HTMLManifoldButtonElement: {
+  var HTMLManifoldButtonElement: {
     prototype: HTMLManifoldButtonElement;
     new (): HTMLManifoldButtonElement;
   };
 
   interface HTMLManifoldButtonLinkElement extends Components.ManifoldButtonLink, HTMLStencilElement {}
-  const HTMLManifoldButtonLinkElement: {
+  var HTMLManifoldButtonLinkElement: {
```
dangodev: @markjamieson you’ll need to run `npm i` to install the latest version of Stencil

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -34,6 +34,26 @@ const query = gql`
...
+            id
+          }
+          status {
+            label
+          }
+        }
+      }
+    }
+  }
+`;
```
dangodev: Is it possible to use `fetchAllPages` here instead? That would help us paginate properly.

src/utils/fetchAllPages.ts
dangodev: 👎 this should be `{ [key: string]: string }` (I think)

dangodev: Sorry—that thumbs down seemed harsh. I’m not _that_ upset by it. I just wanted to make it clear I was asking for a change.

domenicrosati: hahaha no worries i think its good to actually use the types we need

markjamieson: Ok, was just borrowing the definition from the graphqlFetch interface. I should update both I think.

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
```diff
@@ -70,7 +92,8 @@ export class ManifoldDataResourceList {
     }
 
     const { data, errors } = await this.graphqlFetch({
-      query,
+      query: this.ownerId ? queryWithOwner : query,
```
dangodev: If possible, this query should be in a `.graphql` file. Then run `npm run generate:gql`, and use `graphqlFetch<GeneratedType>`.

This is a brand-new thing; no docs for it (sorry; working on it). There are a couple examples of this, but LMK if you need help with it!

markjamieson: No help needed, should be pretty straightforward.
Side note, is there any way to have stencil or storybook watch for changes on those files? I've been restarting storybook when editing those files because the changes don't get picked up.

markjamieson: Actually, I do have a question, or at least an opinion to solicit:

These queries are very similar to the ones in `manifold-resource-list`. Should I just re-use those queries, even though we also request plan data in those? Or is it better to stick to the smallest possible query we can?

dangodev: IMO re-using queries is totally fine, so long as you use the types generated. That makes sense to me. Less to maintain overall, and TypeScript will notify us if anything breaks.

dangodev: > is there any way to have stencil or storybook watch for changes on those files?

Unfortunately not at the moment. As early adopters of Stencil.js, we had to customize Storybook a little bit. But I’m listening for updates on the Stencil.js support issue for Storybook, and it’s gotten a **lot** of activity over the past couple months. So I think better support will come soon.

src/utils/fetchAllPages.ts
dangodev: Oh sorry! yes that’d be good.

dangodev: But yeah, the `object` type in TypeScript might as well be `any` that’s why I’m against it.

markjamieson: Ok, I had to settle on `{ [key: string]: string | number | undefined }`

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
markjamieson: Done

src/components/manifold-data-resource-list/manifold-data-resource-list.tsx
markjamieson: Done

PR #665, opened 2019-10-28
Require other jobs to pass before VRT can run

sslotsky:
## Reason for change

We are running Happo more often than we need to, which can be costly. This change will require other jobs to pass before VRT can run, which will prevent some of our superfluous Happo runs.

## Checklist

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -27,6 +27,7 @@ jobs:
       - run: npm run lint
   vrt:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dangodev:
Note we’re testing out the Semantic PR titles. So prefix this with `ci: …`

Also the Happo check doesn’t seem to be working. It generates a report correctly but has trouble talking back to GitHub. Investigating that currently.

sslotsky:
Closing in favor of #667

PR #668, opened 2019-10-29
feat: Add component name

dangodev:
## Reason for change

We want to track which component sends each GraphQL request, and what version platforms are currently using. This adds both as `x-manifold-component` and `x-manifold-version` headers, respectively.

Screen Shot 2019-10-29 at 13 14 56

Screen Shot 2019-10-29 at 13 15 02

This only adds it to `<manifold-marketplace>` and `<manifold-plan-selector>` as a start. If we like this approach, we can require this and enforce it on all components.

_Note: for easier review, the UI version was split into #670, to follow this PR_

## Testing

- [ ] `<manifold-marketplace>` emits correct headers in Storybook
- [ ] `<manifold-plan-selector>` emits correct headers in Storybook

## Checklist


package.json
```diff
@@ -23,12 +23,13 @@
...
   "scripts": {
-    "build": "stencil build",
+    "build": "NPM_PACKAGE_VERSION=$(git describe --abbrev=0 --tags | sed 's/^v//') stencil build",
     "build:docs": "npm run prepare:docs && cd docs && npm i && npm run build",
     "build:storybook": "npm run build && build-storybook && npm run copy-changelog",
+    "build:watch": "NPM_PACKAGE_VERSION=$(git describe --abbrev=0 --tags | sed 's/^v//') stencil build --watch",
     "bundlesize": "npm run build && npx bundlesize",
     "copy-changelog": "node scripts/copy-changelog",
-    "dev": "concurrently 'stencil build --watch' 'rm -rf dist && wait-on dist/loader/index.cjs.js && manifold run -p manifold-ui-local -- npm run storybook'",
+    "dev": "npm run build:watch & rm -rf dist && wait-on dist/loader/index.cjs.js && manifold run -p manifold-ui-local -- npm run storybook",
```
dangodev: Recently I learned `concurrently` can be replaced with `&`. This works the same, but saves on a package install.

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -110,6 +110,7 @@ export class ManifoldPlanSelector {
       variables: {
         productLabel,
       },
+      component: 'manifold-plan-selector',
```
dangodev: This is a bit manual, but it’s simple.

I wasn’t able to locate this string anywhere (`this.constructor.name` was `class_1` when built), and it’s not easy to deduplicate `@Component({ tag` to use a runtime variable.

package.json
```diff
@@ -23,12 +23,13 @@
   "types": "dist/types/components.d.ts",
   "collection": "dist/collection/collection-manifest.json",
   "scripts": {
-    "build": "stencil build",
+    "build": "NPM_PACKAGE_VERSION=$(git describe --abbrev=0 --tags | sed 's/^v//') stencil build",
```
dangodev: I set this up so that the root `package.json` doesn’t have a `version`. I did this to make it clear that the version is determined from Git, and to reduce errors in people knowing how to release versions. But that also made it a bit tricky to grab the version here.

I know that this isn’t the most straightforward, but this approach does give us one thing keeping the version in `package.json` wouldn’t: assurance it’s correct.

package.json
```diff
@@ -52,7 +54,8 @@
     "test:spec": "stencil test --spec",
     "test:coverage": "stencil test --spec --e2e --coverage",
     "test:debug": "node --inspect-brk node_modules/.bin/stencil test --spec --e2e --runInBand",
-    "test:watch": "stencil test --spec --e2e --watchAll"
+    "test:watch": "stencil test --spec --e2e --watchAll",
+    "update-version": "npm --no-git-tag-version version from-git > /dev/null || true"
```
dangodev: Because we were deriving the package version from Git, which isn’t always accessible to all CI tools, I needed a more secure mechanism. I changed how the version is stored in our `package.json`, but it should automatically update and be more reliable.

I anticipate it’d behave similar to `components.d.ts`—sometimes this file will update in a PR unexpectedly. The same will be true for the version in `package.json`. But somehow, somewhere, the version will get in sync with the Git tags. It’ll likely just happen in someone’s PR.

src/utils/graphqlFetch.ts
```diff
@@ -16,12 +15,12 @@ type GraphqlArgs =
   | {
       mutation: string;
       variables?: { [key: string]: string | number | undefined };
-      emitter?: EventEmitter;
```
dangodev: @sslotsky do we still need the `emitter` option (and if so, should it emit the event in place of `element` or alongside it)?

sslotsky: Nope, `emitter` can go away

src/utils/graphqlFetch.ts
```diff
@@ -16,12 +15,12 @@ type GraphqlArgs =
...
+      element?: HTMLElement;
     }
   | {
       query: string;
       variables?: { [key: string]: string | number | undefined };
-      emitter?: EventEmitter;
+      element?: HTMLElement;
```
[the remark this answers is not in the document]
dangodev: Oh yes agreed. I was going to change everything over in a followup PR, and remove that from being optional. If you’d like I can just do that here, though.

PR #671, opened 2019-10-30
Expose storybook as a container for dev

luizbranco:
How to run:

```
docker build -t manifoldco/ui .
docker run -p 6006:6006 -d manifoldco/ui
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM node:13
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+
+RUN npm install
+
+# Bundle app source
+COPY . .
+
+EXPOSE 6006
+
+CMD [ "npm", "run", "dev" ]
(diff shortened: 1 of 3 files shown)


sslotsky:
What's the motivation behind running this in a container?

luizbranco:
Running in a container would help local dev test without having npm stack installed. We can also use the image as part of the backend docker-compose to bring everything up together.

dangodev:
I’m OK with adding this for the backend folks! But if you’re not actually making UI changes and want hotloading locally, I’d recommend using `npm run build:storybook` instead.

For Now, we’re just using a static site. The dev server is too much memory overhead if you don’t actually need the local server.

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+COPY package*.json ./
+
+RUN npm install
+
+# Bundle app source
+COPY . .
+
+EXPOSE 6006
+
+CMD [ "npm", "run", "dev" ]
```
dangodev: This should probably be `"npm", "run", "build:storybook"`. Then just serve the `storybook-static/` directory it creates in the root with whatever simple static file server you’d like.

You can even use `npm run build:storybook && npx serve -s storybook-static` to spin up a Node server to serve those static files.

gclapperton: 👍 I've updated the docker image to build the static site as suggested.

PR #676, opened 2019-10-30
feat: Add component names to GraphQL

dangodev:
## Reason for change

Followup to #668 that discussed approach. This is just a chore to update all the names to broadcast their names to GraphQL, and enforce that via TypeScript.

## Testing

All components should now send their name to GraphQL via the `x-manifold-component` header.

## Checklist

### DIFF
--- a/src/utils/fetchAllPages.ts
+++ b/src/utils/fetchAllPages.ts
@@ -18,7 +18,7 @@ interface Args<Edge, T> {
   variables?: { [key: string]: string | number | undefined };
   getConnection: (q: T) => Connection<Edge> | null | undefined;
   graphqlFetch?: GraphqlFetch;
-  element?: HTMLElement;
+  element: HTMLElement;
 }
 
 export class MissingPageInfo extends Error {
--- a/src/utils/graphqlFetch.spec.ts
+++ b/src/utils/graphqlFetch.spec.ts
@@ -26,7 +26,7 @@ describe('graphqlFetch', () => {
         status: 200,
         body: { data: {} },
       });
-      await fetcher({ query: '' });
+      await fetcher({ query: '', element: document.createElement('custom-element') });
       expect(fetchMock.called('https://api.manifold.co/graphql')).toBe(true);
     });
 
@@ -47,7 +47,10 @@ describe('graphqlFetch', () => {
         body,
       });
 
-      const result = await fetcher({ query: '' });
+      const result = await fetcher({
+        query: '',
+        element: document.createElement('custom-element'),
+      });
 
       expect(fetchMock.called(graphqlEndpoint)).toBe(true);
       expect(result).toEqual(body);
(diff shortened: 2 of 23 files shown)


src/utils/errorReport.ts
```diff
@@ -1,4 +1,4 @@
-export function report<Detail>(detail: Detail, element?: HTMLElement) {
+export function report<Detail>(detail: Detail, element: HTMLElement) {
```
dangodev: `element` is now mandatory in the error reporting component. If we need to adjust this later, we can.

src/utils/fetchAllPages.ts
```diff
@@ -18,7 +18,7 @@ interface Args<Edge, T> {
   variables?: { [key: string]: string | number | undefined };
   getConnection: (q: T) => Connection<Edge> | null | undefined;
   graphqlFetch?: GraphqlFetch;
-  element?: HTMLElement;
+  element: HTMLElement;
```
dangodev: `element` now mandatory in `fetchAllPages`

src/utils/graphqlFetch.ts
```diff
@@ -11,18 +11,18 @@ interface CreateGraphqlFetch {
...
       variables?: { [key: string]: string | number | undefined };
-      element?: HTMLElement;
     }
   | {
       query: string;
       variables?: { [key: string]: string | number | undefined };
-      element?: HTMLElement;
     }; // require query or mutation, but not both
 
+type GraphqlArgs = GraphqlRequest & { element: HTMLElement };
```
dangodev: This not only makes more logical sense out here, but also it makes the `request` event detail handling below a lot simpler.

src/utils/graphqlFetch.spec.ts
```diff
@@ -119,9 +123,9 @@ describe('graphqlFetch', () => {
 
       fetchMock.mock(graphqlEndpoint, { status: 422, body });
 
-      await fetcher({ query: '' });
+      await fetcher({ query: '', element: document.createElement('custom-element') });
       // graphql format
-      expect(errorReporter).toHaveBeenCalledWith(body.errors, undefined);
+      expect(errorReporter).toHaveBeenCalledWith(body.errors, {});
```
sslotsky: This is fascinating. Does `document.createElement('custom-element')` just return an empty object?

Wondering what happens if we just use a jest mock instead. Either way though, it seems like it would be better to use a variable to hold whatever you're passing in as `element`, and use that variable as the second argument to this `toHaveBeenCalledWith` function.

src/utils/errorReport.ts
dangodev: Ah just realized we can’t because of `auth.ts`. There’s no element associated with that. Reverted that, but I did also add it to `<manifold-auth-token>`.

If we want to enforce that, I’d like to do so in another PR (I know we also talked about combining `<manifold-connection>` and `<manifold-auth-token>`, and decided not to for now, but forgot why).

PR #682, opened 2019-10-31
Fix version bump script

dangodev:
## Reason for change
Originally, we **didn’t keep the package version hard-coded anywhere** for simplicity. But we needed to, in order to send it as a header. As a result, some of the release scripts were affected.

This changes things a bit to keep the version headers while fixing the build script. The end result is that `package.json` now has the correct version, but we don’t have to have people manually update that because that causes problems.

I’d really love to simplify a lot of the pre-release logic, but in order to do that we’d need to remove the `docs/` from this repo and use GitBook (reason is that Gatsby needs to use a local `@manifold/ui` version that’s not a parent directory). Once that happens, the release code can get a lot simpler.

## Checklist


.circleci/config.yml
```diff
@@ -5,7 +5,6 @@ version: 2.1
 defaults: &defaults
   docker:
     - image: circleci/node:12
-  resource_class: xlarge
```
dangodev: @domenicrosati is this right?

domenicrosati: yep!

docs/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@manifoldco/icons": "0.0.3",
-    "@manifoldco/ui": "file:../pkg",
+    "@manifoldco/ui": "^0.6.1",
```
dangodev: I’d like to version-lock the docs for now, rather than use the local version.

I’d really like to move them to GitBook, but we’re not there yet.

Anyways, this prepares us for simplifying the npm release process by not relying on local files.

PR #683, opened 2019-10-31
feat: add no credentials state to <manifold-credentials>

markjamieson:
## Reason for change

Adds a helpful message + SSO button in place of a credentials box for products which do not provide credentials.

## Testing

I added a `no-credentials` checkbox to the resource details Story to force this state. Testing a credential-less product otherwise looked like it would be fairly complex.

## Checklist


src/components/manifold-no-credentials/manifold-no-credentials.tsx
```diff
@@ -0,0 +1,68 @@
...
+import logger from '../../utils/logger';
+import loadMark from '../../utils/loadMark';
+import resourceNoCredentialsQuery from './resourceNoCredentials.graphql';
+import connection from '../../state/connection';
+import { GraphqlFetch } from '../../utils/graphqlFetch';
+import { ResourceNoCredentialsQuery } from '../../types/graphql';
+
+const PLATFORM_DISPLAY_NAMES: { [key: string]: string } = {
+  'manifold.co': 'Manifold',
+  'digitalocean.com': 'Digital Ocean',
```
dangodev: DigitalOcean (no space)

src/components/manifold-no-credentials/manifold-no-credentials.tsx
```diff
@@ -0,0 +1,68 @@
...
+  'render.com': 'Render',
```
dangodev: IMO these are fine at this stage, but what about moving forward? Since the code is public, what about upcoming platforms?

src/components/manifold-no-credentials/manifold-no-credentials.tsx
```diff
@@ -0,0 +1,68 @@
...
+      query: resourceNoCredentialsQuery,
+      variables: { resourceLabel: this.resourceLabel },
+      element: this.el,
+    });
+    const platformName = data && PLATFORM_DISPLAY_NAMES[data.resource.owner.platform.domain];
+    const productName = data && data.resource.plan.product.displayName;
+    this.message = `Access to ${productName ||
+      'this resource'} is authorized through your ${platformName} account without credentials. SSO into the ${productName ||
+      'product'} dashboard to configure your service.`;
+  }
```
dangodev: Actually, alternate suggestion to the above question about platforms. I think this should just be a named slot for platforms to control (e.g. `<div slot="no-credentials">Access to …</div>`). Platforms control this messaging, and we don’t have to worry about personalization.

markjamieson: Will they be able to put a dynamic message in there that includes the name of the product?

dangodev: Yeah slots are just HTML children. So they control their own messaging, can put links in, etc.

src/components/manifold-no-credentials/manifold-no-credentials.tsx
markjamieson: No idea. Would be good to have `displayName` alongside `domain` in the API schema, but I'm not sure we want to track that on the backend.

src/components/manifold-no-credentials/manifold-no-credentials.tsx
markjamieson: Yea, but how will their implementation know the product name? That's what I'm not getting. They have the resourceLabel, but without making their own API call, how can they add a message that refers to the product name?

src/components/manifold-no-credentials/manifold-no-credentials.tsx
markjamieson: I don't think this is a big deal now that the slot is being included for platforms to specify their own messaging. The default messaging falls back to a message that just doesn't include the platform name if it's not found in this map.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@manifoldco/ui",
   "description": "Manifold UI",
-  "version": "0.6.1",
+  "version": "0.5.12",
```
dangodev: just a heads up: a `git fetch` should fix this.

This is a new thing with tracking the NPM version. It doesn’t really matter too much what we commit, but this version # will log to our DD, just FYI. It’ll pull from the latest Git tag you have locally.

dangodev: Anyways, whenever we publish to npm, this will be correct. But expect there to be a bit of noise here in PRs if someone’s local tags are out-of-date or their branch isn’t in sync with `master`

src/components/manifold-resource-credentials/manifold-resource-credentials.tsx
```diff
@@ -20,13 +21,19 @@ export class ManifoldResourceCredentials {
...
           <slot name="show-button" />
         </manifold-forward-slot>
         <manifold-forward-slot slot="hide-button">
           <slot name="hide-button" />
         </manifold-forward-slot>
+        <manifold-forward-slot slot="sso-button">
+          <slot name="sso-button" />
+        </manifold-forward-slot>
+        <manifold-forward-slot slot="no-credentials">
+          <slot name="no-credentials" />
```
dangodev: IMO you don’t have to write a spec test for this, but a Happo test would be perfect.

PR #685, opened 2019-11-01
fix: added read-only state for regions

davidleger95:
## Reason for change

Add the option for regions to be read-only in `<manifold-plan-details>` so that it can be used in `<manifold-resource-plan>` in a read-only state.

## Testing

- check that the region display in resource details as plain text
- check that the select dropdown is displayed in the plan selector

## Checklist

### DIFF
--- a/src/components/manifold-plan-details/manifold-plan-details.tsx
+++ b/src/components/manifold-plan-details/manifold-plan-details.tsx
@@ -34,7 +34,6 @@ export class ManifoldPlanDetails {
   @Prop() region?: Region;
   @Prop() regions?: string[];
   @Prop() resourceRegion?: string;
-  @Prop() readOnly?: boolean;
   @State() features: Gateway.FeatureMap = {};
   @State() regionId?: string;
   @Event({ eventName: 'manifold-planSelector-change', bubbles: true }) planUpdate: EventEmitter;
@@ -251,7 +250,7 @@ export class ManifoldPlanDetails {
                   />
                 ))}
             </dl>
-            {this.regionOptions && (
+            {(this.regionOptions || this.region) && (
               <div class="region">
                 <label
                   class="region-label"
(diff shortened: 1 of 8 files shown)


src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -209,6 +211,7 @@ export class ManifoldPlanDetails {
   @logger()
   render() {
     if (this.plan && this.product) {
+      console.log(this.region);
```
dangodev: Need to remove this

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -31,8 +31,10 @@ export class ManifoldPlanDetails {
   @Prop() scrollLocked?: boolean = false;
   @Prop() plan?: Plan;
   @Prop() product?: Product;
+  @Prop() region?: Region;
   @Prop() regions?: string[];
   @Prop() resourceRegion?: string;
+  @Prop() readOnly?: boolean;
```
dangodev: I like this name. Clear purpose to me

davidleger95:
@dangodev are we still using the old docs or gitbook or both?

dangodev:
> are we still using the old docs or gitbook or both?

I’d like to switch to GitBook to gain versioning and alleviate the engineering burden, but that’ll take approval from product. For now, the docs in this repo (Gatsby) are still our official ones. If you have time to update GitBook as well that’d be nice, but don’t sweat it—we may have changes there before switching over completely.

dangodev:
@davidleger95 if it’s not too much work, could you add this same check to `manifold-plan-selector` as well? Basically it needs the `isReadonly` prop if `resource-label` is specified.

Screen Shot 2019-11-01 at 10 46 26

That’d help so much with what I’m working on now!

davidleger95:
on second pass, turns out i didn't need a `reasdOnly` prop

PR #686, opened 2019-11-01
update create with owner query to use string owner

domenicrosati:
## Reason for change

Because ownerId only accepts manifold Ids (the ID type in gql) and we want folks to be able to create resources on behalf of their own user ids as the docs suggest :) .

## Testing

Creating a resource should still work good :) !!!

## Checklist


src/components/manifold-data-provision-button/create-with-owner.graphql
```diff
@@ -1,12 +1,12 @@
...
+mutation CreateResourceWithOwner($planId: ID!, $productId: ID!, $regionId: ID!, $resourceLabel: String!, $ownerId: ProfileIdentity!) {
   createResource(
     input: {
       displayName: $resourceLabel
       label: $resourceLabel
       planId: $planId
       productId: $productId
       regionId: $regionId
-      ownerId: $ownerId
+      owner: $ownerId
```
sslotsky: The schema seems to indicate that this should still be named `ownerId`

sslotsky: https://github.com/manifoldco/ui/pull/686/files#diff-c643ec2fe42e70cf6e64b253f9765722R1522

domenicrosati: I belive that is for the variable names themselves. The variable is still called ownerId.

The actual field is this:
https://github.com/manifoldco/ui/pull/686/files#diff-c643ec2fe42e70cf6e64b253f9765722R102

sslotsky: I see. Would it be better if the field name matched the variable name.

domenicrosati: ya - revised here

src/components/manifold-data-provision-button/create-with-owner.graphql
```diff
@@ -1,12 +1,12 @@
...
+mutation CreateResourceWithOwner($planId: ID!, $productId: ID!, $regionId: ID!, $resourceLabel: String!, $owner: ProfileIdentity!) {
   createResource(
     input: {
       displayName: $resourceLabel
       label: $resourceLabel
       planId: $planId
       productId: $productId
       regionId: $regionId
-      ownerId: $ownerId
+      owner: $owner
```
sslotsky: So they match now, which is better, but should we have gone the other direction? In this input alone we see `planId`, `productId`, and `regionId` where we're using IDs. Is it better for consistency if we use `ownerId` then?

domenicrosati: unfortunetly we can't because `ownerId` is a valid input in gql that only takes manifold ids :(

sslotsky: Ahhh interesting. So if that's the case, then why don't they both show up in `graphql.ts` underneath the `CreateResourceWithOwnerMutationVariables` type?

domenicrosati: Hmm maybe its because its deprecated? Maybe the deprecation directive on that query makes it not show up. ahhaha i dont quite know

sslotsky: Oh, that sounds reasonable! Ok this is unfortunate but I'll take your word for it :sweat_smile:

PR #696, opened 2019-11-04
refactor: Move plan queries to graphql files

davidleger95:
## Reason for change

Moving queries to graphql files based on this: https://app.zenhub.com/workspaces/engineering-56990653b8516b1b4867e8b2/issues/manifoldco/engineering/9608

## Testing

- Check that `manifold-plan-selector` and `manifold-plan` still fetch data

## Checklist


src/components/manifold-plan-selector/plan-list.graphql
```diff
@@ -0,0 +1,81 @@
+fragment planDetails on Plan {
```
davidleger95: I'd like to move this fragment to a shared file. not sure if that's possible

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -1,47 +1,13 @@
...
-      plan {
-        id
-        product {
-          label
-        }
-      }
-    }
-  }
-`;
+import plansQuery from './plan-list.graphql';
```
dangodev: We don’t really get the full benefit from this until we do `graphqlFetch<PlanDetailsQuery>({` and use the generated type. But in this scenario, that’s quite a bit of work, and doesn’t have to happen in this PR.

We do get a tiny benefit from moving it to `.graphql`—it’ll be validated against our schema. But if we’re using the full types (`Product, Resource, PlanEdge`), there’s still a danger we reference something in TypeScript that’s not really there.

Can we make sure that‘s at least an issue or two to use the generated types for all the plan components?

davidleger95: ah, I see what you mean. yeah, that's a bit more work. I generated the types at least. I'll push those up.

PR #699, opened 2019-11-04
fix: always display contents of manifold-resource-container

davidleger95:
## Reason for change

Resource container would hide it's contents until the `fetchResource` promise was resolved. Now the fetch no longer blocks the display of content and shows loading states.

## Testing

- All the resource stories should never show a blank screen while the resource hasn't been fetched.
- All the resource stories should never show a blank screen while unauthenticated. (This can be tested by setting `env: Production` and `authType: Manual`)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@manifoldco/ui",
   "description": "Manifold UI",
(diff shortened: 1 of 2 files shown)


davidleger95:
All seems to be working well

PR #708, opened 2019-11-06
Manifold Analytics Client

dangodev:
## Reason for change

Screen Shot 2019-11-06 at 14 26 26

Step 1 to analytics Zen is the client to handle reporting. This will be a part of the UI repo short-term, but in its own space for easy extraction.

This is **errors only** for now; events to come in separate tickets. This is the current error payload:

```json
{
  "type": "error",
  "name": "ui_error",
  "properties": {
    "code": "ErrorCode",
    "componentName": "MANIFOLD-MARKETPLACE",
    "message": "I am an error message",
    "uiVersion": "1.2.3"
  },
  "source": "ui"
}
```

`name` is currently just “Error” everywhere now (for many errors in JavaScript, all we get is a string and nothing more). Maybe it should match `message`?

## Testing

- [x] See that `analytics.manifold.co` is called on a component error (you may have to edit code locally to see this)

## Checklist


packages/analytics/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  evt: AnalyticsEvent,
+  userOptions: AnalyticsOptions = { env: 'prod' }
+) {
+  const options: AnalyticsOptions = {
+    env: 'prod',
+    ...(userOptions || {}),
+  };
+  const url = endpoint[options.env || 'prod'];
+  return fetch(url, { method: 'POST', body: JSON.stringify(evt) });
+}
```
dangodev: This is in `packages/` as that’s a common folder structure for repos that contain multiple packages. We could even deploy to npm from this repo as an iterative next step.

packages/analytics/types.ts
```diff
@@ -0,0 +1,67 @@
...
+        time: number;
+      };
+    }
+  | {
+      name: 'token_received';
+      description: string;
+      properties: {
+        time: number;
+      };
+    };
```
dangodev: @domenicrosati @tainacleal this is the proposed JSON structure to send to the endpoint. Let me know what you think!

packages/analytics/types.ts
```diff
@@ -0,0 +1,67 @@
...
+ *  Error analytics event
+ */
+export interface ErrorEvent extends RequiredProperties {
+  type: 'error';
+  name: string;
+  properties: {
+    code: string;
+    componentName: string;
+    message: string;
+    uiVersion: string;
```
dangodev: So I realized that in some places we have `npmVersion`, but I’d prefer `uiVersion`. Reason is: we don’t have to serve UI via CDN, so I don’t feel like `npmVersion` is all-encompassing. But overall I’m 🍌

src/utils/errorReport.ts
```diff
@@ -1,12 +1,30 @@
-/* eslint-disable somethnig */
+import analytics from '../../packages/analytics';
 
-export function report<Detail>(detail: Detail, element?: HTMLElement) {
+interface ErrorDetail {
+  code?: string;
+  componentName?: string;
+  message?: string;
+}
```
dangodev: @sslotsky when I looked at all the different ways the error logger was used, there wasn’t a clear structure. Sometimes it got a string, other times an object… This is an attempt to standardize it. Thoughts?

src/utils/errorReport.ts
```diff
@@ -1,12 +1,30 @@
...
+
+export function report(detail: ErrorDetail, element?: HTMLElement) {
   const extendedDetail = {
     ...detail,
     ...(element ? { componentName: element.tagName } : {}),
+    ...(detail.componentName ? { componentName: detail.componentName } : {}),
```
dangodev: I realized in some situations we provided `componentName` but not `element`. This accounts for that.

src/utils/graphqlFetch.ts
```diff
@@ -80,19 +80,33 @@ export function createGraphqlFetch({
         'x-manifold-ui-version': '<@NPM_PACKAGE_VERSION@>',
       },
       body: JSON.stringify(request),
-    }).catch((e: Response) => {
+    }).catch(e => {
       // handle unexpected errors
-      report(e, element);
+      report(
+        {
+          code: e.code,
```
dangodev: This will be `undefined`. What error code should we send for a network error? cc @TerrySmithDC (not sure where y’all are on error codes)?

dangodev: (for context, we’re sending errors to our analytics endpoint which is separate from the error codes project, but I figured I could at least re-use any error codes y’all have already figured out)

src/utils/graphqlFetch.ts
```diff
@@ -115,7 +129,15 @@ export function createGraphqlFetch({
...
-      report(body.errors, element);
+      body.errors.forEach(e => {
+        report(
+          {
+            code: e.extensions && e.extensions.type,
+            message: e.message,
+          },
+          element
+        );
+      });
```
dangodev: Another instance of `report()` sometimes logging an object, other times logging an array (here)

src/utils/errorReport.ts
```diff
@@ -1,12 +1,30 @@
...
+  analytics({
+    name: 'Error',
+    type: 'error',
+    properties: {
+      code: detail.code || '',
+      message: detail.message || '',
+      componentName: detail.componentName || (element && element.tagName) || '',
+      uiVersion: extendedDetail.npmVersion,
+    },
+  });
```
dangodev: @sslotsky I’d like to think of a way to pull from `<manifold-connection>`, but that may not possible without reworking the `errorReport`

packages/analytics/types.ts
domenicrosati: i am very strongly in the uiVersion camp

src/utils/errorReport.ts
```diff
@@ -1,12 +1,30 @@
...
+export function report(detail: ErrorDetail, element?: HTMLElement) {
   const extendedDetail = {
     ...detail,
     ...(element ? { componentName: element.tagName } : {}),
+    ...(detail.componentName ? { componentName: detail.componentName } : {}),
     npmVersion: '<@NPM_PACKAGE_VERSION@>',
   };
 
+  analytics({
+    name: 'Error',
```
domenicrosati: @tainacleal we are going to have to namespace all of the names that come into analytics ingest based on their source.

Because multiple sources can be reporting a metric with name 'Error'

I don't know how that change how you think @dangodev

src/packages/analytics/types.ts
```diff
@@ -0,0 +1,67 @@
...
+  | {
+      name: 'token_received';
+      description: string;
+      properties: {
+        time: number;
+      };
+    };
+
+export type EventEvent = {
+  type: 'event';
```
domenicrosati: @dangodev can you please also add a `source` key?
we are going to need to require it in the backend hence not a property.
Essentially source will be able to tell us where an event came from.

src/packages/analytics/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+interface AnalyticsOptions {
+  env?: 'stage' | 'prod';
+}
+
+/**
+ * Report an error or analytics event to Manifold
+ * @param {Object} eventData Event data to send to Manifold
+ * @param {string} eventData.type 'event' or 'error'
+ * @param {string} eventData.name name_of_event (lowercase with underscores)
```
domenicrosati: See comment below about adding source.
It would be something like
`source: "ui"` in this case

src/utils/errorReport.ts
dangodev: Would it help if the name was the error message (e.g. `Internal server error`)?

Or should it be something like `UI Error`?

tainacleal: I was thinking `name` would be something we would map internally based on the `type`, because if that's the metric name we should be setting on datadog for example, that's something the analytics service should have control. So I don't even think `name` is needed from the front end.
And about namespacing, @domenicrosati @dangodev how do you feel about the endpoint defining the source? So ui would send events to `v1/ui/events`, or something like that?

domenicrosati: The reason why `name` is important which perhaps isn't clear here with Errors is we need two piece of data about an event - what core type of event, Errors, Analytics, Operations, ect.
and what the specific metrics is ie. UI Error, Round Trip time, Loading, ect.

domenicrosati: i am still thinking on the different endpoints right now I am on the fence

domenicrosati: @dangodev  it would be helpful if the name was `ui_errror`

dangodev: I’m open to namespacing or specifying endpoints. Happy to change it in this PR

src/packages/analytics/types.ts
dangodev: Added

src/packages/analytics/types.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface ErrorEvent extends RequiredProperties {
+  type: 'error';
+  name: 'ui_error';
+  properties: {
+    code: string;
+    componentName: string;
+    message: string;
+    uiVersion: string;
+  };
+  source?: 'ui';
```
dangodev: Added `source` to events and errors @domenicrosati

src/packages/analytics/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  const options: AnalyticsOptions = {
+    env: 'prod',
+    ...(userOptions || {}),
+  };
+  const url = endpoint[options.env || 'prod'];
+  return fetch(url, {
+    method: 'POST',
+    body: JSON.stringify({
+      ...evt,
+      source: 'ui', // add source
```
dangodev: @domenicrosati from other comment: we send `source: 'ui'` for all requests, no matter what

src/packages/analytics/types.ts
```diff
@@ -0,0 +1,67 @@
...
+    componentName: string;
+    uiVersion: string;
+  };
+}
+
+/**
+ *  Based on `name`, what data should be sent?
+ */
+export type EventTypes =
+  | {
```
TerrySmithDC: Does this pipe need to be here?

dangodev: Yes it’s a discriminated union in TypeScript, which is my favorite. Based on the `name` you give it, it’ll infer the rest. So with that, you can do special things like ensure `properties.rttGraphql` exists on `name: 'load'` but not `name: 'render_with_data'`. Basically it just prevents mistakes a lot better than marking everything optional

dangodev: Oh but yeah TypeScript inserts a leading `|` if it wraps to a new line for some reason. Dunno why.

TerrySmithDC: Yeah discriminated union's are super awesome! I just didn't think they started with a pipe, I thought it was 
`type Foo = Bar | Biz | Baz`

src/utils/errorReport.ts
sslotsky: Yeah, when I wrote it I think it made sense to be unopinionated about that so I just made it generic. Now that it's been around a while and it's more clear how it's actually used in the wild, I think it's fine to standardize.

src/utils/errorReport.ts
sslotsky: I'm not sure what you mean here. What would we pull from manifold-connection, and why?

src/utils/errorReport.ts
dangodev: I was on the fence on whether or not to `console.error` within the analytics library. I feel like it’s probably better for that to be silent, and leave that decision to the library integrating it.

So I like having the errorReporter be a wrapper for that in UI, where that’s the level we cater to platform things, and the analytics method only cares about reporting to Manifold and nothing else.

src/utils/graphqlFetch.ts
sslotsky: There shouldn't be `e.code` here. This object should be a `Response`, which IIRC would have a `status` or a `statusCode` (can't remember which). I think the `code` you're thinking of here is actually part of the response body. What we're catching in this block is a network error.

src/utils/errorReport.ts
dangodev: The environment—basically sending to `stage.analytics.manifold.co` instead

I’m not sure how to have that be a UI configuration without creating an analytics component in `<manifold-connection>` or something

src/utils/restFetch.ts
```diff
@@ -51,9 +51,9 @@ export function createRestFetch({
     const response = await fetch(`${endpoints()[args.service]}${args.endpoint}`, {
       ...options,
       body: JSON.stringify(args.body),
-    }).catch((e: Response) => {
+    }).catch(e => {
```
sslotsky: Again, I think this change is wrong. This should definitely be a `Response` which would not have a `message` or a `code`.

src/utils/errorReport.ts
sslotsky: Environment is part of the connection state object, which could be passed along to `createGraphqlFetch`. Although if we did that I'd also be tempted to update the signature so that the `endpoint()` function defaults to resolving the URL by the specified env, since it seems redundant to supply both an `env` and an `endpoint()`

src/utils/graphqlFetch.ts
dangodev: Oh gotcha. Yeah I don’t want to mess with JSON / might not be JSON here. I’ll just omit the `code`

src/components/manifold-copy-credentials/manifold-copy-credentials.spec.ts
```diff
@@ -57,6 +57,8 @@ describe('<manifold-copy-credentials>', () => {
     });
   });
 
+  afterEach(() => fetchMock.restore());
```
sslotsky: If you're a fan of Point Free Style, you could use that here

```ts
afterEach(fetchMock.restore);
```

dangodev: Oh nice! I’ll get that in a followup

PR #709, opened 2019-11-07
Add local & staging env to analytics

dangodev:
## Reason for change

Infers local & staging envs from `<manifold-connection>`

## Testing

- [ ] Currently, analytics only reports on error. Trigger an error (may have to edit local code), and ensure that it reports to different endpoints depending on whether `env="prod"`, `env="stage"`, or `env="local"`

## Checklist

### DIFF
--- a/src/packages/analytics/index.spec.ts
+++ b/src/packages/analytics/index.spec.ts
@@ -2,8 +2,9 @@ import fetchMock from 'fetch-mock';
 import report from './index';
 import { ErrorEvent } from './types';
 
-const stage = 'https://analytics.stage.manifold.co';
-const prod = 'https://analytics.manifold.co';
+const local = 'begin:https://analytics.arigato.tools';
+const stage = 'begin:https://analytics.stage.manifold.co';
+const prod = 'begin:https://analytics.manifold.co';
 const error: ErrorEvent = {
   type: 'error',
   name: 'ui_error',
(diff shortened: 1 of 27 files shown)


src/packages/analytics/index.spec.ts
```diff
@@ -2,6 +2,7 @@ import fetchMock from 'fetch-mock';
 import report from './index';
 import { ErrorEvent } from './types';
 
+const local = 'https://analytics.arigato.tools';
 const stage = 'https://analytics.stage.manifold.co';
 const prod = 'https://analytics.manifold.co';
```
dangodev: cc @domenicrosati and @tainacleal: this is where we’d update those endpoints. How should these change?

src/utils/graphqlFetch.ts
```diff
@@ -4,6 +4,7 @@ import { waitForAuthToken } from './auth';
 
 interface CreateGraphqlFetch {
   endpoint?: () => string;
+  env?: 'local' | 'stage' | 'prod';
```
dangodev: @sslotsky I was confused on whether or not I can just import `connection.env` here. Is that all the same instance?

src/packages/analytics/index.spec.ts
domenicrosati: awesome.

Lets just at `/v1/ui/events` for now

dangodev: For errors too?

src/utils/graphqlFetch.ts
dangodev: Actually, wait—I actually want to keep in this way. If I import `connection` in this file, then we have a circular dependency. It may work, but I like to avoid those as they can break things.

src/packages/analytics/index.spec.ts
domenicrosati: umm i am thinking it through i'll think out loud on slack

PR #710, opened 2019-11-08
Add page visibility check to graphql fetch.

TerrySmithDC:
## Reason for change

Refreshing authorization was failing on attempts while the page did not focus. This check prevents retry's till the page is brought into focus again.

## Testing

Local black magic @domenicrosati is performing.

Alternatively make sure you're authenticated, open up the network tab in it's own winder, then focus on another tab and watch for manifold network requests.

## Checklist

### DIFF
--- a/src/utils/graphqlFetch.ts
+++ b/src/utils/graphqlFetch.ts
@@ -121,7 +121,7 @@ export function createGraphqlFetch({
         return e.extensions && e.extensions.type === 'AuthFailed';
       });
 
-      if (authExpired) {
+      if (authExpired && !document.hidden) {
         setAuthToken('');
         if (attempts < retries) {
           return waitForAuthToken(getAuthToken, wait(), () => graphqlFetch(args, attempts + 1));


src/utils/graphqlFetch.ts
```diff
@@ -121,7 +121,7 @@ export function createGraphqlFetch({
         return e.extensions && e.extensions.type === 'AuthFailed';
       });
 
-      if (authExpired) {
+      if (authExpired && !document.hidden) {
```
sslotsky: What I'd really like to accomplish is to pause _all_ API activity if the document is hidden, and then resume that activity when it becomes visible again.

This check will prevent us from retrying if the auth has expired while the document is hidden. But if the auth hasn't expired, we'll have polling loops that keep firing off requests every three seconds, which is unnecessary.

There is a [visibility change](https://developer.mozilla.org/en-US/docs/Web/API/Document/onvisibilitychange) event that we can listen for that would allow us to deal with this. That event could prompt us to flip a `paused` flag in the connection state. If we enter `graphqlFetch` and find that `paused` evaluates to true, than we can return some sort of `waitForPageVisibility` promise that works similar to how `waitForAuthToken` works, but without the timeout logic.

TerrySmithDC: Thanks Sam, I'll give that a shot. I'm still trying to get a local testing environment set up so I can actually test this locally.

sslotsky: Cool, hit me up any time if you have questions!

Also I realized that the `paused` flag is totally unnecessary since we have the `document.hidden` flag, but everything else I suggested still stands.

sslotsky: Oh, and here's another thought while I'm at it: the components that actually perform the polling could just skip their own API calls if `document.hidden` is true. So for example `manifold-resource-list` is calling `fetchResources` every 3 seconds.So we could short circuit that function easily enough. 

```ts
  fetchResources = async () => {
    if (!this.graphqlFetch || document.hidden) {
      return;
    }
```

That could fall into the quick-win category, whereas pausing _all_ API activity would be the robust solution in case more polling loops are added that fail to account for that.

PR #724, opened 2019-11-13
Add metrics opt-in

dangodev:
## Reason for change

Adds a `<manifold-connection metrics>` property to opt-in for analytics.

This PR **basically does nothing** right now, because we don’t have any analytics events. But it gives us a hook on the connection state to see if we should send analytics events or not.

## Checklist

### DIFF
--- a/src/components/manifold-connection/manifold-connection.tsx
+++ b/src/components/manifold-connection/manifold-connection.tsx
@@ -8,11 +8,12 @@ import connection from '../../state/connection';
 @Component({ tag: 'manifold-connection' })
 export class ManifoldConnection {
   @Prop() env?: 'local' | 'stage' | 'prod';
+  @Prop() metrics?: boolean;
   @Prop() waitTime?: number | string;
 
   @loadMark()
   componentWillLoad() {
-    connection.initialize({ env: this.env, waitTime: this.waitTime });
+    connection.initialize({ env: this.env, metrics: this.metrics, waitTime: this.waitTime });
   }
 
   @logger()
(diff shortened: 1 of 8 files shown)


src/components/manifold-connection/manifold-connection.tsx
```diff
@@ -8,11 +8,12 @@ import connection from '../../state/connection';
 @Component({ tag: 'manifold-connection' })
 export class ManifoldConnection {
   @Prop() env?: 'local' | 'stage' | 'prod';
+  @Prop() metrics?: boolean;
```
dangodev: I went with `<manifold-connection metrics>` for the API. My reasoning is that `analytics` feels like a skeezy word to me? And we’re not collecting full analytics; just anonymous metrics. I want this to feel, to platforms, like their users’ privacy isn’t being compromised (it isn’t)

dangodev: Is this a good naming?

domenicrosati: I like it alot

src/utils/graphqlFetch.ts
```diff
@@ -6,11 +6,12 @@ import awaitPageVisibility from './awaitPageVisibility';
 interface CreateGraphqlFetch {
   endpoint?: () => string;
   env?: () => 'local' | 'stage' | 'prod';
-  wait?: () => number;
-  retries?: number;
   getAuthToken?: () => string | undefined;
-  setAuthToken?: (token: string) => void;
+  metrics?: () => boolean;
```
dangodev: So how this will work is something like this:

```ts
if (metrics()) {
  analytics(/* send some metric event */)
}
```

We’ll just wrap any analytics events like this.

PR #726, opened 2019-11-13
Add rtt_graphql metrics event

dangodev:
## Reason for change

Adds a `rtt_graphql` event to test the performance of GraphQL as perceived by the end-user

Screen Shot 2019-11-13 at 13 35 56

## Testing

In Storybook, see that `analytics.manifold.co` is called after every GraphQL request

## Checklist

### DIFF
--- a/src/global/settings.ts
+++ b/src/global/settings.ts
@@ -1,0 +1,1 @@
+export const METRICS_ENABLED = false; // default setting for whether or not analytics.manifold.co receives metrics events
--- a/src/packages/analytics/index.ts
+++ b/src/packages/analytics/index.ts
@@ -7,7 +7,7 @@ const endpoint = {
 };
 
 interface AnalyticsOptions {
-  env?: 'local' | 'stage' | 'prod';
+  env: 'local' | 'stage' | 'prod';
 }
 
 /**
@@ -21,15 +21,8 @@ interface AnalyticsOptions {
  * @param {Object} [options] Analytics options
  * @param {string} [options.env] 'prod' (default) or 'stage'
  */
-export default function report(
-  evt: AnalyticsEvent,
-  userOptions: AnalyticsOptions = { env: 'prod' }
-) {
-  const options: AnalyticsOptions = {
-    env: 'prod',
-    ...(userOptions || {}),
-  };
-  const url = endpoint[options.env || 'prod'];
+export default function report(evt: AnalyticsEvent, options: AnalyticsOptions) {
+  const url = endpoint[options.env];
   return fetch(url, {
     method: 'POST',
     body: JSON.stringify({
--- a/src/packages/analytics/index.ts
+++ b/src/packages/analytics/index.ts
@@ -10,6 +10,16 @@ interface AnalyticsOptions {
   env: 'local' | 'stage' | 'prod';
 }
 
+function stringifyProperties(evt: AnalyticsEvent) {
+  return {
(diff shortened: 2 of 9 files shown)


src/utils/graphqlFetch.ts
```diff
@@ -90,6 +93,7 @@ export function createGraphqlFetch({
       report({ message: `${e.statusText || e.status}` }, { env: env(), element });
       return Promise.reject(e);
     });
+    const rttEnd = performance.now(); // end RTT timer
```
dangodev: @markjamieson I felt that measuring the response, and only the response, would be best for RTT. I noticed that we were technically measuring other JS operations, such as JSON parsing. While in all likelihood that’s not significant, I wanted to be scientific.

But does this make sense / do you agree?

stories/data/connectionDecorator.js
```diff
@@ -15,7 +15,7 @@ export const manifoldConnectionDecorator = storyFn => {
   localStorage.setItem('manifold_api_token', token); // update localStorage to persist
 
   return `
-  <manifold-connection env="${env}">
+  <manifold-connection env="${env}" metrics>
```
dangodev: Add metrics to Storybook

src/global/settings.ts
```diff
@@ -0,0 +1 @@
+export const METRICS_ENABLED = false; // default setting for whether or not analytics.manifold.co receives metrics events
```
dangodev: This is kind of an awkward new file, but as this pops up in a few places, I wanted to abstract it somewhere that other files can get to it, without someone having to hunt down this value all over the place if it changed.

I couldn’t put it in `state/connection.ts` because that would cause circular deps. Open to other suggestions if anyone has any

src/packages/analytics/index.ts
```diff
@@ -21,15 +21,8 @@ interface AnalyticsOptions {
...
-export default function report(
-  evt: AnalyticsEvent,
-  userOptions: AnalyticsOptions = { env: 'prod' }
-) {
-  const options: AnalyticsOptions = {
-    env: 'prod',
-    ...(userOptions || {}),
-  };
-  const url = endpoint[options.env || 'prod'];
+export default function report(evt: AnalyticsEvent, options: AnalyticsOptions) {
```
dangodev: I changed the analytics function to require `{ env: 'prod' }` as a required param. Reason is: I wrote this and I just messed it up. This will make it easier on future implementers.

src/packages/analytics/types.ts
```diff
@@ -20,26 +20,26 @@ export type EventTypes =
...
       properties: {
-        time: number;
+        duration: number;
       };
     }
   | {
       name: 'token_received';
       properties: {
-        time: number;
+        duration: number;
```
dangodev: Use `duration` for consistency

src/packages/analytics/index.ts
```diff
@@ -21,19 +34,12 @@ interface AnalyticsOptions {
...
+  const url = endpoint[options.env];
   return fetch(url, {
     method: 'POST',
     body: JSON.stringify({
-      ...evt,
+      ...stringifyProperties(evt),
```
dangodev: @domenicrosati 👀

src/utils/graphqlFetch.ts
markjamieson: Agreed.

PR #730, opened 2019-11-14
Report token_received event to analytics backend

markjamieson:
## Reason for change

Updates `<manifold-oauth>` to report `token_received` duration to analytics backend, rather than emitting duration along with `receiveManifoldToken` event. 

## Testing

1. In Storybook, go to Scenarios -> auth story
1. You should a `token_received` event sent in the network tab

## Checklist


src/components/manifold-oauth/manifold-oauth.tsx
```diff
@@ -13,27 +15,41 @@ export class ManifoldOauth {
...
         error: pumaToken.error,
-        duration: new Date().getTime() - (this.loadTime ? this.loadTime.getTime() : 0),
       });
+      if ((connection.metrics || METRICS_ENABLED) && this.loadTime){
+        report(
+          {
+            name: 'token_received',
+            type: 'metric',
+            properties: {
+              componentName: 'manifold-oauth',
```
sslotsky: Since we have a host element, can we use `this.el.tagName` instead of hard coding the component name?

markjamieson: Sure, will do

src/components/manifold-oauth/manifold-oauth.tsx
```diff
@@ -13,27 +15,41 @@ export class ManifoldOauth {
...
+    const receivedTime = performance.now();
 
     if (ev.origin === new URL(this.oauthURL).origin) {
       this.receiveManifoldToken.emit({
         token: pumaToken.access_token,
         expiry: pumaToken.expiry,
         error: pumaToken.error,
-        duration: new Date().getTime() - (this.loadTime ? this.loadTime.getTime() : 0),
       });
+      if ((connection.metrics || METRICS_ENABLED) && this.loadTime){
```
dangodev: I think you might have to call `connection.metrics()`? IIRC the function helps it get the value from `<manifold-connection>` and not the default value.

markjamieson: `createRestFetch` and `createGraphqlFetch` seem to expect a function, but `connection.metrics` itself is a boolean

markjamieson: Looks like it's just populated by the presence of the `metrics` attribute on `<manifold-connection>`, and it falls back to `METRICS_ENABLED` if that attribute isn't present

dangodev: Ah OK. Yeah it seems to fire if `metrics` is present, and not fire if it’s not. So seems to be working as-is

dangodev: I guess the REST fetch & GraphQL fetchers have to do that because they get initialized before `<manifold-connection>` or something. TBH I don’t really understand what goes on there; I just remember trying a boolean and having to change it to a function. But again, your PR seems to work fine.

PR #733, opened 2019-11-14
Use typed query for manifold-resource-container

dangodev:
## Reason for change

Converts `<manifold-resource-container>` to use a typed GraphQL query.

This is cleanup work that will make the upcoming ticket https://github.com/manifoldco/engineering/issues/9877 easier.

## Testing

All tests should pass. This doesn’t change any production code; it just makes the types more resilient.

## Checklist


src/data/resource.tsx
```diff
@@ -1,18 +1,16 @@
 import { h } from '@stencil/core';
 import { createProviderConsumer } from '@stencil/state-tunnel';
-import { Gateway } from '../types/gateway';
-import { Resource } from '../types/graphql';
+import { GetResourceQuery } from '../types/graphql';
 
 export interface ResourceState {
-  data?: Gateway.Resource;
```
dangodev: This finally removes `data` from the tunnel 🎉  (might rename it in a followup PR, but here I didn’t want to change anything more)

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -1,130 +1,12 @@
...
-                  cost
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-  }
-`;
```
dangodev: This deletes the Gateway mock in `<manifold-resource-mock>`

src/components/manifold-resource-plan/manifold-resource-plan.tsx
```diff
@@ -36,12 +32,13 @@ export class ManifoldResourcePlan {
...
 
+    // TODO: remove `as` and update types
     return (
       <manifold-plan-details
         scrollLocked={false}
-        plan={plan}
-        product={product}
-        region={region}
+        plan={this.gqlData.plan}
+        product={this.gqlData.plan.product as Product}
```
dangodev: This was basically the only temporary hack I had to make. Ideally we’d use generated types for this, too, but that’ll be a whole PR of work once we go down the plan rabbit hole.

But good news is that as long as the shape of `resource.graphql` matches `plan-list.graphql`, we’re fine. But converting that will take some time.

src/components/manifold-resource-container/resource.graphql
```diff
@@ -0,0 +1,116 @@
...
+              unit
+              costTiers {
+                limit
+                cost
+              }
+            }
+          }
+        }
+      }
+      configurableFeatures(first: 20) {
```
sslotsky: I didn't catch this when we first did the conversion, but... what do we use all the regions and features for? Is it for plan resize, or something else?

dangodev: We need the resource region for display. And we need the features for plan resize. While we probably don’t need to get all regions for resize (you can’t reprovision in a different region), when we convert the plan selector over the types will have to match if we want to reuse the same component.

sslotsky: When we load the plan selector, we load all of the plans with their features, right? And that includes the selected plan, so we'd be loading that information twice. Maybe I'm wrong but I _think_ the only relevant information that the plan selector needs about the resource is the selected plan ID and region.

dangodev: We don’t load the plan selector through this; we load `<manifold-plan-details>`. It’s just that `<manifold-plan-details>` needs the exact same type shape, which includes all regions (right now we’re cheating, and using `Plan`, but when we type queries, the types will need to match)

sslotsky: Ohh, gotcha. In that case, I think we need to put `manifold-plan-details` on the radar for components that need to account for pagination and currently do not. How do we go about doing that? Just have a chat with Meg about it?

dangodev: Yeah. She started working on this change and has a design somewhere

sslotsky: Ok. I was aware that she had started this for the plan selector, but was not aware about the plan details. How do we make sure that we are tracking everything that needs pagination support?

dangodev: Just stay in communication with Meg; she can point you in the right direction in regard to that. That will probably be MUI’s responsibility to figure out a system to track that.

PR #734, opened 2019-11-14
Minify bundle

dangodev:
## Reason for change

Free performance 🤷‍♂ 

## Testing

This should be safe-ish because I went with “light” minification which means most unsafe code mutations were avoided (see https://github.com/terser/terser#terser-fast-minify-mode). This has been tested in Render Dashboard and everything seems to work as expected. Though it wouldn’t hurt for the reviewer to poke through Storybook and make sure there are no more errors than usual.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "types": "dist/types/components.d.ts",
   "collection": "dist/collection/collection-manifest.json",
   "scripts": {
-    "build": "npm run update-version && stencil build",
+    "build": "npm run update-version && stencil build && node scripts/minify",
     "build:docs": "npm run copy-changelog && cd docs && npm i && npm run build",
     "build:storybook": "npm run build && build-storybook && npm run copy-changelog",
     "build:watch": "npm run update-version && stencil build --watch",
@@ -42,7 +42,6 @@
     "generate:specs:provisioning": "npx @manifoldco/swagger-to-ts src/spec/provisioning/v1.yaml -w 'export namespace Provisioning' -o src/types/provisioning.ts",
     "generate:specs:connector": "npx @manifoldco/swagger-to-ts src/spec/connector/v1.yaml -w 'export namespace Connector' -o src/types/connector.ts",
     "happo": "npm run build && npm run generate:gql && manifold run -t manifold -p manifold-ui-ci -- happo run",
-    "testcafe:dev": "DEPLOY_URL=http://localhost:6006 npx testcafe chrome e2e/",
     "happo-ci-circleci": "npm run build && npm run generate:gql && happo-ci-circleci",
     "heroku-postbuild": "npm run build:docs",
     "lint": "npm run lint:js & npm run lint:css & cd docs && npm i && npm run lint",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -58,22 +57,28 @@
...
       "path": "./dist/esm/loader.mjs",
-      "maxSize": "15 kB",
+      "maxSize": "10 kB",
       "compression": "none"
     },
     {
-      "path": "./dist/esm/manifold-!(mock|number-input|button_3)*.js",
-      "maxSize": "18 kB",
+      "path": "./dist/esm/manifold-!(button_3|mock|number-input)*.@(js|mjs)",
+      "maxSize": "10 kB",
```
dangodev: Proof of minification: `18 KB` -> `10 KB`

package.json
```diff
@@ -58,22 +57,28 @@
...
+      "compression": "none"
+    },
+    {
+      "path": "./dist/esm/manifold-button_3*.js",
+      "maxSize": "15 kB",
```
dangodev: Added because this was missing before (it wasn’t being profiled)

package.json
sslotsky: :clap: :clap: :clap:

package.json
sslotsky: I've never really looked at this config closely and now I.... have questions.

1. The naming is weird. Why `_3`? Do we have any way of knowing that it will consistently append a `_3`?
1. Do we not care about the size of the `esm-esm5/` folder? AFAICT that's the folder that actually gets used by Render.

scripts/minify.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Minifies Stencil output after build.
+ * This was chosen instead of rollup-plugin-terser for safety (that affects file contents)
```
sslotsky: How will this technique hold up if we also decide to generate source maps?

sslotsky:
Would we be able to test this PR on the CDN if we tagged an alpha release?

package.json
dangodev: That’ll change when the components / dependencies change. So yeah we’ll have to update it if that shifts, but that shouldn’t happen unless the button family is changed.

But I _think_ that number is how many components it’s bundled together. It does bundle components together into one chunk when it thinks they’re almost always served together. See https://stenciljs.com/docs/output-targets#differential-bundling

scripts/minify.js
dangodev: Good question. Terser itself can generate source maps, I think. But you can also choose to not mangle names. Anyway we have several options here. It might be a bit of fiddling but I’m confident we can figure something out.

dangodev: I can explain more of what I was seeing with rollup-plugin-terser before maybe in Slack or something. It’s actually pretty interesting, and has to do with Rollup’s treeshaking and not rollup-plugin-terser.

dangodev:
> Would we be able to test this PR on the CDN if we tagged an alpha release?

Yeah that’d be a good idea

package.json
sslotsky: Any thoughts on my second question?

dangodev: Not sure. I haven’t, because I just assumed that was based off the `esm` folder but with some transformations. I don’t have any qualms about tracking it if you think it’s important.

sslotsky: Can we make a ticket to at least investigate the relationship between the two and figure out if it's prudent to track both? We certainly don't need to do anything about it in this PR.

sslotsky:
I know you're putting this on hold, but just sharing a thought before I forget about it:

If we minify the stuff that Render uses in their Dashboard, it actually hurts our ability to debug. If they're the ones responsible for minifying it, then

1. we can step through the code as-is in dev mode, and
2. if they agree to upload source maps in production, then we can step through the code there as well.

If we minify that code ourselves, we can't do any of that, I think.

sslotsky:
Webpack is going to wrap our JavaScript files with some webpack specific code so our source maps wouldn't be accurate. I think the only place where minifying our own code and adding source maps actually makes sense would be the CDN. I mean, we could try it, but I don't think it will work properly.

dangodev:
Closing; this was a PoC and it makes debugging harder.

PR #743, opened 2019-11-19
Adds features data (free boolean) to manifold-plan-selector custom event

TerrySmithDC:
## Reason for change

Adds free flag to custom plan selector events, which helps integration engineers to better add custom validations around Manifold UI.

## Testing

From storybook->plan selector ensure `manifold-planSelector-change` and `manifold-planSelector-load` custom events include features data.

## Checklist

### DIFF
--- a/src/components/manifold-plan-details/manifold-plan-details.tsx
+++ b/src/components/manifold-plan-details/manifold-plan-details.tsx
@@ -18,6 +18,9 @@ interface EventDetail {
   productLabel: string | undefined;
   regionId?: string;
   regionName?: string;
+  feature?: {
+    free: boolean;
+  };
 }
 
 @Component({
@@ -48,6 +51,9 @@ export class ManifoldPlanDetails {
       productLabel: this.product && this.product.label,
       regionId: defaultRegion && defaultRegion.id,
       regionName: defaultRegion && defaultRegion.displayName,
+      feature: {
+        free: newPlan.free,
+      },
     };
 
     if (!oldPlan) {
@@ -83,6 +89,9 @@ export class ManifoldPlanDetails {
         productLabel: this.product && this.product.label,
         regionId: defaultRegion && defaultRegion.id,
         regionName: defaultRegion && defaultRegion.displayName,
+        feature: {
+          free: this.plan.free,
+        },
       };
       this.planLoad.emit(detail);
       // reset features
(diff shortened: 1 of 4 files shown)


src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -18,6 +18,9 @@ interface EventDetail {
   productLabel: string | undefined;
   regionId?: string;
   regionName?: string;
+  feature?: {
```
dangodev: Why is this in a `feature` object? Just curious

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -123,6 +135,9 @@ export class ManifoldPlanDetails {
         productLabel: this.product.label,
         regionId: e.detail.value,
         regionName: defaultRegion && defaultRegion.displayName,
+        feature: {
+          free: this.plan.free,
```
dangodev: This looks fine but we’ll need to update the `manifold-plan-details.spec.ts` test to ensure that someone else doesn’t break this new addition. Events are super important! If they are missing data, this is the quickest way to break a platform, so we always want to make sure to add tests for events.

src/components/manifold-plan-details/manifold-plan-details.tsx
dangodev: I’d recommend just having something on the top-level, like `freePlan`. It not only is safer not having a nested object, but a plan’s cost may be free independent of features.

TerrySmithDC: This was a bit of an assumption on my part based on [The events docs](https://ui.sandbox.manifold.co/components/plan-selector)

```js
document.addEventListener('manifold-planSelector-change', ({ detail }) => {
  console.log(detail);
});
// {
//   planId: '2357v8j36f5h866c32ddwwjxvfe8j',
//   planLabel: 'nvidia-1080ti-100gb-ssd',
//   planName: 'NVIDIA 1080TI',
//   productLabel: 'zerosix',
//   features: {
//     // …
//   },
// }
```
Based off that I was under the impression you guys planned to have additional data under features so I added free there. Where would you prefer it?

I'm also realizing I didn't pluralize feature like I should have.

dangodev: Oh right. Originally `features` was an object letting you know what configurable features the user selected. So `features` should probably be reserved for when we re-introduce that (https://github.com/manifoldco/engineering/issues/9424).

dangodev: For example, in JawsDB, if a user selected `5 GB` of storage, that’d be `{ features: { storage: '5 GB' } }`, etc.

TerrySmithDC: Ok I've moved it out of features into just `freePlan` 💃

PR #744, opened 2019-11-19
Test cleanup

dangodev:
## Reason for change
Edit: this PR is just test cleanup now. No functionality should be changed.

## Checklist


src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -80,7 +78,7 @@ export class ManifoldResourceContainer {
   @logger()
   render() {
     return (
-      <ResourceTunnel.Provider state={{ gqlData: this.gqlData, loading: this.loading }}>
+      <ResourceTunnel.Provider state={{ gqlData: this.gqlData }}>
```
dangodev: 🚨 This is a big PR, but this is the central change here 👆. All the other changes cascade from this.

I feel it’s better to infer loading than have an independent piece of state that is deeply-connected to `gqlData`, for reasons outlined in the PR description.

src/components/manifold-resource-deprovision/manifold-resource-deprovision.spec.ts
```diff
@@ -1,51 +1,58 @@
...
 
-    await page.waitForChanges();
+  const root = page.root as HTMLDivElement;
+  root.appendChild(component);
+  await page.waitForChanges();
 
-    const button = element.querySelector('manifold-data-deprovision-button');
-    expect(button).toBeDefined();
+  return { page, component };
+}
```
dangodev: Updates test to use our new preferred way of running tests.

src/components/manifold-resource-deprovision/manifold-resource-deprovision.spec.ts
```diff
@@ -1,51 +1,58 @@
...
-      expect(button.loading).toBeFalsy();
-      expect(button.resourceId).toEqual(resource.id);
-      expect(button.resourceLabel).toEqual(resource.label);
-    }
+    it('[gqlData]: button not disabled if present', async () => {
+      const { page } = await setup({ gqlData: resource as GetResourceQuery['resource'] });
+      const button = page.root && page.root.querySelector('button');
+      expect(button && button.getAttribute('disabled')).toBeNull();
+    });
   });
```
dangodev: This spec test was made more resilient, because we’re not testing implementation details and instead reaching down to see if the bottom-level `button` is disabled or not.

src/components/manifold-resource-rename/manifold-resource-rename.spec.ts
```diff
@@ -1,51 +1,69 @@
...
-      expect(button.resourceLabel).toEqual(resource.label);
-    }
+    it('[newLabel]: button not disabled if present', async () => {
+      const { page } = await setup({
+        newLabel: 'my-new-resource',
+        gqlData: resource as GetResourceQuery['resource'],
+      });
+      const button = page.root && page.root.querySelector('button');
+      expect(button && button.getAttribute('disabled')).toBeNull();
+    });
```
dangodev: Basically same improvements here. Test was updated for the `rename` button to make our tests more consistent, and now the test is more reliable because it tests HTML.

src/components/manifold-resource-sso/manifold-resource-sso.tsx
```diff
@@ -7,8 +7,8 @@ import { GetResourceQuery } from '../../types/graphql';
 
 @Component({ tag: 'manifold-resource-sso' })
 export class ManifoldResourceSso {
+  @Prop() disabled?: boolean;
```
dangodev: Fixed inconsistency: some Web Components allow `disabled` prop; others don’t. I made all the button components consistent by letting them all pass `disabled` through (and adding tests).

This is a deeply-related change to the loading state, because the loading state’s purpose was to control whether buttons are disabled or not.

src/components/manifold-resource-product/manifold-resource-product.spec.ts
```diff
@@ -1,63 +1,60 @@
...
+      serviceCard &&
+      serviceCard.shadowRoot &&
+      serviceCard.shadowRoot.querySelector('manifold-skeleton-text');
 
-    expect(element).toEqualHtml(`
-      <manifold-resource-product>
-        <manifold-service-card-view description="This is a loading product..." logo="loading.jpg" name="loading..." skeleton="">
-          <manifold-forward-slot slot="cta"></manifold-forward-slot>
-        </manifold-service-card-view>
-      </manifold-resource-product>
```
dangodev: Snapshot-y tests were replaced with element queries. Makes tests more stable and clearer in purpose.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@manifoldco/ui",
   "description": "Manifold UI",
-  "version": "0.6.5",
+  "version": "0.7.0",
```
sslotsky: I've been noticing this field changing automatically for me, which I assume is what happened here. Do you know what causes it?

src/components/manifold-data-deprovision-button/manifold-data-deprovision-button.tsx
```diff
@@ -121,7 +121,11 @@ export class ManifoldDataDeprovisionButton {
   @logger()
   render() {
     return (
-      <button type="submit" onClick={() => this.deprovision()} disabled={!this.resourceId}>
+      <button
+        type="submit"
+        onClick={() => this.deprovision()}
```
sslotsky: Another opportunity for Point Free Style, if you feel so inclined :smile:

package.json
davidleger95: i remember @dangodev mentioning something about auto versioning. I think it grabs the latest changelog version when you run dev.

src/components/manifold-data-sso-button/manifold-data-sso-button.tsx
```diff
@@ -101,7 +101,7 @@ export class ManifoldDataSsoButton {
       <button
         type="submit"
         onClick={this.sso}
-        disabled={this.loading}
+        disabled={this.disabled || (!this.resourceId && !this.resourceLabel)}
```
sslotsky: So if I understand correctly, this component can function with a `resourceId` _or_ a `resourceLabel`, but it doesn't require both to be present?

src/components/manifold-mock-resource/manifold-mock-resource.tsx
```diff
@@ -101,20 +101,18 @@ const GraphQLResource: GetResourceQuery['resource'] = {
...
-      this.loading = false;
       this.resource = this.gqlMock;
     }, 300);
   }
 
   @logger()
   render() {
     return (
-      <ResourceTunnel.Provider state={{ loading: this.loading, gqlData: this.gqlMock }}>
+      <ResourceTunnel.Provider state={{ gqlData: this.gqlMock }}>
```
sslotsky: This is the first time I've noticed that we have a `manifold-mock-resource`. Why does this exist?

src/components/manifold-resource-container/manifold-resource-container.tsx
sslotsky: So as I understand this then, if there's no data present, we infer that we are in the process of loading. I think that also means that unless we are properly handling errors, we will continue to show skeleton screens if we receive an error. Is that right?

src/components/manifold-resource-sso/manifold-resource-sso.tsx
sslotsky: Here's one thing I don't understand (and maybe I'm just missing it): lots of components are now accepting a `disabled` prop, but I don't see `disabled` being passed into most of them at any time.

src/components/manifold-mock-resource/manifold-mock-resource.tsx
dangodev: I don’t remember; I think it was made for the docs at one point. Or possibly Storybook before Knobs. @Minivera is this still needed?

src/components/manifold-data-sso-button/manifold-data-sso-button.tsx
dangodev: Right

package.json
dangodev: Yeah it’s fine. IIRC pulling the version from Git works fine in CI, but when we started placing the UI version in headers, Now didn’t have Git context so this was easier for testing. This may fluctuate based on whether or not someone recently fetched locally, but doesn’t matter too much because the CI that deploys to npm will always be correct.

I’m open to other ideas if someone can think of how else to get that value into Now without this method

src/components/manifold-mock-resource/manifold-mock-resource.tsx
Minivera: This component's goal is to allow for quick prototyping of a resource details page, I used it extensively when creating the resource details page on render. It quickly loads a valid resource and simulates a loading time.

I'm not against removing it if you feel it's useless, but I thought that use case was pretty useful for platforms who want to start getting their marketplace together before they have the oauth ready.

dangodev:
> Although I think my long term vision for the container was that it would keep track of both loading and error states

I have reasons for wanting to simplify `loading`, and adding the error state would be required either way, but I don’t mind keeping it if others find it useful. 

But my long-term version for the container is that it goes away, and we find a more clever way for components to share queries based on matching resource information. That way each component decides when it should be in a “loading” state, or when it should silently just update because it’s polling. As-is, I think it’s too easy to have loading flickers if the loading logic ever changed.

src/components/manifold-resource-sso/manifold-resource-sso.tsx
dangodev: Well in Render, we had an issue of that property being missing, so we’d add them ad hoc or upon pain of deadline we’d find a temporary workaround. But disabling the buttons based on external logic is always helpful.

sslotsky:
> But my long-term version for the container is that it goes away

Agree :100: with that, FWIW. And I'm hoping we can drive towards that goal sooner than later. I also hope that polling goes away and that we can do subscriptions instead. I do think your point about introducing a flicker if we change the loading logic is valid.

dangodev:
Based on feedback, I reverted the `loading` changes and now it just makes the tests more consistent, and tests a few more things that weren’t before. Would love a ✅ if this seems good.

src/components/manifold-image-gallery/manifold-image-gallery.e2e.ts
```diff
@@ -1,40 +0,0 @@
...
-      (elm: any, props: any) => {
-        elm.images = props.images;
-      },
-      { images }
-    );
-    await page.waitForChanges();
-    const els = await page.findAll('manifold-image-gallery >>> img[data-test="thumbnail"]');
-    expect(els.length).toBe(4);
-  });
-});
```
dangodev: This E2E test kept flaking and failing just now, even though it was untouched.

I removed it since we’re catching changes here with VRT. Is that all right?

PR #748, opened 2019-11-19
Use mutation observer to track load event start time

markjamieson:
## Reason for change

Proof of concept for using a `MutationObserver` on `<manifold-connection>` to track load event start time (i.e. time that any `<manifold-*>` element is inserted into the DOM).

This PoC is blocked by `time_to_render` event being inaccurate.

If we proceed with this approach, I also need some code in the mutation callback to cleanup removed nodes.

## Testing

1. There is currently an _illegal_ `console.log` in the `logger` decorator where you can see logs of the format `<tagname> <time from load to first render> <time from component js initialization to first render>`
2. You should see that the first number is always greater than or equal to the second, which accounts for the time from component being added to the DOM until that component's `componentWillLoad` method is called

## Checklist

### DIFF
--- a/src/components/manifold-oauth/manifold-oauth.tsx
+++ b/src/components/manifold-oauth/manifold-oauth.tsx
@@ -31,7 +31,7 @@ export class ManifoldOauth {
         expiry: pumaToken.expiry,
         error: pumaToken.error,
       });
-      if ((connection.metrics || METRICS_ENABLED) && this.loadTime){
+      if ((connection.metrics || METRICS_ENABLED) && this.loadTime) {
         report(
           {
             name: 'token_received',
--- a/src/global/app.ts
+++ b/src/global/app.ts
@@ -1,0 +1,39 @@
+const connectionEl = document.getElementsByTagName('manifold-connection').length
+  ? document.getElementsByTagName('manifold-connection')[0]
+  : undefined;
+if (connectionEl) {
+  performance.mark(`${connectionEl.tagName}-load-start`);
+  Array.prototype.slice
+    .call(connectionEl.getElementsByTagName('*'))
+    .filter((el: HTMLElement) => el.tagName.startsWith('MANIFOLD-'))
+    .forEach((el: HTMLElement) => {
+      const markName = `${el.tagName}-load-start`;
+      if (performance.getEntriesByName(markName, 'mark').length === 0) {
+        performance.mark(markName);
+      }
+    });
+}
+
+let observer: MutationObserver;
+function mutationCallback(mutationsList: MutationRecord[]) {
+  mutationsList
+    .filter(mutation => mutation.type === 'childList')
+    .forEach(mutation => {
+      if (mutation.addedNodes) {
+        mutation.addedNodes.forEach((node: HTMLElement) => {
+          if (node.tagName && node.tagName.startsWith('MANIFOLD-')) {
+            const markName = `${node.tagName}-load-start`;
+            if (performance.getEntriesByName(markName, 'mark').length === 0) {
+              performance.mark(markName);
+            }
+            if (node.shadowRoot) {
+              observer.observe(node.shadowRoot, { childList: true, subtree: true });
+            }
+          }
+        });
+      }
+    });
+}
+
+observer = new MutationObserver(mutationCallback);
+observer.observe(document, { childList: true, subtree: true });
--- a/src/global/app.ts
+++ b/src/global/app.ts
@@ -1,16 +1,15 @@
+import { mark } from '../packages/analytics';
+
 const connectionEl = document.getElementsByTagName('manifold-connection').length
   ? document.getElementsByTagName('manifold-connection')[0]
   : undefined;
 if (connectionEl) {
-  performance.mark(`${connectionEl.tagName}-load-start`);
+  mark(connectionEl, 'load');
   Array.prototype.slice
     .call(connectionEl.getElementsByTagName('*'))
     .filter((el: HTMLElement) => el.tagName.startsWith('MANIFOLD-'))
     .forEach((el: HTMLElement) => {
-      const markName = `${el.tagName}-load-start`;
-      if (performance.getEntriesByName(markName, 'mark').length === 0) {
-        performance.mark(markName);
-      }
+      mark(el, 'load');
     });
 }
 
@@ -22,10 +21,7 @@ function mutationCallback(mutationsList: MutationRecord[]) {
       if (mutation.addedNodes) {
         mutation.addedNodes.forEach((node: HTMLElement) => {
           if (node.tagName && node.tagName.startsWith('MANIFOLD-')) {
-            const markName = `${node.tagName}-load-start`;
-            if (performance.getEntriesByName(markName, 'mark').length === 0) {
-              performance.mark(markName);
-            }
+            mark(node, 'load');
             if (node.shadowRoot) {
               observer.observe(node.shadowRoot, { childList: true, subtree: true });
             }
--- a/src/state/connection.ts
+++ b/src/state/connection.ts
@@ -11,7 +11,7 @@ const INITIALIZED = 'manifold-connection-initialize';
 
 interface NodeMapEntry {
   el: HTMLElement;
-  loadTime: number;
+  loadMark: number;
 }
 
 export interface NodeMap {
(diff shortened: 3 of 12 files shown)


markjamieson:
Screen Shot 2019-11-19 at 5 32 05 PM

src/state/connection.ts
```diff
@@ -9,18 +9,32 @@ export type Subscriber = (newToken?: string) => void;
 
 const INITIALIZED = 'manifold-connection-initialize';
 
+interface NodeMapEntry {
+  el: HTMLElement;
+  loadTime: number;
+}
+
+export interface NodeMap {
+  [tagName: string]: NodeMapEntry[];
```
markjamieson: Indexing the NodeMap by tagName to make node lookup faster.

src/utils/logger.tsx
```diff
@@ -55,6 +56,21 @@ export default function logger<T>() {
               duration: this.performanceRenderedMark - this.performanceLoadMark,
             },
           });
+          if (this.el) {
```
markjamieson: This code requires that all components we want to track have a `Element()` property named `el`. Maybe a safer approach is to pass the element as an argument to the `logger` decorator, rather than assuming the name of the property. Comments welcome.

sslotsky: Does this remove the need for the `loadMark` decorator? If so, I think requiring an `Element()` is a good compromise, and if passing that element into the `logger` decorator works, that sounds solid to me.

markjamieson: I'm not sure if it removes the need. I'd like to hear from @dangodev if there's value in both measuring time from load -> render, and time from component instantiation -> render.

src/state/connection.ts
```diff
@@ -91,6 +110,31 @@ export class ConnectionState {
...
+      .filter(mutation => mutation.type === 'childList')
+      .forEach(mutation => {
+        if (mutation.addedNodes) {
+          mutation.addedNodes.forEach((node: HTMLElement) => {
+            const tagName = node.tagName && node.tagName.toLowerCase();
+            if (tagName && tagName.startsWith('manifold-')) {
+              const entry = {
+                el: node,
+                loadMark: performance.now(),
+              };
```
dangodev: I like the approach of filtering by `manifold-`. This makes sense.

dangodev: nit: we can probably omit the `.toLowerCase()` and filter by `MANIFOLD-`. `tagName` should always be upper-case (like the good Lord intended when he made the HTML spec)

dangodev: q: How does this keep track of multiple instances of the same element? On the one hand, it’d be good to track that. But on the other, I hadn’t thought about the fact that if you have multiple instances of the same element, only the first load would be “cold” and need its JS downloaded. Subsequent renders would already have that JS file cached in the browser so the time would be different.

Maybe for that reason, rather than an object you should have a `Map()`, or something that only lets you set everything once?

src/utils/logger.tsx
dangodev: Yeah there’s value in tracking both, IMO. From something Leah said, even if we know something is fast, it’s good to track it so we have data to back it up. You could maybe even call it a “vanity metric.” But these do come up in convos, and it’s just nice to point to a chart to confirm that things like instantiation -> render is as fast as we say it is.

src/components/manifold-oauth/manifold-oauth.tsx
```diff
@@ -31,7 +31,7 @@ export class ManifoldOauth {
         expiry: pumaToken.expiry,
         error: pumaToken.error,
       });
-      if ((connection.metrics || METRICS_ENABLED) && this.loadTime){
+      if ((connection.metrics || METRICS_ENABLED) && this.loadTime) {
```
dangodev: I think I messed up the `lint` npm-script trying to run it in parallel with `&`. Thanks for fixing that.

dangodev:
I want to test this a little more, but at first glance these numbers seem right? It’s hard to tell because in Render, the JS files are serving within ms because they’re local, but I’m roughly seeing that `load` metric be JS load + parse time + instantiation -> render. But really hard to tell because all of those numbers are in `ms`.

I’m feeling really good about this PR! But I’m going to spin up a demo app and throw it on Now or something to test network connection to make sure these numbers are realistic.

src/state/connection.ts
markjamieson: I guess I was assuming that we wanted to keep track of each instance of each component. I was only mapping by tagName to prevent a linear search of all nodes on every lookup. (I know it's still a linear search, but over a significantly smaller set).

Now that I'm looking at your question though, I guess `time_to_render` on every instance of a component, and `load` on the first instance makes more sense. Is that what you're thinking? If so, it makes lookup more efficient. I was worried about causing slowdowns with this code, which is probably the worst thing we could do while trying to measure performance.

dangodev: Yeah I’m still thinking it through but I think that makes sense. Since we’re testing the JS download time, only the first instance is meaningful.

And yeah that’s tough to track, too, if our performance monitoring would affect performance. I suppose there are always web workers if that becomes an issue though.

I’m thinking maybe `render_with_spinner` and `render_with_data` are probably clearer names with what we want to test with connected components (what should we call the event name for things like buttons with no data? Not sure if we should reuse either of those names as it’ll affect the average).

markjamieson: I don't think we should worry about averages being meaningful. It's more useful to think of metrics like this as distributions. In DataDog I'm sure we can visualize `time_to_render` as a distribution where we can something like:

- 50% of components render within 200ms
- 75% render within 100ms
- 90% render within 25ms
- etc.

That way we don't have to worry about the average. What we get is a feel for the majority case, as well as the outliers.

dangodev: True, but I think the term “rendering” carries with it a lot of assumptions. From a user perspective, “time to render” means how long between when they load the page to our components load with data. From our perspective, we could define rendering as time it took to mutate the DOM.

I’d still like to be able to roughly divide render latency into buckets where we’re waiting on an API, and buckets where we’re not, somehow. Once we start logging metrics, if we change how they’re measured, then we invalidate all data up till that point.

Does that make sense? I agree with what you’re saying, but I’d at least like to figure out those 50%/75%/90% numbers based on whether or not they’re waiting on our APIs.

markjamieson: Maybe we can do `time_to_first_render` and `time_to_first_render_with_data`. Basically, on a render where we see that the component has already reported a `graphql_rtt` event, we can then report that second event.

markjamieson:
@dangodev I updated the PR to include the changes we discussed, specifically:
- Moving the MutationObserver to the global setup script
- Using performance.mark to mark the start and end of only the first instance of each component
- Immediately setting a `load-start` mark for any component already in the DOM at the time the setup script is executed

src/global/app.ts
```diff
@@ -0,0 +1,39 @@
+const connectionEl = document.getElementsByTagName('manifold-connection').length
+  ? document.getElementsByTagName('manifold-connection')[0]
+  : undefined;
+if (connectionEl) {
+  performance.mark(`${connectionEl.tagName}-load-start`);
```
dangodev: “Performance Mark” is my new nickname for you BTW

src/utils/logger.tsx
```diff
@@ -55,6 +56,24 @@ export default function logger<T>() {
...
+            const startMarks = performance.getEntriesByName(startMarkName, 'mark');
+            const endMarks = performance.getEntriesByName(endMarkName, 'mark');
+            if (startMarks.length && !endMarks.length) {
+              performance.mark(endMarkName);
+              /* eslint-disable no-console */
+              console.log(
+                el.tagName,
+                this.performanceRenderedMark - startMarks[0].startTime,
+                evt.detail.duration
+              );
```
dangodev: I’m seeing this in my console logs, but what is each # again? Could you also log the name of the event (`time_to_first_render` and `time_to_first_render_with_data`)?

dangodev: These seem to be correct at first glance but want to make sure.

dangodev:
Wanted to check on this again. Maybe rather than doing more `console.log`s, let’s just hook them up to the actual events in DataDog? And let’s test it in DataDog to make sure that the numbers are right in Storybook. If that all looks good let’s get this merged!

markjamieson:
@dangodev This PR has left discovery and is officially hoping to be shipped. As discussed, I'm tracking everything with performance marks in the format:
`<tagname>-<event name>-<'start'|'end'>`

This PR now reports `load`, `first_render`, and `first_render_with_data` events.

I tested in Storybook and the numbers look reasonable to me.

src/global/app.ts
```diff
@@ -0,0 +1,35 @@
+import { mark } from '../packages/analytics';
```
markjamieson: Importing in the global setup script might raise some eyebrows. I basically just moved code I would have duplicated here to a helper utility method, but I'm open to suggestions.

PR #751, opened 2019-11-20
Make image gallery display full images uncropped

davidleger95:
## Reason for change

- Image gallery was displaying images cropped to fit the aspect ratio

### Before

logdna-screenshots

### After

Screen Shot 2019-11-20 at 11 30 40 AM

## Testing

- go to product-page component in storybook
- choose a product with screenshots (logdna, db33)
- flip through screenshots
- screenshots should be sized to display the entire image and not be cropped

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Screenshots now display full images (uncropped). (#751)
+
 ## [0.7.2] - 2019-11-19
 
 ### Added
--- a/src/components/manifold-image-gallery/manifold-image-gallery.css
+++ b/src/components/manifold-image-gallery/manifold-image-gallery.css
@@ -54,6 +54,10 @@ img {
   width: 100%;
   height: 100%;
   overflow: hidden;
+
+  & img {
+    object-fit: contain;
+  }
 }
 
 .image-button {


davidleger95:
I want to get @megthesmith's thoughts on this. (1) Should the image gallery have a fixed ratio box that the image fits into [current], or (2) should the image gallery adjust based on the image ratio?

Option 2 will make the images as large as possible to fill available space, but that means different sized screenshots will make the UI jump around as the user flips through them.

megthesmith:
@davidleger95 So the purpose for the change was so that people could see more of the screenshot at first glance, without any UI getting cut off?

I don't love an interaction that bounces around due to varying sizes- can we make the height consistent and the width variable according to the image's inherent ratio? On the Ecosystem side, we'll have to talk to Kirsten about this change.

In the screencap you shared - is the main gallery image totally centered? It looks off by a little bit, like maybe a margin is borking it?

davidleger95:
Here's a better view of the after UI. It's centred, but might have looked slightly off because the 3 thumbnails almost span the full width of the container.

Screen Shot 2019-11-20 at 11 51 29 AM

dangodev:
The layout looks good! But we might want to set the image width while loading. Otherwise it thrashes a bit if the images are large and take a while to load:

Screen Shot 2019-11-20 at 08 52 53

davidleger95:
> can we make the height consistent and the width variable according to the image's inherent ratio

I believe this is what is happening. you can check it out here and select a product using the dropdown at the bottom: https://ui-dokhda3uw.now.sh/?path=/story/catalog--product-page

dangodev:
Also might not be fixable in this PR, but OAuth is missing value props, and the screenshot section is really big (not sure why the sidebar is shrinking so much). Just a heads up:

Screen Shot 2019-11-20 at 08 56 50

davidleger95:
> The layout looks good! But we might want to set the image width while loading.

This didn't change in this PR, the changes here affect the selected image, but I can look at this now as well

megthesmith:
@davidleger95 Thumbs up from me on this work as well! No more blockers to shipping this, similar to Drew, we'll iterate in the future on stuff.

davidleger95:
> it thrashes a bit if the images are large and take a while to load

There wasn't a quick fix for this and I don't think it takes priority right now. I'll throw an issue in the backlog and maybe we'll pick it up at some point.

PR #752, opened 2019-11-20
refactor product card graphql query

davidleger95:
## Testing

- go to service card in storybook
- does it load correctly?

## Checklist


src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,34 +1,17 @@
...
-            free
-          }
-        }
-      }
-    }
-  }
-`;
+import productCardQuery from './product-card.graphql';
+
+type PartialProduct = Omit<ProductCardQuery, 'product'>;
```
davidleger95: This doesn't work for some reason 🤔 any thoughts?

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -1,34 +1,17 @@
...
-            free
-          }
-        }
-      }
-    }
-  }
-`;
+import productCardQuery from './product-card.graphql';
+
+type PartialProduct = Pick<ProductCardQuery, 'product'>;
```
davidleger95: This doesn't work for some reason 🤔 any thoughts?

davidleger95:
no sure why the formatting in `src/spec/graphql.schema.json` is all messed up. doesn't `npm run generate:gql` format that for us?

src/components/manifold-service-card/manifold-service-card.tsx
dangodev: Try `ProductCardQuery['product']` (you can just specify this on the `@Prop`)

src/components/manifold-service-card/product-card.graphql
```diff
@@ -0,0 +1,17 @@
+query ProductCard($productLabel: String!) {
+    product(label: $productLabel) {
+      id
+      displayName
+      tagline
+      label
+      logoUrl
+      # must get all plans until 'free' filter works
+      plans(first: 500, orderBy: { field: COST, direction: ASC }) {
```
dangodev: Is this just to see if a product has 1 free plan? Instead you can do `plans(first: 1, free: true)`. If it returns `1` it has a free plan; if `0` it doesn’t.

This was broken before, but we can use that now.

dangodev:
@davidleger95 Have you done an `npm install` recently? You might have an out-of-date gql-codegen (`graphql.schema.json` shouldn’t have changed that much).

src/components/manifold-service-card/product-card.graphql
davidleger95: so free plans are properly ordered now? would that mean the `free` filter works as well? so we could do this:
```graphql
plans(first: 1, free: true) { ... }
```

dangodev: Yup I think that’s what I said above

davidleger95:
Github keeps emailing me saying all the checks failed on this PR... but they all passed 🤔

src/components/manifold-service-card/product-card.graphql
```diff
@@ -0,0 +1,16 @@
+query ProductCard($productLabel: String!) {
+  product(label: $productLabel) {
+    id
+    displayName
+    tagline
+    label
+    logoUrl
+    plans(first: 1, free: true) {
```
dangodev: Alternately, we could do `freePlans: plans(first: 1, free: true)` and change the logic to see if `freePlans.edges.length > 0` to see if it’s free or not.

FWIW I think this is fine, and not a suggestion to change it. Just a micro-optimization I guess.

src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -83,9 +68,10 @@ export class ManifoldServiceCard {
       return;
     }
 
-    const { data, errors } = await this.graphqlFetch({
-      query,
-      variables: { productLabel },
+    const variables: ProductCardQueryVariables = { productLabel };
```
dangodev: 👍

src/types/unpack.ts
```diff
@@ -0,0 +1,3 @@
+export type Unpack<T, K extends string> = T extends { [k in K]: infer U }
+  ? U // if T has K then we extarct the type of K
+  : T;
```
dangodev: What’s the advantage of this as opposed to just using `ProductCardQuery['product']`?

src/components/manifold-service-card/product-card.graphql
davidleger95: I think that's what we're doing. We can't leave the query body blank so we have to add at least one field. for example, this:
```graphql
{
	product(label: "scoutapp") {
    plans(first: 1, free: true)
  }
}
```
sadly gives us this error:
```
"Field \"plans\" of type \"PlanConnection\" must have a selection of subfields. Did you mean \"plans { ... }\"?"
```

so `free` seemed like a reasonable field to grab here. 

A while  ago I suggested having a `count` field so we could just do this:

```graphql
{
	product(label: "scoutapp") {
    freePlans: plans(first: 1, free: true) {
      count
    }
  }
}
```

but I don't think that's a high priority for backend at the moment

src/types/unpack.ts
davidleger95: I thoguht I tried that and didn't think it worked. I'll double check. that would be cleaner

src/components/manifold-service-card/product-card.graphql
dangodev: Yeah I’d like a `count` too, but I talked with James and he said it’s actually a design (performance) choice not to have a total count. So we probably won’t ever get it 😞

src/types/unpack.ts
davidleger95: 🤦‍♂ welp, that's good to know that works at least! 😄

PR #753, opened 2019-11-21
Refactor GraphQL query for manifold-product component

davidleger95:
## Checklist

### DIFF
--- a/src/components/manifold-product-page/manifold-product-page-happo.ts
+++ b/src/components/manifold-product-page/manifold-product-page-happo.ts
@@ -1,4 +1,7 @@
 import jawsdbMock from '../../spec/mock/jawsdb/product';
+import { ProductQuery } from '../../types/graphql';
+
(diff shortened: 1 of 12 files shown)


src/components/manifold-product-page/manifold-product-page-happo.ts
```diff
@@ -1,4 +1,7 @@
 import jawsdbMock from '../../spec/mock/jawsdb/product';
+import { ProductQuery } from '../../types/graphql';
+
+const jawsdbMockProduct = jawsdbMock as ProductQuery['product'];
```
dangodev: 👍  We should never use `as` but in cases here where it’s a) mock data and b) only generating screenshots, this is fine.

davidleger95: yeah, it was either this or switch the mock data to be `any` or some sort of union type. To me this one made the most sense

PR #759, opened 2019-11-22
wip fix types for API data

davidleger95:
## Reason for change

Use correct types for data coming from GraphQL. This is very much a WIP and types are broken everywhere. just seemed like something that needed more input from others.

There are some queries still using the old types generated from the schema. This is a first pass attempt at trying to fix that. High cohesion though is preventing this from being done easily. There's a need for union types and some types need to deeply select parts of the Query type in order to be correct.

I'd like opinions on how we can use these generated types more easily so that the types don't become a burden to deal with and that they're easy to interpret for us as developers.

## Checklist

### DIFF
--- a/src/components/manifold-cost-display/manifold-cost-display.tsx
+++ b/src/components/manifold-cost-display/manifold-cost-display.tsx
@@ -3,8 +3,10 @@ import { h, Component, Element, Prop, FunctionalComponent } from '@stencil/core'
 import { $ } from '../../utils/currency';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
-import { PlanMeteredFeatureEdge } from '../../types/graphql';
+import { PlanListQuery } from '../../types/graphql';
 import { meteredFeatureDisplayValue } from '../../utils/plan';
+
+type MeteredFeatures = PlanListQuery['product']['plans']['edges'][0]['node']['meteredFeatures']['edges'];
 
 /**
  * Base Cost
@@ -38,7 +40,7 @@ const BaseCost: FunctionalComponent<BaseCostProps> = props => {
  * Metered Cost
  */
 interface MeteredCostProps {
-  meteredFeatures: PlanMeteredFeatureEdge[];
+  meteredFeatures: MeteredFeatures;
 }
 const MeteredCost: FunctionalComponent<MeteredCostProps> = ({ meteredFeatures }) => {
   if (meteredFeatures.length === 0) {
@@ -62,7 +64,8 @@ export class ManifoldCostDisplay {
   @Element() el: HTMLElement;
   @Prop() baseCost?: number;
   @Prop() compact?: boolean = false;
-  @Prop() meteredFeatures: PlanMeteredFeatureEdge[] = [];
+  @Prop()
+  meteredFeatures: PlanListQuery['product']['plans']['edges'][0]['node']['meteredFeatures']['edges'] = [];
   @Prop() configurable?: boolean = false;
 
   get isFreeMonthly() {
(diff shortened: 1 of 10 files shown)


src/components/manifold-cost-display/manifold-cost-display.tsx
```diff
@@ -62,7 +64,8 @@ export class ManifoldCostDisplay {
   @Element() el: HTMLElement;
   @Prop() baseCost?: number;
   @Prop() compact?: boolean = false;
-  @Prop() meteredFeatures: PlanMeteredFeatureEdge[] = [];
+  @Prop()
+  meteredFeatures: PlanListQuery['product']['plans']['edges'][0]['node']['meteredFeatures']['edges'] = [];
```
davidleger95: This is annoyingly long. Is there a better way?

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -30,8 +32,10 @@ export class ManifoldPlanDetails {
   @Element() el: HTMLElement;
   @Prop() isExistingResource?: boolean = false;
   @Prop() scrollLocked?: boolean = false;
-  @Prop() plan?: Plan;
-  @Prop() product?: Product;
+  @Prop() plan?:
+    | PlanListQuery['product']['plans']['edges'][0]['node']
+    | GetResourceQuery['resource']['plan'];
```
davidleger95: components that get their data from different queries need union types. Is there a better way to do this?

src/components/manifold-cost-display/manifold-cost-display.tsx
dangodev: I haven’t found a better way, no. But if I reference something a lot, I’ll save it as a `type`. From that type you can pick up where you left off, e.g.:

```ts
type PlanEdge = PlanListQuery['product']['plans']['edges'][0];
type MeteredFeatureEdge = PlanEdge['node']['meteredFeatures']['edges'][0]
```

You have to do this because when it generates types from a document, every object you make it doesn’t make a unique name for, so it’s all one big glob. I tried looking for a way to name some of these in gql-codegen but couldn’t.

In GraphQL, there’s not a way to save a sub-query and name it, is there? Because if so, then it’d probably be easier to grab the thing you want if it had a name.

src/components/manifold-plan-details/manifold-plan-details.tsx
dangodev: I’d say this might be a code smell. The way these types work, you should be able to use one or the other, and it should work because the structure should be the same. And if the data structure is different, then it’s a red flag why this component is handling 2 different schemas.

`PlanListQuery` and `GetResourceQuery`’s `plan` should both request the same data, and using one query or the other should expose when those types diverge. It doesn’t really matter which you use; it’ll be surfaced as an error somewhere (usually at the top-level component that feeds it data).

davidleger95: Yeah, these types don't align, so there might be some work to do there.

There's also fields that are exclusive to configurable features which aren't in the queries at the moment, so those fields aren't defined on the type either. I'm debating on pulling that out or not. It's dead code, so I feel like the answer is yes, it should be removed. But we're also going to add configurable features back at some point. 🤔

dangodev: Dead code is dead code 🤷🏻‍♂️. I’d say pull it.

But yeah IMO types should align for safety. If they don’t it’s worth asking if we should be using the same component or not (if component A and B serve different purposes, how different are they, and when should we separate them?).

PR #783, opened 2019-12-12
Publish docker image for easier testing

gclapperton:
Publishes a docker image on all pushes to master

## Reason for change

To make the entire stack easier to test, I'd like to add `manifoldco/ui` to docker compose. To do this, first, we need to publish a docker image.

This PR introduces `make` to the mix, buuut it does not alter the workflow of those working on this project in any way and also matches the what's been implemented in `starcourt`

Assuming that this PR is accepted, I will work with the right people to configure the secrets and create the repo

## Testing

Currently, you will only be able to test `make docker` which builds and tags the image locally. Once fully configured, you should be able to pull `arigato/ui` from docker

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,17 @@
+name: release
+
+on: [push] # testing
+  #push:
+  #  branches:
+  #    - master
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: make docker-login
+      env:
+        DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+    - run: make release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,9 @@
 name: release
 
-on: [push] # testing
-  #push:
-  #  branches:
-  #    - master
+on:
+  push:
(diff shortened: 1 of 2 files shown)


sslotsky:
No opposition here, but can you explain how it makes the testing easier? Is it because you can pull the latest docker image and run it when you want to look at Storybook? How much easier is that really than checking out the branch and doing a Docker build on your machine? Mainly just curious what the improvements are.

gclapperton:
> No opposition here, but can you explain how it makes the testing easier? Is it because you can pull the latest docker image and run it when you want to look at Storybook? How much easier is that really than checking out the branch and doing a Docker build on your machine? Mainly just curious what the improvements are.

Yes, that's a fair question. 

Currently, to setup the entire backend env for testing locally, we have to start over 50 (yes, over FIFTY!!!... I just just counted :) different docker services (some short lived). We can do this with just `docker-compose up` and, if we need to get the latest images, we just run `docker-compose pull` first.

However, if we want to test something via the UI in, say, Billing-as-a-Service, we'd have to also check out the UI project (if we don't already have it), do a pull, then `npm run docker` (which isn't terribly fast) and open http://localhost:8080

Now, this doesn't sound like too much, but when combined with all of the other things that we have to do/remember, it does add just a little extra friction, especially for those that never actually need to touch the UI code itself and especially for new hires who are still trying to figure out all of the pieces (I had to actively go and figure out how how to run the UI)

Once done, it will be as simple as `docker-compose up` then opening http://ui.arigato.tools

PR #819, opened 2020-01-08
Add additional safety check for mutation observer

TerrySmithDC:
## Reason for change

Adds a check for the custom length property of MutationRecord.addedNodes before attempting a forEach.

## Testing

Affected code is on the initialization of ui, so just ensure ui loads on the browser as normal.

## Checklist

### DIFF
--- a/src/global/app.ts
+++ b/src/global/app.ts
@@ -18,7 +18,11 @@ function mutationCallback(mutationsList: MutationRecord[]) {
   mutationsList
     .filter(mutation => mutation.type === 'childList')
     .forEach(mutation => {
-      if (mutation.addedNodes && mutation.addedNodes.length > 1) {
+      if (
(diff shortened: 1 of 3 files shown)


src/global/app.ts
```diff
@@ -18,7 +18,7 @@ function mutationCallback(mutationsList: MutationRecord[]) {
   mutationsList
     .filter(mutation => mutation.type === 'childList')
     .forEach(mutation => {
-      if (mutation.addedNodes) {
+      if (mutation.addedNodes && mutation.addedNodes.length > 1) {
```
dangodev: Maybe `Array.isArray(mutation.addedNodes)`?

TerrySmithDC: Sure, might as well 💃

PR #835, opened 2020-01-13
Changes client node requirements to be 10.x and up

TerrySmithDC:
## Reason for change

Having client set was accidentally enforcing during npm installs, loosening up the installs to be anything above 8.x

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Update engine version of node
+
 ### [0.9.1]
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@manifoldco/ui",
   "description": "Manifold UI",
-  "version": "1.0.0-rc.0",
+  "version": "0.9.1",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/manifoldco/ui.git"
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Update engine version of node
```
sslotsky: This should be under a `### Fixed` heading

----------------

## [Unreleased]

### Fixed

Update engine version of node

TerrySmithDC: Thanks, I realized I messed up the version tag as well. I'm having a morning 😅

package.json
```diff
@@ -127,6 +127,6 @@
     "webpack": "^4.41.2"
   },
   "engines": {
-    "node": "12.13.x"
+    "node": "~0.8.0"
```
dangodev: I don’t think this is correct

TerrySmithDC: Definitely not, now broke too 😞

dangodev: A good way to test this would be to tag this with a `-test` flag, like so: `0.9.2-test.0`. Then try and install that from npm locally using the latest version of `dashboard-render`.

I don’t remember the specifics, but I do think that sometimes `node` is an array of versions, and that works better? But for now if it works in Render, it works so 🤷‍♂. But we can verify that this is working in their dashboard without merging so we probably should

TerrySmithDC: It also is looking like the now deployments require at least 10.x in order to deploy, I'm attempting to set it to greater then 10.x now

dangodev: Hm. Is it possible to configure Now using `now.json`? We shouldn’t modify our package contents just for our own previews. And for Render’s needs, we don’t have to support 8.x; we can just start at 10.x (I only suggested 8.x because I thought it was possible and was trying to save work if another platform used that)

PR #842, opened 2020-01-13
Relax Node restriction

dangodev:
## Reason for change

Allow Node 12.x and 13.x

## Checklist

### DIFF
--- a/src/components/manifold-service-card/manifold-service-card.tsx
+++ b/src/components/manifold-service-card/manifold-service-card.tsx
@@ -1,4 +1,4 @@
-// Newer version of <manifold-service-card>. Keep this in 1.0.
+// Replaced by <manifold-product-card> (delete this in 1.0)
 import { h, Component, Element, State, Prop, Watch } from '@stencil/core';
 
 import connection from '../../state/connection';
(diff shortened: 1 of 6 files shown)


src/components/manifold-service-card/manifold-service-card.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (errors) {
+      this.errors = errors;
+    }
+  }
+
+  @logger()
+  render() {
+    if (this.product) {
+      return [
+        <manifold-product-card-view
```
dangodev: @davidleger95 I copied this file from `<manifold-product-card>` but internally it just uses `<manifold-product-card-view>`. WDYT? It basically just allows people to use the element + TS types until 1.0

dangodev: Also also, time permitted, I’d probably just leave `<manifold-service-card>` as an empty shell wrapper. But given the quickfix situation, I think it’s less risky to literally copy+paste the thing we know works and is tested than develop a new shell component.

davidleger95: Sounds good to me

PR #856, opened 2020-01-16
add read-only setting for resource plan configurable features

davidleger95:
## Reason for change

Prevent inputs from being changed when viewing a resource plan. Only allow editing when resizing a plan. 

## Testing

These changes will require thorough testing of the plan selector, creating/resizing a resource, and vieing resource details. 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     },
     {
       "path": "./dist/esm/manifold-number-input*.js",
-      "maxSize": "36 kB",
+      "maxSize": "40 kB",
       "compression": "none"
     }
   ],
--- a/src/components/manifold-active-plan/manifold-active-plan.tsx
+++ b/src/components/manifold-active-plan/manifold-active-plan.tsx
@@ -60,13 +60,13 @@ export class ManifoldActivePlan {
       />,
       <manifold-plan-details
         isExistingResource={this.isExistingResource}
+        configuredFeatures={formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges)}
         plan={this.selectedPlan}
         product={this.product}
         regions={this.regions}
         resourceRegion={resourceRegion}
         scrollLocked={true}
         region={this.selectedResource && this.selectedResource.region}
-        configuredFeatures={formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges)}
       >
         <manifold-forward-slot slot="cta">
           <slot name="cta" />
--- a/src/components/manifold-active-plan/manifold-active-plan.tsx
+++ b/src/components/manifold-active-plan/manifold-active-plan.tsx
@@ -2,6 +2,8 @@ import { h, Component, State, Prop, Watch } from '@stencil/core';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
 import { Product, PlanEdge, Resource } from '../../types/graphql';
+import { formatGatewayFeatures } from '../../utils/configuredFeatures';
+
 @Component({
   tag: 'manifold-active-plan',
   styleUrl: 'manifold-active-plan.css',
@@ -49,7 +51,7 @@ export class ManifoldActivePlan {
     const resourceRegion =
       (this.selectedResource && this.selectedResource.region && this.selectedResource.region.id) ||
       undefined;
-
+    console.log('PLAN', formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges));
     return [
       <manifold-plan-menu
         plans={this.plans}
@@ -64,6 +66,7 @@ export class ManifoldActivePlan {
         resourceRegion={resourceRegion}
         scrollLocked={true}
         region={this.selectedResource && this.selectedResource.region}
+        configuredFeatures={formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges)}
       >
         <manifold-forward-slot slot="cta">
           <slot name="cta" />
--- a/src/components/manifold-active-plan/manifold-active-plan.tsx
+++ b/src/components/manifold-active-plan/manifold-active-plan.tsx
@@ -51,7 +51,7 @@ export class ManifoldActivePlan {
     const resourceRegion =
       (this.selectedResource && this.selectedResource.region && this.selectedResource.region.id) ||
       undefined;
-    console.log('PLAN', formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges));
+
     return [
       <manifold-plan-menu
         plans={this.plans}
--- a/src/components/manifold-plan-details/manifold-plan-details.tsx
+++ b/src/components/manifold-plan-details/manifold-plan-details.tsx
@@ -1,14 +1,14 @@
 import { h, Component, Prop, State, Event, EventEmitter, Watch, Element } from '@stencil/core';
-import { check } from '@manifoldco/icons';
 
 import { Option } from '../../types/Select';
-import { Gateway } from '../../types/gateway';
 import { Product, Plan, Region, RegionEdge } from '../../types/graphql';
 import logger from '../../utils/logger';
 import loadMark from '../../utils/loadMark';
 import { configurableFeatureDefaults } from '../../utils/plan';
+import FixedFeature from './components/FixedFeature';
 import MeteredFeature from './components/MeteredFeature';
 import ConfigurableFeature from './components/ConfigurableFeature';
+import { Gateway } from '../../types/gateway';
 
 interface EventDetail {
   planId: string;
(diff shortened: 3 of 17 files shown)


src/components/manifold-select/manifold-select.tsx
```diff
@@ -32,7 +33,12 @@ export class ManifoldSelect {
   @logger()
   render() {
     return (
-      <select name={this.name} required={this.required} onChange={this.onChangeHandler}>
+      <select
+        name={this.name}
+        required={this.required}
+        onChange={this.onChangeHandler}
+        disabled={this.disabled}
```
davidleger95: One question I have about this is do we want to disable the inputs (as I'm doing here) or display the configured features as if they were fixed?

sslotsky: Is the problem here that this component is used in two different contexts? i.e. we use it in the plan selector but we also use it by itself to display the current features?

The question you're asking here is a design question, but my instinct says that it should display as if it were fixed. And if that's the case, then I think that `manifold-resource-plan` becomes a little bit kludgy, because there will be multiple instances where it has to display one component if it's read only and another component otherwise. In that case I'd almost prefer it if `manifold-plan-details` displayed everything fixed, and then the plan selector displayed a different component. What do you think?

dangodev: I could go either way but I think I like the disabled inputs better. This has an affordance that “you _can_ edit these in certain situations; just not now.” It communicates more is possible to the user w/o saying anything.

src/components/manifold-active-plan/manifold-active-plan.tsx
```diff
@@ -64,6 +66,7 @@ export class ManifoldActivePlan {
         resourceRegion={resourceRegion}
         scrollLocked={true}
         region={this.selectedResource && this.selectedResource.region}
+        configuredFeatures={formatGatewayFeatures(this.selectedResource?.configuredFeatures?.edges)}
```
dangodev: This is something to watch—TypeScript doesn’t transform these, so this will break most non-bleeding-edge browsers. Check the output JS to see if these still exist or not.

src/components/manifold-plan-selector/resource.graphql
```diff
@@ -1,5 +1,22 @@
...
+
+          ... on BooleanConfiguredFeature {
+            booleanValue: value
+          }
+          ... on NumberConfiguredFeature {
+            numberValue: value
+          }
+          ... on StringConfiguredFeature {
+            stringValue: value
+          }
```
davidleger95: Problem here: we can't generate types from graphql with this :/

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -61,15 +65,21 @@ export class ManifoldPlanDetails {
...
+  @Watch('configuredFeatures') configuredFeaturesChange(configuredFeatures: Gateway.FeatureMap) {
+    if (this.plan) {
+      this.resetFeatures(this.plan, configuredFeatures);
+    }
+  }
+
   @loadMark()
   @loadMark()
-  componentWillLoad() {
+  componentDidLoad() {
```
dangodev: Why the change here?

src/components/manifold-resource-plan/manifold-resource-plan.tsx
```diff
@@ -26,14 +27,15 @@ export class ManifoldResourcePlan {
       );
     }
 
-    // TODO: remove `as` and update types
```
dangodev: This still needs to stay until the `as` is removed (we should be using generated types, but we need to finish converting `manifold-plan-details` first)

dangodev: Or if the `as` can go it probably should!

src/components/manifold-plan-selector/resource.graphql
dangodev: Does this work? https://graphql-code-generator.com/docs/plugins/fragment-matcher#__docusaurus

dangodev: Actually nevermind—this is for something different.

I don’t think codegen is the problem; I think our GraphQL server is the problem. This should be able to work! But our server needs to have `fragment`s in its schema, and it doesn’t.

Our GraphQL schema should have something like:

```graphql
fragment StringConfiguredFeature on Resource {
  …
}
fragment NumberConfiguredFeature on Resource {
  …
}
…
```

See [this issue](https://github.com/dotansimha/graphql-code-generator/pull/2540) as an example

davidleger95: fragments are on the client though, aren't they? We use fragments for the plan query already: https://github.com/manifoldco/ui/blob/1f053ae1120f21c735afd9dbee20f6b8a8ff23e5/src/components/manifold-plan/plan.graphql

If that works then this should also work, but it appears the types aren't being exposed by the server. however, they are exposed in GraphQL Playground

src/components/manifold-resource-plan/manifold-resource-plan.tsx
davidleger95: ah, oops i'll put it back

src/components/manifold-plan-selector/resource.graphql
davidleger95: Update: Sam made me realize the changes to the API weren't deployed to prod 🤦‍♂ it all works fine haha

dangodev: That’s good news!

src/components/manifold-plan-details/manifold-plan-details.tsx
davidleger95: was just trying testing things out

src/components/manifold-plan-details/manifold-plan-details.tsx
```diff
@@ -61,13 +65,12 @@ export class ManifoldPlanDetails {
...
     // update features
-    this.resetFeatures(newPlan);
+    this.resetFeatures(newPlan, this.configuredFeatures);
   }
   @Watch('resourceRegion') resourceRegionChange(newRegion: string) {
     this.regionId = newRegion;
   }
 
-  @loadMark()
   @loadMark()
```
davidleger95: Any reason why there were 2 of these here?

sslotsky: Not a good one, no doubt :grinning:

src/utils/configuredFeatures.ts
```diff
@@ -0,0 +1,46 @@
+import { Gateway } from 'types/gateway';
+
+export interface ConfiguredFeatureGraphQL {
```
sslotsky: Is there a way to get this type from the resource query definition in graphql.ts? Something like `ResourceQuery['configuredFeatures']`?

src/utils/configuredFeatures.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export interface ConfiguredFeature {
+  label: string;
+  value?: string;
+}
+
+const isDefined = (x: unknown) => typeof x !== 'undefined';
+
+export const formatConfiguredFeatures = (
```
sslotsky: Doesn't look like this one is used?

src/utils/configuredFeatures.ts
davidleger95: yes, i think so. had this here originally because i couldn't generate types for it. but now i can :)

src/utils/configuredFeatures.ts
davidleger95: I'll be used in my next PR, I can take it out for now if you'd like though. what do you think?

sslotsky: Yeah I'd say remove it for now, bring it back when you're ready to use it.

src/components/manifold-active-plan/manifold-active-plan.tsx
davidleger95: I checked the outputs (esm, esm-es5, cjs, collection) and they do transpile to ternaries

src/utils/configuredFeatures.ts
davidleger95: removed

package.json
```diff
@@ -77,7 +77,7 @@
     },
     {
       "path": "./dist/esm/manifold-number-input*.js",
-      "maxSize": "36 kB",
+      "maxSize": "40 kB",
```
davidleger95: Had to increase this limit. Seems like this module is getting quite large in comparison to others though. 🤔

dangodev: That’s fine. This special little chunk is several components in one to save requests. That’s why it’s a separate size benchmark from the others—so it can be bumped on its own without affecting any others.

PR #875, opened 2020-01-24
implement upgradable/downgradable in plan selector

davidleger95:
## Checklist

### DIFF
--- a/src/components/manifold-plan-details/components/ConfigurableFeature.tsx
+++ b/src/components/manifold-plan-details/components/ConfigurableFeature.tsx
@@ -1,6 +1,11 @@
 import { h, FunctionalComponent } from '@stencil/core';
+import { lock } from '@manifoldco/icons';
 import { $ } from '../../../utils/currency';
-import { PlanConfigurableFeatureEdge, PlanFeatureType } from '../../../types/graphql';
+import {
+  PlanConfigurableFeatureEdge,
+  PlanFeatureType,
+  PlanConfigurableFeature,
+} from '../../../types/graphql';
 import { Option } from '../../../types/Select';
 import { fixedDisplayValue } from './FixedFeature';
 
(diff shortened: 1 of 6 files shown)


src/components/manifold-plan-details/components/ConfigurableFeature.tsx
```diff
@@ -9,64 +14,77 @@ interface ConfigurableFeatureProps {
...
+      <dd class="feature-value">
+        <manifold-tooltip labelText="Feature cannot be changed from current plan">
+          <span class="value" data-value={value} data-locked>
+            <manifold-icon class="icon" icon={lock} marginRight />
+            {getDisplayValue(value, configurableFeature.node)}
+          </span>
+        </manifold-tooltip>
+      </dd>,
+    ];
+  }
```
davidleger95: this is probably gonna need some refinements `getDisplayValue()` doesn't quite do what I want it to. just pondering the right abstraction

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
```diff
@@ -9,64 +14,77 @@ interface ConfigurableFeatureProps {
...
+      const selectedValueIndex = featureOptions?.findIndex(o => o.value === value) || -1;
       const selectOptions: Option[] = featureOptions
-        ? featureOptions.map(o => ({
+        ? featureOptions.map((o, i) => ({
             label: `${o.displayName} (${$(o.cost)})`,
             value: o.value,
+            disabled:
+              selectedValueIndex !== -1 &&
+              isExistingResource &&
+              ((downgradable && selectedValueIndex > i) || (upgradable && selectedValueIndex < i)),
```
dangodev: Are we sure this is how we want it to work? Should upgradable & downgradable come from option order? I’m wondering about that because now those can never be re-ordered.

I’ll defer to someone else that can say for certainty, but I’d probably only `disable` a string feature if both upgradable & downgradable are `false`

davidleger95: this is also icky. @dangodev you mentioned that number features were the only ones that updgradable/downgradable applies to but if the picnic basket from stage is a possible product config, it doesn't seem that's true. Features could definitely be unchangeable though, and might be represented by `upgradable: true` and `downgradable: true`. In that case, this code would not be reached though because of the "locked" state above that checks for both of these.

Anyway, I'm just rambling. I know you've said repeatedly that it only applies to number features, but the data keeps telling me otherwise, if you are correct, then our test data and data model is wrong and they should probably change.

cc: @domenicrosati as a backend person who might know more about the possible reasons for upgradable/downgradable being the way it is. :)

domenicrosati: Heheh I wish I could answer but I do not know

domenicrosati: We can find out on monday

dangodev: The intent of upgradable/downgradable is to just surface provider-level information sooner to users. There are a number of reasons that we will never know about on the provider end that will cause a resize operation to fail. However, upgrading/downgrading is one piece of logic that’s so simple, we expect a provider to throw a resize error on their end, but we can also warn the user beforehand that something will fail.

In this light—on what resizes a provider allows—we can say for sure that numbers have clear “ups” and “downs.” If a provider sets it here, then this seems intentional.

But our data model was very polymorphic—there are so many fields that are contingent on other fields’ values that makes upgradable/downgradable tough to parse. Was this contingent on a feature type? I had always assumed so.

Anyway, regarding feature order of strings and numbers, yes, it is possible that we could enforce upgradable/downgradable based on order in our DB. We can do that, and there probably is code there that can do that. But my bigger question is: **are providers aware that the specific order of their features determines upgradability when they configure that setting?** That’s the thing I’ve always cast doubt on.

dangodev:
Screen Shot 2020-01-24 at 3 11 10 PM

When a feature is at its minimum, could there be better language here? This seems different than “downgradable”

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
dangodev: This is one of those rare things that I think is best in a unit test, over VRT or E2E. I’d recommend just writing tests first to make sure this displays what you want it to.

But FWIW I think this pattern is fine? 🤷‍♂

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
```diff
@@ -101,8 +119,8 @@ const ConfigurableFeature: FunctionalComponent<ConfigurableFeatureProps> = ({
             decrement-disabled-label="This feature is not downgradable"
             increment-disabled-label="This feature is not upgradable"
             increment={numericDetails.increment}
-            max={numericDetails.max}
-            min={numericDetails.min}
+            max={isExistingResource && upgradable ? (value as number) : numericDetails.max}
+            min={isExistingResource && downgradable ? (value as number) : numericDetails.min}
```
dangodev: Ah one change actually: we need a `disabled` attribute here. I can still edit the `<input>` itself even if it’s `!upgradable && !downgradable`

dangodev: Oh! And if I change the `<input>` then it removes the `disabled` on the increase/decrease because of this logic too:

Screen Shot 2020-01-24 at 3 18 46 PM

Screen Shot 2020-01-24 at 3 19 40 PM

davidleger95: It should be caught before that point with the "locked" UI

dangodev: Ah OK. In full transparency I did only set `.isExistingResource = true` on the `<manifold-plan-details>` component manually in Storybook so if this bug is nonexistent then ignore

src/components/manifold-plan-details/components/ConfigurableFeature.tsx
davidleger95: I tested out picnic basket stage just to see what it would to if i downgraded a feature with `downgradable: false` and it let me downgrade, whereas the number feature which has both false, errors when you try to change it. Now I know that can't be representative of all products but it's not telling me that the flag is necessarily relevant for string features on that product. 

So I'm going to assume you are correct with number features being the only type where these flags are relevant. I'll check all out configurable plans and see what the flags are for those too.

PR #930, opened 2020-03-02
One timeout to rule them all

sslotsky:
## Reason for change

When the `manifold-resource-container` gets a GraphQL error, it seems that it's actually setting _two_ timeouts to re-fetch the resource. Only one of these timeouts gets cleared when the component unmounts, so the other timeout keeps on going.

## Testing

This issue was discovered when testing team resources, so it's probably best to test this on Render as follows:

1. Make sure you're part of a team
2. Switch to the team context
3. Create a new resource
4. Click on the resource you created
5. Observe the GraphQL polling loop occurring on the resource page
6. Navigate to the home page
7. Make sure it's no longer polling

## Checklist

### DIFF
--- a/src/components/manifold-resource-container/manifold-resource-container.tsx
+++ b/src/components/manifold-resource-container/manifold-resource-container.tsx
@@ -63,13 +63,9 @@ export class ManifoldResourceContainer {
       this.resourceLoad.emit();
     }
 
-    if (
-      this.refetchUntilValid &&
-      (errors ||
-        !data ||
-        !data.resource ||
-        data.resource.status.label !== ResourceStatusLabel.Available)
-    ) {
+    const resourceNotAvailable = data?.resource?.status.label !== ResourceStatusLabel.Available;
+
+    if (this.refetchUntilValid && (errors || resourceNotAvailable)) {
       this.timeout = window.setTimeout(() => this.fetchResource(this.resourceLabel), 3000);
     }
   };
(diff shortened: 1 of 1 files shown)


src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -65,16 +65,13 @@ export class ManifoldResourceContainer {
 
     if (
       this.refetchUntilValid &&
-      (!data || !data.resource || data.resource.status.label !== ResourceStatusLabel.Available)
+      (errors ||
+        !data ||
+        !data.resource ||
+        data.resource.status.label !== ResourceStatusLabel.Available)
     ) {
```
davidleger95: This seems like a long conditional, I think it would be more readable if it was labeled with variables:
```js
const noResource = !data ||  !data.resource;
const resourceNotAvailable = data.resource.status.label !== ResourceStatusLabel.Available;

if (this.refetchUntilValid && (noResource || resourceNotAvailable)) {
```

davidleger95: could even go anouther step:
```js
const noResource = !data || !data.resource;
const resourceNotAvailable = data.resource.status.label !== ResourceStatusLabel.Available;
const notValid = errors || noResource || resourceNotAvailable;

if (this.refetchUntilValid && notValid) {
```

sslotsky: Thought about this some and optional chaining seems nice here

```ts
    const resourceNotAvailable = data?.resource?.status.label !== ResourceStatusLabel.Available;

    if (this.refetchUntilValid && (errors || resourceNotAvailable)) {
      this.timeout = window.setTimeout(() => this.fetchResource(this.resourceLabel), 3000);
    }
```

PR #935, opened 2020-03-03
Combine connection in global script, and combine logger & loadMark together

dangodev:
## Reason for change

Improves code-splitting in a webpack setup to reduce duplication of common utilities.

**Before**
Screen Shot 2020-03-03 at 12 06 52

- `connection` duplicated across **39** chunks in an example app, footprint `9 KB` each. That’s `351 KB` in total 
- `logger` and `loadMark` are their own separate modules, and while they almost always make it in the same chunk, there’s still a little duplication.

**After**
Screen Shot 2020-03-03 at 12 06 59

- `connection` is combined with entry code, total size `30 KB`. Now it’s only loaded once in the entire app (it gets loaded whenever `@manifoldco/ui` is imported directly, which only happens once in this app).
- `logger` and `loadMark` by their nature weren’t easy to dedupe (it’s a decorator that generates JSX). However they were combined so that even though they are still duplicated, it’s as good as it can be without rewriting them all (or manually inlining them.)

## Testing

The preview links should work. I believe it will be pretty obvious whether or not this is working.

## Feedback

So this has some pros and cons. Pros: it saves users > `320 KB` of downloaded JS (pre-min, etc. etc.), and it ~also solves any weirdness from multiple `ConnectionState()`s possibly conflicting.~

Edit: it actually looks like webpack prevents duplicated code from executing a 2nd time, which I didn’t know! So perhaps this is a footprint optimization and nothing more.

Cons: it combines files into a less-clean structure than before.

Is this OK?

## Checklist


src/global/app.ts
```diff
@@ -1,5 +1,120 @@
...
+    setAuthToken: this.setAuthToken,
+    wait: () => this.waitTime,
+  });
+}
+
+export const connection = new ConnectionState();
+
+/**
+ * Performance Monitoring Module
+ */
```
dangodev: This is 1/2 major changes in this PR: `state/connection` was moved ➡️ `global/app`. This takes advantage of Stencil’s `globalScript` setting to make sure this only gets loaded once.

src/utils/logger.tsx
```diff
@@ -1,14 +1,35 @@
...
+        mark(this.el, 'first_render');
+        mark(this.el, 'first_render_with_data');
+      }
+      return originalMethod.apply(this); // call original method
+    };
+
+    return descriptor;
+  };
+}
+
```
dangodev: 2/2 major changes in this PR: `loadMark` is moved inside `logger` since they’re almost always used in tandem.

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -1,9 +1,8 @@
 import { h, Element, Component, Prop, State, Watch, Event, EventEmitter } from '@stencil/core';
 
-import connection from '../../state/connection';
 import ResourceTunnel from '../../data/resource';
-import logger from '../../utils/logger';
-import loadMark from '../../utils/loadMark';
+import { connection } from '../../global/app';
+import logger, { loadMark } from '../../utils/logger';
```
dangodev: The rest of this PR is just updating imports

src/components/manifold-plan-selector/manifold-plan-selector.tsx
```diff
@@ -87,7 +86,7 @@ export class ManifoldPlanSelector {
 
     this.resource = undefined;
 
-    const { data } = await this.graphqlFetch({
+    const { data } = await this.graphqlFetch<{ resource?: Resource }>({
```
dangodev: This was a weird TypeScript error that I had to fix. Not sure why it was blowing up on me this go-round

dangodev:
> I'm interested to know if that's been happening.

So as best I can see, fortunately this hasn’t been happening? When I examine the chunks of Render, that `connection.ts` file is duplicated and embedded in 39 different chunks. However, if I `console.log()` in that constructor locally, I am not not seeing that class initialized more than once.

I guess webpack does some magic in IDing modules, and it flags code as duplicate when it executes.

PR #938, opened 2020-03-06
Add owner-id to more components

dangodev:
## Reason for change

Adds `owner-id` to more components that make resource queries.

Components added:

| Component | Test Added |
|:-------------|:-----------:|
| `<manifold-copy-credentials>` | ✅ |
| `<manifold-credentials>` | ✅ |
| `<manifold-data-deprovision-button>` | ✅ |
| `<manifold-data-has-resource>` | ✅ |
| `<manifold-data-product-name>` | ✅ |
| `<manifold-data-resource-logo>` | ✅ |
| `<manifold-data-resize-button>` | ✅ |
| `<manifold-data-rename-button>` | ✅|
| `<manifold-data-sso-button>` | ✅ |
| `<manifold-no-credentials>` | (covered in `manifold-resource-credentials`) |
| `<manifold-plan-selector>` | ✅ |
| `<manifold-resource-card>` | ✅ |
| `<manifold-resource-credentials>` | ✅ |
| `<manifold-resource-deprovision>` | ✅ |
| `<manifold-resource-rename>` | ✅ |
| `<manifold-resource-sso>` | ✅ |

## Testing

This shouldn’t affect existing functionality. It’s been tested in Render and components seem to work normally, but I’ll need @jelmersnoeck’s help next week as I’m encountering a staging bug with my specific profile.

## Checklist

### DIFF
--- a/src/components/manifold-data-deprovision-button/delete.graphql
+++ b/src/components/manifold-data-deprovision-button/delete.graphql
@@ -1,5 +1,5 @@
-mutation DeleteResource($resourceId: ID!) {
-  deleteResource(input: { resourceId: $resourceId }) {
+mutation DeleteResource($resourceId: ID!, $owner: ID) {
+  deleteResource(input: { resourceId: $resourceId, ownerId: $owner }) {
     data {
       id
       label
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -9,35 +9,35 @@ jobs:
       matrix:
         node-version: [10.x, 12.x, 13.x]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - run: npm install
       - run: npm run build
   bundlesize:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - run: npm install
       - run: npm run bundlesize
   changelog:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - run: npm install
       - run: npm run format:changelog
   schema:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - run: npm install
       - run: npm run test:schema
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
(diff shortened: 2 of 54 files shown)


src/components/manifold-data-deprovision-button/delete.graphql
```diff
@@ -1,5 +1,5 @@
-mutation DeleteResource($resourceId: ID!) {
-  deleteResource(input: { resourceId: $resourceId }) {
+mutation DeleteResource($resourceId: ID!, $owner: ID) {
+  deleteResource(input: { resourceId: $resourceId, ownerId: $owner }) {
```
jelmersnoeck: It makes me sad that we use `owner` and `ownerId` in our API. Is this something we can clean up quickly and push out here?

Eg. only use `owner` or `ownerId`, deprecate the other field (allow both for now) and then only just use the new one in this PR?

dangodev: This is the Web Components, so no; I’d have to make a separate PR for our GraphQL library. +1 I think `owner` should consistently be an object w/ fields, and `ownerId` should be the ID in queries in mutations

garyposter: IIUC, you are saying that we _could_ do what Jelmer is proposing, but it would have to be in stages: this change, then the GraphQL library, then another change here (probably my preference); or put this on hold, make the GraphQL library change, and come back to this to adjust. Is that right? If so, is this something we can do in the available time, with other constraints and considerations, perhaps spreading work across the team?

dangodev: Yes, exactly. Deprecation takes time because we have to be sure the old field isn’t being used anymore. So we can make a PR for that in GraphQL fairly easily, but we’d have to support the old field for at least a while either way. We also have some hard-coded queries in the Render Dashboard so renaming `owner` to `ownerId` is not a one-step change.

domenicrosati: Its a bit tricky because some places we do both:

https://github.com/manifoldco/graphql/blob/a3b2f26d2a302c4b817ea986a3095cc3d0a2c3ec/schema/resource.graphql#L123-L127

This is a good example because what do we do in that case?

ID can only take a manifoldID which this is may not be (it might be a render ID).

Do we add a third field here? `ownerId2` (better name desired)

dangodev:
Gonna hold on merging this until I test locally that it works in Render. I want to make sure that I didn’t introduce any breaking changes in this PR

src/components/manifold-data-deprovision-button/delete.graphql
dangodev: Oh no I’m good with that (the current version) 👆 . IMO that’s what we want: `owner` is always an object (which includes `id`); `ownerId` is always an ID string. That we could make more consistent across our API

PR #945, opened 2020-03-10
Remove conditional GraphQL queries

dangodev:
## Reason for change

Before, we would load GraphQL query A or B based on props. That could lead to type errors. This PR consolidates queries so that every component only uses one query, and optional props are just passed straight through to GraphQL.

I tested optional `owner` against our staging GraphQL API, and it works if you have `$owner` in the query but it’s marked optional.

## Testing

This should be a fairly-safe change, because we’re testing all our `variables` against the generated GraphQL types. So if anything’s off, it should fail to build.

Further, I’ve tested in Render Dashboard locally, and it works; Storybook works; and I tested queries in GraphQL playground if they work. UI should throw a warning if the GraphQL queries are off. 

I don’t think the reviewer has to rigorously test this PR since it’ll be tested pretty thoroughly on Render (or at minimum, just try resource queries in Storybook)

## Checklist


src/components/manifold-data-provision-button/create-with-owner.graphql
```diff
@@ -4,7 +4,7 @@ mutation CreateResourceWithOwner(
   $regionId: ID!
   $resourceLabel: String!
   $configuredFeatures: [ConfiguredFeatureInput!]
-  $owner: ProfileIdentity!
+  $owner: ProfileIdentity
```
dangodev: The entire pattern of this PR is summed up in this pattern:

Step 1: make `$owner` optional

src/components/manifold-data-provision-button/create.graphql
```diff
@@ -1,23 +0,0 @@
...
-      regionId: $regionId
-      configuredFeatures: $configuredFeatures
-    }
-  ) {
-    data {
-      id
-      label
-    }
-  }
-}
```
dangodev: Step 2: delete the owner-less mirror of the query (between the two, sometimes we’d would only check types on one which could lead to type errors)

src/components/manifold-data-provision-button/manifold-data-provision-button.tsx
```diff
@@ -141,8 +139,8 @@ export class ManifoldDataProvisionButton {
       owner: this.ownerId,
     };
 
-    const { data, errors } = await this.graphqlFetch<CreateResourceMutation>({
-      query: this.ownerId ? createResourceWithOwner : createResource,
+    const { data, errors } = await this.graphqlFetch<CreateResourceWithOwnerMutation>({
+      query: createResourceWithOwner,
```
dangodev: Step 3: remove `this.ownerId ? …` ternaries and just use one query for safety

dangodev: The rest of this PR just repeats this pattern for every other component.

Tests for `owner` were added in the previous PR

davidleger95:
I was thinking about this ever since i saw the first `withOwner` query but it was one of those things where I just assumed there was a reason for it and didn't question it haha

davidleger95:
Just need that pesky CI to pass :)

dangodev:
>  I just assumed there was a reason for it and didn't question it haha

Yeah I’m not sure who started the pattern (was probably me). I wouldn’t say this PR fixes a bug; just in dealing with this just now hopefully it makes someone else’s life easier! And hopefully saved us from a feature triage

src/components/manifold-data-resource-list/resources-with-owner.graphql
```diff
@@ -1,4 +1,4 @@
-query DATA_RESOURCES_WITH_OWNER($first: Int!, $after: String!, $owner: ProfileIdentity!) {
+query DATA_RESOURCES_WITH_OWNER($first: Int!, $after: String!, $owner: ProfileIdentity) {
```
dangodev: @domenicrosati @jelmersnoeck we have inconsistent naming patterns in our queries due to switching halfway. What are your thoughts on this? Consistent naming? Or better to leave alone for logs?

domenicrosati: @dangodev can you give an example of what you mean?

dangodev: Here it’s `SCREAMING_SNAKE_CASE` but more recently we use `PascalCase` for query names (`CreateResourceWithOwner`)

domenicrosati: oooooo. I don't have an opinon other than turbo pascal is a great IDE

dangodev: It’s just a GraphQL query ID; basically it can be named anything and doesn’t affect the query itself. It just provides an easy shorthand to reference in logs.

dangodev: OK. Maybe we’ll just rename the query names as a future chore… but not now

domenicrosati: and by that i mean consistant naming seems great but i can't say that from a logging backend persictive it brings that much value TBH

dangodev:
It seems like CI is passing… but GHA is being slow / buggy at the moment

PR #953, opened 2020-03-12
Fix GITHUB_SHA env var

dangodev:
## Reason for change

GitHub Actions is not easy to validate without merging code

## Checklist


.github/workflows/branch.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
       - run: npm install
-      - run: npm run build
+      - run: make package
```
dangodev: Not part of this package, but this just tests out the publish script on every PR, rather than only running it when we deploy

PR #978, opened 2020-03-30
Delay fetching resource info until there’s a token

dangodev:
## Reason for change

Render is getting a lot of unauthenticated logs in their system, caused mostly by these 2 components firing requests before waiting for a token. While this isn’t a perfect solution, it at least takes care of the noisier components.

## Testing

This has been tested locally in Render Dashboard and seems to work fine.

## Checklist

### DIFF
--- a/src/components/manifold-resource-container/manifold-resource-container.tsx
+++ b/src/components/manifold-resource-container/manifold-resource-container.tsx
@@ -55,6 +55,11 @@ export class ManifoldResourceContainer {
       return;
     }
 
+    // for this component, wait for auth before fetching
+    if (!connection.getAuthToken()) {
+      return;
+    }
+
     const variables: ResourceWithOwnerQueryVariables = { resourceLabel, owner: this.ownerId };
     const { data, errors } = await this.graphqlFetch<ResourceWithOwnerQuery>({
       query: queryWithOwner,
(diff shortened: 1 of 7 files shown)


src/components/manifold-resource-list/manifold-resource-list.tsx
```diff
@@ -54,6 +54,11 @@ export class ManifoldResourceList {
       return;
     }
 
+    // for this component, wait for auth before fetching
+    if (!connection.getAuthToken()) {
+      return;
+    }
```
davidleger95: is there a reason to have this in a separate if-statement from the one above?

dangodev: readability, and findability. I’m adding something special here (possibly temporarily) that I didn’t want to invisibly mix into the same `if()` checks you see in all the other components.

src/components/manifold-resource-container/manifold-resource-container.tsx
```diff
@@ -55,6 +55,11 @@ export class ManifoldResourceContainer {
       return;
     }
 
+    // for this component, wait for auth before fetching
+    if (!connection.getAuthToken()) {
+      return;
+    }
```
dangodev: @sslotsky is this dumb Y/N

